Issue #23, opened 2015-04-22
Update repo docs


tbetbetbe:
Add README.md, CONTRIBUTING.md, CHANGELOG.md and other high-level docs cf [ruby](https://github.com/google/google-auth-library-ruby)(https://github.com/google/google-auth-library-ruby)

N.B. This is different from #4, which involves updating documentation on Google's official documentation site.

PR #30, opened 2015-04-24
Update README.md

tbetbetbe:
Additions to the README before going public.
- clarify relationship with google-api-php-client
- more detail to the composer install instructions


stanley-cheung:
For Issue #23

Issue #24, opened 2015-04-22
Add travis.yml


tbetbetbe:
To allow the tests to run on Travis.

PR #33, opened 2015-04-24
Update firebase/php-jwt dependency

stanley-cheung:
grpc side does not like dev-master.


tbetbetbe:
Can you update .travis.yml and composer.json to exclude PHP 5.3 ?

stanley-cheung:
Yay it passed!

stanley-cheung:
For issue #24

PR #39, opened 2015-05-06
fix gce credentials class

stanley-cheung:


stanley-cheung:
Related to the refactor done in issue #16

PR #41, opened 2015-05-07
increase timeout on gce connection

stanley-cheung:
I tracked down the intermittent error on the compute_engine test. When we connect to GCE to get the initial metadata, the call is set to have a timeout of 0.1s. That fails around 50% of the time. So I tested increasing the timeout:

failure rate:
 * 0.1s: 53%
 * 0.2s: 49%
 * 0.25s: 10%
 * 0.3s: 0%


stanley-cheung:
For Issue #42

PR #44, opened 2015-05-13
Add jwt access credentials

stanley-cheung:
* Add ServiceAccountJwtAccessCredentials class
 * Change updateMetadata callback function signature to accept authUri parameter
 * if no scope is set in ServiceAccountCredentials, use ServiceAccountJwtAccessCredentials instead
 * to calculate the token for the ServiceAccountJwtAccessCredentials class, use the passed in authUri to encode the JWT. no outbound call needs to be made
 * add some unit tests


stanley-cheung:
Fixes issue #43

stanley-cheung:
Please review, thanks!

src/ServiceAccountCredentials.php
```diff
@@ -104,4 +104,102 @@ public function getCacheKey()
...
+/**
+ * Authenticates requests using Google's Service Account credentials via
+ * JWT Access.
+ *
+ * This class allows authorizing requests for service accounts directly
+ * from credentials from a json key file downloaded from the developer
+ * console (via 'Generate new Json Key').  It is not part of any OAuth2
+ * flow, rather it creates a JWT and sends that as a credential.
+ */
+class ServiceAccountJwtAccessCredentials extends CredentialsLoader
```
kesar: I would add this class in a different file. At the same time now that there are two different ServiceAccountCredentials an interface could make sense with the method  updateMetadata

tbetbetbe: +1 for having a separate class.

anthmgoogle:
Yes, this looks already done in this case.

Issue #48, opened 2015-05-22
fromWellKnownFile non-Windows path


stanley-cheung:
https://github.com/google/google-auth-library-php/blob/master/src/CredentialsLoader.php#L99

Should have .config/ as part of the path for non-Windows

PR #49, opened 2015-05-22
fix non-windows well known file path

stanley-cheung:
Fixes issue #48 

 * add back an extra component in non-windows well known file path
 * fixes unit test


src/CredentialsLoader.php
```diff
@@ -95,8 +95,12 @@ public static function fromEnv($scope = null)
   public static function fromWellKnownFile($scope = null)
   {
     $rootEnv = self::isOnWindows() ? 'APPDATA' : 'HOME';
-    $root = getenv($rootEnv);
-    $path = join(DIRECTORY_SEPARATOR, [$root, self::WELL_KNOWN_PATH]);
+    $path = [getenv($rootEnv)];
+    if (!self::isOnWindows()) {
+      $path[] = '.config';
```
kesar: this could be a constant near ```WELL_KNOWN_PATH```

PR #57, opened 2015-08-20
adds client id/secret to code,refresh,and password grant types

bshaffer:
the call to [`generateCredentialsRequest`](https://github.com/google/google-auth-library-php/blob/master/src/OAuth2.php#L369), called from [`fetchAuthToken`](https://github.com/google/google-auth-library-php/blob/master/src/OAuth2.php#L420), does not include the client credentials. This results in a `400 Bad Request`

```json
{
	"error": "invalid_request",
	"error_description": "Missing parameter: client_id"
}
```

I have changed this so the credentials are added to the request if they've been supplied to the 0Auth client

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -381,13 +381,16 @@ class OAuth2 implements FetchAuthTokenInterface
       case 'authorization_code':
         $params['code'] = $this->getCode();
         $params['redirect_uri'] = $this->getRedirectUri();
+        $this->addClientCredentials($params);
         break;
       case 'password':
         $params['username'] = $this->getUsername();
         $params['password'] = $this->getPassword();
+        $this->addClientCredentials($params);
         break;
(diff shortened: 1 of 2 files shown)


tbetbetbe:
Off topic comment: I think I saw one PR where cla:yes marked OK, but this is the second where it's not ? 
Are you always using the email address that's part of the google project when sending PRs ?  It'd be useful to trigger cla:yes consistentnly

bshaffer:
@tbetbetbe I don't really understand how the google bot works, but I am using the account that is a part of the Google / Google Cloud Platform organizations.

It's probably because I don't have access to *these* repos, and I've never filled out the `cla` as a non-googler. I don't really know how to get myself whitelisted, but I'll look into it.

bshaffer:
Okay, I found some info about it and I should be good to go now :D

PR #69, opened 2015-09-17
fix #67 - consolidate DefaultCredentials into CredentialsLoader

### LINKED ISSUE #67: DefaultCredentials can be consolidated into CredentialsLoader
1. The autoloader will not be able to locate `DefaultCredentials` because it is in the file of a different class name
2. the call to `CredentialsLoader::fromWellKnownFile` will throw a fatal error because `makeCredentials` does not exist in this class

I recommend we move `function makeCredentials` into the `CredentialsLoader` class and remove the `DefaultCredentials` class all together.

bshaffer:


tbetbetbe:
LGTM

PR #74, opened 2015-10-01
adds docs, changes is_null to empty, adds clone

bshaffer:


src/AuthTokenFetcher.php
```diff
@@ -113,7 +114,7 @@ public function onBefore(BeforeEvent $event)
     //
     // TODO: correct caching; enable the cache to be cleared.
     $cached = $this->getCachedValue();
-    if (!is_null($cached)) {
+    if (!empty($cached)) {
```
bshaffer: without this, I've run into empty strings being saved and then causing auth errors

src/OAuth2.php
```diff
@@ -558,7 +558,7 @@ public function buildFullAuthorizationUri(array $config = null)
     }
 
     // Construct the uri object; return it if it is valid.
-    $result = $this->authorizationUri;
+    $result = clone $this->authorizationUri;
```
bshaffer: without this, if you build the authorization URI twice in one session, the querystring parameters will be added twice

bshaffer:
The `clone` issue can be seen in the failing tests [here](https://travis-ci.org/google/google-api-php-client/jobs/83237594)

PR #75, opened 2015-10-08
adds AppIdentityCredentials

### LINKED ISSUE #46: PHP Default Credentials check for and user GAE App Identity Credentials
The current implementation of ApplicationDefaultCredentials in PHP does not yet support the GAE App Identty. The first class GAE lanauges should check for this environment and use this identity after checking for the environment variable and well-known-file, but before checking for GCE.

bshaffer:
fixes #46

### DIFF
--- a/src/AppIdentityCredentials.php
+++ b/src/AppIdentityCredentials.php
@@ -1,0 +1,106 @@
+<?php
+/*
+ * Copyright 2015 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Client;
+use google\appengine\api\app_identity\AppIdentityService;
+
+/**
+ * AppIdentityCredentials supports authorization on Google App Engine.
+ *
+ * It can be used to authorize requests using the AuthTokenFetcher, but will
+ * only succeed if being run on App Engine:
+ *
+ *   use GuzzleHttp\Client;
+ *   use Google\Auth\AppIdentityCredentials;
+ *   use Google\Auth\AuthTokenFetcher;
+ *
+ *   $gae = new AppIdentityCredentials('https://www.googleapis.com/auth/books');
+ *   $subscriber = new AuthTokenFetcher($gae);
+ *   $client = new Client([
+ *      'base_url' => 'https://www.googleapis.com/books/v1',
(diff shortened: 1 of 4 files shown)


src/AppIdentityCredentials.php
```diff
@@ -0,0 +1,106 @@
...
+   * @param $client GuzzleHttp\ClientInterface optional client.
+   * @return array the auth metadata:
+   *  array(2) {
+   *   ["access_token"]=>
+   *   string(3) "xyz"
+   *   ["expiration_time"]=>
+   *   string(10) "1444339905"
+   *  }
+   */
+  public function fetchAuthToken(ClientInterface $client = null)
```
stanley-cheung: We had been naming this kind of unused variable `$unusedClient` elsewhere in the library. See https://github.com/google/google-auth-library-php/blob/master/src/IAMCredentials.php#L75

src/AppIdentityCredentials.php
```diff
@@ -0,0 +1,106 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+use GuzzleHttp\ClientInterface;
+use GuzzleHttp\Client;
+use google\appengine\api\app_identity\AppIdentityService;
```
stanley-cheung: Do we need to include this at the top level `composer.json` file?

src/AppIdentityCredentials.php
```diff
@@ -0,0 +1,106 @@
...
+  public static function onAppEngine()
+  {
+    return (isset($_SERVER['SERVER_SOFTWARE']) &&
+        strpos($_SERVER['SERVER_SOFTWARE'], 'Google App Engine') !== false);
+  }
+
+  /**
+   * Implements FetchAuthTokenInterface#fetchAuthToken.
+   *
+   * Fetches the auth tokens using the AppIdentityService it is available.
```
stanley-cheung: Missing an `if` I believe

src/AppIdentityCredentials.php
bshaffer: We actually don't, because if we do then it conflicts with the library on appengine (which gets loaded automatically).

I'm not sure how to best handle that... I suppose a comment on this line would be appropriate. Or adding a mock class somewhere.

bshaffer:
Thank you for the thorough review, @stanley-cheung! I have addressed your concerns, and added a mock [AppIdentityService](https://github.com/google/google-auth-library-php/pull/75/files#diff-7bfa85c2e884aa91fc9a7d09ded8a7f1R98) class for when developing outside of App Engine. Really, the best way would be for the user to not even be aware the library exists, as these calls made locally will fail anyway.

However, as long as the developers call `$client->useApplicationDefaultCredentials()` and they define the environment variable `GOOGLE_APPLICATION_CREDENTIALS` as a path to their service account credentials in their dev environment, this will function as expected both locally and in production.

src/AppIdentityCredentials.php
stanley-cheung: So you are saying that if this library is used in App Engine, the `google\appengine` library will be loaded automatically? That is interesting.

But what if the library is used outside of App Engine in a different context? Will this line throw an error?

stanley-cheung: Ah I just saw your comment below and looks like we are going to be fine here.

PR #79, opened 2015-10-29
Upped dependency to Firebase 3.0, fixed use statements

SamanthaAdrichem:
We use a specific vendor for the slim framework and jwt authentication ( tuupola/slim-jwt-auth ) which requires firebase 3.0,

The differences according to the firebase log are
- namespacing
- better PHP Doc
- Require a non-empty key to decode and verify a JWT

(https://github.com/firebase/php-jwt/commit/fa8a06e96526eb7c0eeaa47e4f39be59d21f16e1)

The unittests succeed with these changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
   "homepage": "http://github.com/google/google-auth-library-php",
   "license": "Apache-2.0",
   "require": {
-    "firebase/php-jwt": "2.0.0",
+    "firebase/php-jwt": "3.0.0",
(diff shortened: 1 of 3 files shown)


SamanthaAdrichem:
I signed it!

kesar:
:+1:

bshaffer:
LGTM - newer is better. @tbetbetbe?

tbetbetbe:
+1

PR #80, opened 2015-11-02
supports both firebase jwt versions

bshaffer:


tests/mocks/JWT.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+namespace Google\Auth\Tests;
+
+class JWT
```
SurferJeffAtGoogle: Add a comment for this class.

src/OAuth2.php
```diff
@@ -1072,6 +1071,25 @@ private function coerceUri($uri)
     }
   }
 
+  private function jwtDecode($idToken, $publicKey, $allowedAlgs)
+  {
+    if (class_exists('Firebase\JWT\JWT')) {
+      return \Firebase\JWT\JWT::decode($idToken, $publicKey, $allowedAlgs);
+    }
+
+    return \JWT::decode($idToken, $publicKey, $allowedAlgs);
```
tmatsuo: You removed the `use` for `\JWT` namespace at the top. Does this code path really work? Do you have a test case?

src/OAuth2.php
```diff
@@ -1072,6 +1071,25 @@ private function coerceUri($uri)
...
+  }
+
+  private function jwtEncode($assertion, $signingKey, $signingAlgorithm)
+  {
+    if (class_exists('Firebase\JWT\JWT')) {
+      return \Firebase\JWT\JWT::encode($assertion, $signingKey,
+                       $signingAlgorithm);
+    }
+
+    return \JWT::encode($assertion, $signingKey, $signingAlgorithm);
```
tmatsuo: ditto

bshaffer:
@tmatsuo yes, I do have a test case. If you check out the [travis.ci tests](https://travis-ci.org/google/google-auth-library-php/builds/88862712), you'll notice the environment variable FIREBASE_JWT_VERSION (set in [.travis.yaml](https://github.com/google/google-auth-library-php/pull/80/files#diff-354f30a63fb0907d4ad57269548329e3R12))

You can also manually require one version or the and run the tests.

Since  `firebase/php-jwt:2.0` is an older version, I'm not opposed to just running it on one platform (presumably PHP 5.5, the minimum non-eol'd version of PHP) rather than on all four. This is especially true once we add Guzzle, since that'd split the matrix to 16.

src/OAuth2.php
bshaffer: I could surround `\JWT` with a `class_exists` as well, and throw an exception saying "you must install either 2.0 or 3.0 of firebase/php-jwt to call these functions". That would be a more obvious error to the user, anyway.

tests/mocks/JWT.php
```diff
@@ -0,0 +1,28 @@
...
+{
+  public static function encode()
+  {
+    $args = func_get_args();
+    $method = 'JWT';
```
tmatsuo: Nit: Isn't it rather a `class`?

bshaffer: well I'm not a professional developer but... oh crap, yes I am!!! Better fix that one before my boss sees it.

src/OAuth2.php
bshaffer: see main comment thread, but yes there are tests.

tmatsuo: I see. Seems like I somehow misunderstood how namespace works.

src/OAuth2.php
tmatsuo: I don't think you need to do that since we have composer requirements for them.

tmatsuo:
LGTM

src/OAuth2.php
bshaffer: yeah, I was thinking of the case where someone opts out of using composer and clones the library directly.

kesar:
Its now a big refactoring but probably oAuth class shouldn't contain so much logic related to JWT. It should have injected in the constructor an object with an interface JWTInterface that has methods decode and encode. That would also help with UT because you could mock easily all the logic related to JWT and test it properly in a different set of tests.

bshaffer:
@kesar I agree, this is one reason why I hate static methods.

PR #84, opened 2015-12-03
add support for guzzle 6

dwsupplee:
Notable changes:

- Organized classes under new namespaces (ex. ```Google\Auth\GCECredentials``` is now found under the ```Google\Auth\Credentials\GCECredentials``` namespace
- Renamed 'getFetcher' to 'getSubscriber'
```
use Google\Auth\ApplicationDefaultCredentials;

ApplicationDefaultCredentials::getFetcher() // no-mo
ApplicationDefaultCredentials::getSubscriber() // +1
```
- Added guzzle 6 middleware implementations 
```
use Google\Auth\ApplicationDefaultCredentials;
use GuzzleHttp\Client;
use GuzzleHttp\HandlerStack;

$middleware = ApplicationDefaultCredentials::getMiddleware(
    'https://www.googleapis.com/auth/taskqueue'
);
$stack = HandlerStack::create();
$stack->push($middleware);

$client = new Client([
    'handler' => $stack,
    'base_uri' => 'https://www.googleapis.com/taskqueue/v1beta2/projects/',
    'auth' => 'google_auth' // authorize all requests
]);

$res = $client->get('myproject/taskqueues/myqueue');
```
- Support for guzzle 5/6 as the default transport layer - however you can now also define a custom layer as follows:
```
use Google\Auth\Credentials\GCECredentials;
use GuzzleHttp\Psr7\Response;
use Psr\Http\Message\RequestInterface;

$isOnGCE = GCECredentials::onGce(function (RequestInterface $request, $options = []) {
    // define your own http implementation which returns a PSR7 response
    return Response(200);
});
```

The code is functional but please keep in mind there are still a few ```@todos``` which need to be addressed.

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -27,7 +27,7 @@ use GuzzleHttp\Psr7\Request;
 /**
  * GCECredentials supports authorization on Google Compute Engine.
  *
- * It can be used to authorize requests using the AuthTokenFetcher, but will
+ * It can be used to authorize requests using the AuthTokenMiddleware, but will
  * only succeed if being run on GCE:
  *
  *   use Google\Auth\Credentials\GCECredentials;
@@ -152,7 +152,13 @@ class GCECredentials extends CredentialsLoader
       )
     );
     $body = (string) $resp->getBody();
-    return json_decode($body, true);
+
+    // Assume it's JSON; if it's not throw an exception
+    if (null === $json = json_decode($body, true)) {
+      throw new \Exception('Invalid JSON response');
+    }
+
+    return $json;
   }
 
   /**
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -15,14 +15,14 @@
  * limitations under the License.
  */
 
-namespace Google\Auth;
+namespace Google\Auth\Credentials;
 
-use GuzzleHttp\ClientInterface;
-use GuzzleHttp\Client;
-use GuzzleHttp\Stream\Stream;
+use Google\Auth\CredentialsLoader;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
 use GuzzleHttp\Exception\ClientException;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Exception\ServerException;
+use GuzzleHttp\Psr7\Request;
 
 /**
  * GCECredentials supports authorization on Google Compute Engine.
@@ -30,17 +30,22 @@ use GuzzleHttp\Exception\ServerException;
  * It can be used to authorize requests using the AuthTokenFetcher, but will
  * only succeed if being run on GCE:
  *
+ *   use Google\Auth\Credentials\GCECredentials;
+ *   use Google\Auth\Middleware\AuthTokenMiddleware;
  *   use GuzzleHttp\Client;
- *   use Google\Auth\GCECredentials;
- *   use Google\Auth\AuthTokenFetcher;
+ *   use GuzzleHttp\HandlerStack;
  *
  *   $gce = new GCECredentials();
- *   $scoped = new AuthTokenFetcher($gce);
+ *   $middleware = new AuthTokenMiddleware($gce);
+ *   $stack = HandlerStack::create();
+ *   $stack->push($middleware);
+ *
  *   $client = new Client([
- *      'base_url' => 'https://www.googleapis.com/taskqueue/v1beta2/projects/',
- *      'defaults' => ['auth' => 'google_auth']
+ *      'handler' => $stack,
+ *      'base_uri' => 'https://www.googleapis.com/taskqueue/v1beta2/projects/',
+ *      'auth' => 'google_auth'
  *   ]);
- *   $client->getEmitter()->attach($gce);
+ *
  *   $res = $client->get('myproject/taskqueues/myqueue');
  */
 class GCECredentials extends CredentialsLoader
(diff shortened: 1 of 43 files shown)


dwsupplee:
I signed it!

Nyholm:
@dwsupplee Have you seen #81?

dwsupplee:
I have! HTTP Plug is pretty interesting, it is definitely something to keep an eye on as it evolves and gains traction with the community. Keep me posted! :)

src/Credentials/GCECredentials.php
```diff
@@ -105,8 +110,11 @@ public static function onGce(ClientInterface $client = null)
       // could lead to false negatives in the event that we are on GCE, but
       // the metadata resolution was particularly slow. The latter case is
       // "unlikely".
-      $resp = $client->get($checkUri, ['timeout' => 0.3]);
-      return $resp->getHeader(self::FLAVOR_HEADER) == 'Google';
+      $resp = $httpHandler(
+        new Request('GET', $checkUri),
+        ['timeout' => 0.3]
```
bshaffer: What defines the valid `options` for http handlers?

dwsupplee: That's a great question.

When using guzzle the options are:
http://docs.guzzlephp.org/en/5.3/clients.html#request-options
http://docs.guzzlephp.org/en/latest/request-options.html

But that doesn't really help someone who defines their own implementation. Maybe we should come up with a whitelist of valid options or at the least document the options that are required to be implemented such as `timeout` above?

bshaffer: We can just throw it in the `README` in a section **Creating Custom Http Handlers** and then just mention the options used. At this time, I believe this is the only instance of the library using the options anyway.

bshaffer:
@Nyholm I imagine since this is PSR-7 compatible and uses middlewares, PHP-HTTP will be able to play nicely?

Nyholm:
This PR replicates half of the things I've done in #81. 

> @Nyholm I imagine since this is PSR-7 compatible and uses middlewares, PHP-HTTP will be able to play nicely?

Im not too sure you understand ~~PHP-HTTP~~ (sorry, "Httplug". PHP-HTTP is the organisation). We want our code to be PSR-7 because it's great and a community standard. That is something we should do and I'm sure that you agree. 

Guzzle on the other hand, is also great but it is not a standard. (Though, it is popular) By using the power of PSR-7 we can remove the dependency from Guzzle. Which is good because you should not rely on a implementation, that's not SOLID. 

So, we need an interface to rely on that Guzzle implements, say a `PSR7RequestSenderInterface`. That is Httplug.

dwsupplee:
@Nyholm 

>Im not too sure you understand PHP-HTTP (sorry, "Httplug". PHP-HTTP is the organisation). We want our code to be PSR-7 because it's great and a community standard. That is something we should do and I'm sure that you agree.

I think we understand what HTTP Plug is trying to achieve - the issue is it seems we are not comfortable relying on the implementation at this point in time. 

>Guzzle on the other hand, is also great but it is not a standard. (Though, it is popular) By using the power of PSR-7 we can remove the dependency from Guzzle. Which is good because you should not rely on a implementation, that's not SOLID.

This code does adhere to PSR-7 and you are able to override the transport layer with whatever implementation you choose (including php-http's!):

```php
use Google\Auth\Credentials\GCECredentials;
use Http\Adapter\Guzzle5HttpAdapter;
use Psr\Http\Message\RequestInterface;

$adapter = new Guzzle5HttpAdapter();

$isOnGCE = GCECredentials::onGce([$adapter, 'sendRequest']);
```

>So, we need an interface to rely on that Guzzle implements, say a PSR7RequestSenderInterface. That is Httplug.

HttpPlug is _one way_ of achieving this goal. I genuinely appreciate where you guys are coming from and am interested to see where it goes.

bshaffer:
This is now merged, but because I rebased, it doesn't realize it's merged.

sagikazarmark:
Hey guys, HTTPlug developer here.

Although I am not a fan of passing around callables, I have to admit it works and usually provides great flexibility (which is not always an advantage), like in the case above. I like the implementation, but to be completely fair, I would like to correct some things about our interface:

It is HTTPlug, not HTTP Plug. We played with the words a bit. :smile: 

> the issue is it seems we are not comfortable relying on the implementation at this point in time. 

HTTPlug is a set of interfaces, there is no implementation in it. We PROVIDE some implementations. I would be happy to hear some feedback: are there any actual issues you see or just felt that the callable/middleware way is more flexible (which is possibly true)?

From google/google-auth-library-php#81

> mostly because it has a smaller dependency tree

We are talking about one dependency: the contract package itself. No other dependencies are required for consumer packages. I guess you meant actual implementations, which are indeed not just single dependencies, but neither Guzzle is.

bshaffer:
Thank you for the info and clarification @sagikazarmark. Really this comes down to the amount of time to work on this, and the demand of the community. 

Our main goal is to get this working and useful. Guzzle6/PSR7 are the fastest ways to accomplish this. Also, I'm sure you'll agree, PR #81 would have required a tremendous amount of work to get into shape. The tests were failing and even smaller things like whitespace issues made it not a viable option.

That being said, this is a community project, and we can open an issue for HTTPlug support, to see how many `+1`'s we receive. I really appreciate the work being done by you guys, and I hope to incorporate your library into this one in the future.

sagikazarmark:
@bshaffer it wasn't my intention to promote the library or force using it. I don't believe it is a standard (although some PSR could be depend on it), and I definitely don't think there is only one possible solution for problems (that's one of the idea of HTTPlug actually).

Just wanted to clear some things, which I believed to be easily misunderstandable.

I agree that at the current state of the library it requires a big amount of work to make it work, and the interfaces are not 100% stable yet, so any code depending on it should not be considered production ready.

PR #85, opened 2015-12-04
Add support for guzzle 6

bshaffer:


dwsupplee:
Good with me :)

dwsupplee:
Okay with me

bshaffer:
@tbetbetbe this is a large change, I would love another pair of eyes on it

.travis.yml
```diff
@@ -3,18 +3,20 @@ language: php
 sudo: false
 
 php:
-  - 5.4
```
bshaffer: We should keep `5.4` compatibility. Yes it's EOL'd, but only recently. And many customers still use it.

dwsupplee: Guzzle 6 requires >= 5.5, I can add an exclusion for those requirements if that sounds good.

bshaffer: We will still support `5.4` for Guzzle 5

bshaffer:
@tbetbetbe @tmatsuo @sjlangley I'll be merging this by EOD unless there are any concerns

tbetbetbe:
LGTM

src/Credentials/GCECredentials.php
```diff
@@ -120,25 +128,37 @@ public static function onGce(ClientInterface $client = null)
...
+        'GET',
+        self::getTokenUri(),
+        [self::FLAVOR_HEADER => 'Google']
+      )
+    );
+    $body = (string) $resp->getBody();
+
+    // Assume it's JSON; if it's not throw an exception
+    if (null === $json = json_decode($body, true)) {
+      throw new \Exception('Invalid JSON response');
```
tmatsuo: Is it possible to give a hint for debugging when someone gets this exception?

tests/BaseTest.php
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  public function onlyGuzzle5()
+  {
+    $version = ClientInterface::VERSION;
+    if ('5' !== $version[0]) {
+      $this->markTestSkipped('Guzzle 5 only');
+    }
+  }
+}
```
tmatsuo: nit: add newline?

tmatsuo:
Looks good except few minor nits

tests/bootstrap.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2014 Google Inc.
+ * Copyright 2015 Google Inc.
```
tmatsuo: I don't think you should update the year

bshaffer: why not?

tmatsuo: It should stay at the year when it was first written.
http://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated

bshaffer: Oh! Good to know!

Well this was actually written in `2015` and I just forgot to update it when I first committed it.

tmatsuo: Aha, I confirmed the first commit of this file was `2015` and then that's fine.

src/Credentials/GCECredentials.php
bshaffer: yes, using [`json_last_error`](http://php.net/manual/en/function.json-last-error.php), but it's not super straightforward

PR #87, opened 2015-12-15
Allow use of Guzzle 5.3 too

DavidPrevot:
Nothing in the current code seems to prevent the use of Guzzle 5.3.

Of course, this PR is superseded by https://github.com/google/google-auth-library-php/pull/85.


bshaffer:
This is very true! However, this will be merged in as part of #85 later today.

Thank you for your contribution!

PR #88, opened 2015-12-16
Use PHP_OS constants instead of php_uname().

tmatsuo:
According to:
http://php.net/manual/en/function.php-uname.php
It is recommended using PHP_OS if possible.


bshaffer:
This seems great, although I can't find the version constraints of `PHP_OS`. If the tests all pass though, this is LGTM'd!

tmatsuo:
Thanks, merging

PR #89, opened 2015-12-16
Removed dev-master from README.

tmatsuo:
Since we have releases, can we remove dev-master from the README?


tmatsuo:
@bshaffer OK to merge?

PR #93, opened 2016-02-23
misc fixes - usability and doc fix

bshaffer:


bshaffer:
@tbetbetbe main change here is that if someone sets a code but forgets to set their redirect URI, they get an ambiguous `"grant_type" is required` error. This way, they'll get the proper `"redirect_uri" required` error. 

I am not sure if this decision was made intentionally however. I cannot think of a case where the authorization code would be set and the grant type should *not* be `code`.

PR #95, opened 2016-02-26
Fixing some service account flow JWT issues.

### LINKED ISSUE #94: OAuth2 class seems to have some bugs with service account JWT flow.
(1) The DEFAULT_EXPIRY_MINUTES is actually used as seconds, so it should be renamed to _SECONDS and updated to 3600:
https://github.com/google/google-auth-library-php/blob/master/src/OAuth2.php#L38

(2) According to this doc:
https://developers.google.com/identity/protocols/OAuth2ServiceAccount
"prn" is no longer used and "scope" is required. So some of the logic here:
https://github.com/google/google-auth-library-php/blob/master/src/OAuth2.php#L366

needs to be updated.

Can you confirm this is the case. I can submit a pull request for this. Thanks.

vtsao:
- Fixed default expiry time to be in seconds, also renamed skew variable for clarity.
 - Fixed documentation on service accounts creds class.
 - Removed obsolete 'prn' JWT claim and all references to it.
 - Updated `CredentialsLoader::TOKEN_CREDENTIAL_URI` to v4.

Fixes #94.

### DIFF
--- a/src/Credentials/ServiceAccountCredentials.php
+++ b/src/Credentials/ServiceAccountCredentials.php
@@ -37,12 +37,10 @@ use GuzzleHttp\Psr7;
  *   use Google\Auth\Middleware\AuthTokenMiddleware;
  *   use GuzzleHttp\Client;
  *   use GuzzleHttp\HandlerStack;
- *   use GuzzleHttp\Psr7;
  *
- *   $stream = Psr7\stream_for(file_get_contents(<my_key_file>));
  *   $sa = new ServiceAccountCredentials(
  *       'https://www.googleapis.com/auth/taskqueue',
- *       $stream
+ *       '/path/to/your/json/key_file.json'
  *   );
  *   $middleware = new AuthTokenMiddleware($sa);
  *   $stack = HandlerStack::create();
@@ -64,10 +62,8 @@ class ServiceAccountCredentials extends CredentialsLoader
    * @param string|array $scope the scope of the access request, expressed
    *   either as an Array or as a space-delimited String.
    *
-   * @param array $jsonKey JSON credentials.
-   *
-   * @param string $jsonKeyPath the path to a file containing JSON credentials.  If
-   *   jsonKeyStream is set, it is ignored.
+   * @param string|array $jsonKey JSON credential file path or JSON credentials
+   *   as an associative array
    *
    * @param string $sub an email address account to impersonate, in situations when
    *   the service account has been delegated domain wide access.
@@ -75,12 +71,16 @@ class ServiceAccountCredentials extends CredentialsLoader
   public function __construct(
     $scope,
     $jsonKey,
-    $jsonKeyPath = null,
     $sub = null
   ) {
-    if (is_null($jsonKey)) {
-      $jsonKeyStream = Psr7\stream_for(file_get_contents($jsonKeyPath));
-      $jsonKey = json_decode($jsonKeyStream->getContents(), true);
+    if (is_string($jsonKey)) {
+      if (!file_exists($jsonKey)) {
+        throw new \InvalidArgumentException('file does not exist');
+      }
+      $jsonKeyStream = Psr7\stream_for(file_get_contents($jsonKey));
+      if (!$jsonKey = json_decode($jsonKeyStream, true)) {
+        throw new \LogicException('invalid json for auth config');
+      }
     }
     if (!array_key_exists('client_email', $jsonKey)) {
       throw new \InvalidArgumentException(
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -28,7 +28,7 @@ use Psr\Http\Message\StreamInterface;
  */
 abstract class CredentialsLoader implements FetchAuthTokenInterface
 {
-  const TOKEN_CREDENTIAL_URI = 'https://www.googleapis.com/oauth2/v3/token';
+  const TOKEN_CREDENTIAL_URI = 'https://www.googleapis.com/oauth2/v4/token';
   const ENV_VAR = 'GOOGLE_APPLICATION_CREDENTIALS';
   const WELL_KNOWN_PATH = 'gcloud/application_default_credentials.json';
   const NON_WINDOWS_WELL_KNOWN_PATH_BASE = '.config';
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -34,9 +34,8 @@ use Psr\Http\Message\UriInterface;
  */
 class OAuth2 implements FetchAuthTokenInterface
 {
-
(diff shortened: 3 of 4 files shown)


src/CredentialsLoader.php
```diff
@@ -169,4 +169,13 @@ public function updateMetadata(
     $metadata_copy[self::AUTH_METADATA_KEY] = array('Bearer ' . $result['access_token']);
     return $metadata_copy;
   }
+
+  /**
+   * Gets the underlying `OAuth2` instance used to conduct authorization.
+   *
+   * @return OAuth2
+   */
+  public function getAuth() {
```
bshaffer: I'm not a huge fan of exposing this unless it's 100% necessary. I think the main issue here is there's no way to get at your access token without using `Cache` classes, yes?

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -37,12 +37,11 @@
...
  *   use GuzzleHttp\Client;
  *   use GuzzleHttp\HandlerStack;
- *   use GuzzleHttp\Psr7;
  *
- *   $stream = Psr7\stream_for(file_get_contents(<my_key_file>));
  *   $sa = new ServiceAccountCredentials(
  *       'https://www.googleapis.com/auth/taskqueue',
- *       $stream
+ *       null,
+ *       '/path/to/your/json/key_file.json'
```
bshaffer: Good catch. I think this is better. Although we'd be better off combining argument 2 and 3 of `ServiceAccountCredentials` like we do [here](https://github.com/google/google-api-php-client/blob/master/src/Google/Client.php#L816)

vtsao: Sure, I can do that.

src/CredentialsLoader.php
vtsao: I'm not a huge fan either.

But, the main issue is that if we don't or can't use the middleware classes to help manage our access tokens (we're using SOAPClient), we can't use the getExpiresIn() methods in order to deal with when we need to refresh the access token for the user. The TODOs on the cache usage in the middleware classes also seem to not deal with this properly right now right?

I can always store the access token and calculate this myself, but seems like since the logic is already available, I should be using it from the library.

I also want a common object that abstracts away the underlying flow that's being used, which is why I wanted to use the OAuth2 object. CredentialsLoader could also be that object, except there is an implicit contract for (most of the) cred loader subclasses to impl fetchAuthToken().

I could also expose a method called getExpiresIn() in the cred loaders that I care about that delegates to the OAuth2->getExpiresIn() method as well as another solution. But this wouldn't be defined in the parent or an interface, which makes it a bit risky to assume it's always on a cred loader.

bshaffer: This is a tough one. Currently the way to do this (and I'm not saying it's good) would be to send in a `CacheInterface` object. Any access token fetched would then be stored in the cache object (although the methods to retrieve it require `getCacheKey`, which is private).

In the [README](https://github.com/google/google-auth-library-php#call-the-apis) workflow, we don't ever obtain the `CredentialsLoader` object. It might make more sense to have a method on the `Middleware` classes to get the current access token from the cache. The cache would then need to be updated to set not just the token, but the expiry, etc.

Or we can just expose `OAuth2`, but again this would require you to call `ApplicationDefaultCredentials::getCredentials` and then create the Middleware or Subscriber yourself. Also it exposes the library's class structure, which shouldn't really be necessary.

src/Credentials/ServiceAccountCredentials.php
vtsao: This is done.

src/CredentialsLoader.php
vtsao: For my use case, I don’t want to and should not use the middleware classes. I also realized my statement above about `fetchAuthToken()` being an implicit contract is incorrect since I forgot that cred loader impls `FetchAuthTokenInterface`. I also realized that `getExpiresAt()` is actually the method I want, since it calculates the expiry time whereas `getExpiresIn()` doesn’t (it should probably be documented that it’s a snapshot of the token metadata when the token was first received). There’s also a bug I fixed with `getExpiresAt()` in `updateToken()` being overriden to `null`.

Anyways, what I propose for now is we implement `getExpiresAt()` on the `FetchAuthTokenInterface` interface, since in `AuthTokenMIddleware’s` `fetchToken()` method, you’re going to need to get the expiry time of the token if you want to set it in the cache. I know the TODO’s say that this should be read from the access token array returned from `fetchAuthToken()` and set in the cache, but that means you’d be calculating the expiry time again, since the logic is already available in the `OAuth2` class. So how does exposing `getExpiresAt()` on `FetchAuthTokenInterface` sound? All the subclasses will just delegate this to the underlying `OAuth2` instance anyways. I can do this in another pull request (or this one) - up to you.

This way we don’t need to expose the underlying `OAuth2` instance on cred loader.

Also, I think the larger issue here is that we don’t have an access token class. The access token itself is not stored in the `OAuth2` object and only stored if you provide a cache. But the metadata (like expiry time) for the token is stored in the `OAuth2` object - so it’s decoupled. An access token class should probably be created and this metadata moved there?

bshaffer: Yeah, this makes sense. I really like this solution. Thank you for taking the time to look into this.

vtsao: Thanks! Do you care if I do it in this pull request or a new one?

bshaffer: Whichever is easiest!

vtsao: I'll do it in a follow-up CL so it's cleaner. Filed #96 to track this.

vtsao:
This CL should be ready now, PTAL.

FYI I also updated `CredentialsLoader::TOKEN_CREDENTIAL_URI` to v4.

src/CredentialsLoader.php
```diff
@@ -28,7 +28,7 @@
  */
 abstract class CredentialsLoader implements FetchAuthTokenInterface
 {
-  const TOKEN_CREDENTIAL_URI = 'https://www.googleapis.com/oauth2/v3/token';
+  const TOKEN_CREDENTIAL_URI = 'https://www.googleapis.com/oauth2/v4/token';
```
bshaffer: What implications does this have? I don't feel comfortable blindly updating this API version.

vtsao: All our docs now use v4 as the token URL:
https://developers.google.com/identity/protocols/OAuth2InstalledApp

but unsure what the difference is, so was updating this to reflect our docs.

bshaffer: I'd prefer to keep the URL unchanged until we know what the implications are.

bshaffer:
merged in 965e6fb

PR #97, opened 2016-03-02
Exposing getExpiresAt() and fixing some other minor issues.

### LINKED ISSUE #96: 'code' documentation is incorrect for OAuth2 constructor config
The `OAuth2` constructor config documents it accepts `code`, which it doesn't.

vtsao:
- Removed documentation for 'code' as a config option in OAuth2 constructor, since it's something you will set later on in the authorization flow and probably not something you construct the OAuth2 object with (it wasn't being read anyways).
- Merged jsonKey and jsonKeyPath constructor args into one in UserRefreshCredentials.

Fixes #96.

### DIFF
--- a/src/Credentials/ServiceAccountCredentials.php
+++ b/src/Credentials/ServiceAccountCredentials.php
@@ -62,8 +62,6 @@ class ServiceAccountCredentials extends CredentialsLoader
    * @param string|array $scope the scope of the access request, expressed
    *   either as an Array or as a space-delimited String.
    *
-   * @param array $jsonKey JSON credentials.
-   *
    * @param string|array $jsonKey JSON credential file path or JSON credentials
    *   as an associative array
    *
--- a/src/Credentials/ServiceAccountCredentials.php
+++ b/src/Credentials/ServiceAccountCredentials.php
@@ -37,12 +37,10 @@ use GuzzleHttp\Psr7;
  *   use Google\Auth\Middleware\AuthTokenMiddleware;
  *   use GuzzleHttp\Client;
  *   use GuzzleHttp\HandlerStack;
- *   use GuzzleHttp\Psr7;
  *
- *   $stream = Psr7\stream_for(file_get_contents(<my_key_file>));
  *   $sa = new ServiceAccountCredentials(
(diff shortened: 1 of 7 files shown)


vtsao:
Uh, guess I can't diffbase things on GitHub...

bshaffer:
Thank you for this @vtsao !

I have looked a little deeper at this, and there is a problem with `AppIdentityCredentials` and `GCECredentials`, since neither of them create the `OAuth2` object.

possible solutions:
  1. Remove the method from `CredentialsLoader` and add it to each of the classes which have an `$oauth` property. Then add the equivalent methods to the other two classes
  2. Instead of `getExpiresAt` method, have a `getLastReceivedToken` method on the credentials classes  which returns a standard associative array, including `expires`

I think I like 2, and I can implement it (with tests). thoughts?

vtsao:
I noticed that too, but I feel like having the `auth` object stored in `CredentialsLoader` in the first place is then a larger design problem as well if not every subclass will use it.

So `getLastReceivedToken()` will have the last received access token and the expires then (does this conflict with the caching paradigm)? And it will be a method on `FetchAuthTokenInterface`? As long as it still fills in the `expires` field using `OAuth2->getExpiresAt()` where appropriate (so we're re-using existing logic), I'm fine with that.

To clarify, you want to take on this task? I'm fine with that (and prefer that) if you have the cycles to do this soon, since I'm blocked on this for one of my projects.

bshaffer:
Yes, I'm working on it. And yes, you're right about the oauth being stored in CredentialsLoader. I will move that out of there.

vtsao:
Okay, SGTM. Then I'll remove that part from this CL and just have this CL fix the other two minor issues I mentioned.

bshaffer:
Thank you! Sorry for the fire drill. I really want to get this right.

vtsao:
No worries, can you accept:
https://github.com/google/google-auth-library-php/pull/95

if it's all good.

and then I'll merge into this one and then you can review this one again.

bshaffer:
closing in favor of #95

vtsao:
There are some minor issues that I've fixed in this pull request, we should still submit this.

PR #98, opened 2016-03-02
adds getLastReceivedToken and tests

bshaffer:


bshaffer:
see #97 @vtsao PTAL

bshaffer:
These tests are failing because we need to merge #95 first.

src/Credentials/AppIdentityCredentials.php
```diff
@@ -104,11 +112,27 @@ public function fetchAuthToken(callable $httpHandler = null)
...
   }
 
   /**
+   * Implements FetchAuthTokenInterface#getLastReceivedToken.
+   */
+  public function getLastReceivedToken()
+  {
+    if ($this->lastReceivedToken) {
+      return [
+        'access_token' => $this->lastReceivedToken['access_token'],
```
vtsao: How does this work with our caching paradigm? Since we're essentially caching this now. I guess it's alright?

src/Credentials/AppIdentityCredentials.php
```diff
@@ -104,11 +112,27 @@ public function fetchAuthToken(callable $httpHandler = null)
...
+        'expires' => $this->lastReceivedToken['expiration_time'],
```
vtsao: Can we name this expires_at so it's explicit.

src/FetchAuthTokenInterface.php
```diff
@@ -40,4 +40,17 @@ public function fetchAuthToken(callable $httpHandler = null);
   * @return string a key that may be used to cache the auth token.
   */
   public function getCacheKey();
+
+  /**
+   * Returns an associative array with the token and expiration time
```
vtsao: Might want to make clear that it's the expiration time in seconds since the Unix epoch.

tests/FetchAuthTokenTest.php
```diff
@@ -0,0 +1,137 @@
...
+
+    $this->assertEquals('xyz', $accessToken['access_token']);
+    $this->assertEquals(strtotime('2001'), $accessToken['expires']);
+  }
+
+  public function provideAuthTokenFetcher()
+  {
+    $scopes = ['https://www.googleapis.com/auth/drive.readonly'];
+    $jsonPath = __DIR__ . '/fixtures/.config/' . CredentialsLoader::WELL_KNOWN_PATH;
+    $jsonPath2 = __DIR__ . '/fixtures2/.config/' . CredentialsLoader::WELL_KNOWN_PATH;
```
vtsao: fix line length

tests/FetchAuthTokenTest.php
```diff
@@ -0,0 +1,137 @@
...
+      'access_token' => 'xyz',
+      'expiration_time' => strtotime('2001'),
+    ]);
+
+    return $credentials;
+  }
+
+  private function getServiceAccountCredentials($scopes, $jsonPath)
+  {
+    $class = new \ReflectionClass('Google\Auth\Credentials\ServiceAccountCredentials');
```
vtsao: Not sure what the character limit you guys are using for your project is, so not sure if this line length is okay or needs to be fixed.

vtsao:
Added comments, looks great! Thanks for doing this so quickly.

src/Credentials/AppIdentityCredentials.php
bshaffer: The caching happens outside this cache, at the Middleware/Subscriber layer. So it won't be affected.

src/Credentials/AppIdentityCredentials.php
bshaffer: Yeah, I agree. I thought the [oauth spec](https://tools.ietf.org/html/rfc6749) used "expires", but I can find no examples of this.

src/FetchAuthTokenInterface.php
```diff
@@ -40,4 +40,17 @@ public function fetchAuthToken(callable $httpHandler = null);
...
+   *
+   * @return array|null the last received token
+   *  array(2) {
+   *   ["access_token"]=>
+   *   string(3) "xyz"
+   *   ["expires"]=>
```
vtsao: expires_at

src/Credentials/AppIdentityCredentials.php
bshaffer: To make this more clear, no it won't be cached. The caching happens with the result coming directly from `fetchAuthToken`.

src/Credentials/UserRefreshCredentials.php
```diff
@@ -42,17 +46,21 @@ class UserRefreshCredentials extends CredentialsLoader
    *
    * @param array $jsonKey JSON credentials.
```
vtsao: remove duplicate @param

src/Credentials/UserRefreshCredentials.php
```diff
@@ -42,17 +46,21 @@ class UserRefreshCredentials extends CredentialsLoader
...
    *
-   * @param string $jsonKeyPath the path to a file containing JSON credentials.  If
-   *   jsonKey is set, it is ignored.
+   * @param string|array $jsonKey JSON credential file path or JSON credentials
+   *   as an associative array
    */
   public function __construct(
     $scope,
```
vtsao: Also, as an FYI scope is not used to refresh the access token, but seems like you guys are using it for the cache key. Not sure if there's a different way you want to do this, since for people not using the cache stuff, they'd just be passing in `null` for the scope, or if they pass in something, the scope is a no-op.

bshaffer: You're right. It's just for the cache key. The caching implementation in this library could use some consideration, since cache keys would be the same for multiple users with the same scope.

PR #105, opened 2016-04-12
Add doc-blocks to OAuth2 where appropriate to allow better analysis

cedricziel:
With PHP IDEs being able to provide smart assistance and context 
for code, it's necessary to add soft type hints through doc blocks.

This also helps humans when inspecting the library code visually.

Adds appropriate @var, @return and @param tags where possible.

I propose to do this gradually or the whole library.

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -17,10 +17,10 @@
 
 namespace Google\Auth;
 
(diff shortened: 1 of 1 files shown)


tbetbetbe:
LGTM

PR #108, opened 2016-05-05
use PSR-6 for caching

bshaffer:
addresses https://github.com/google/google-api-php-client/issues/941


bshaffer:
@tmatsuo PTAL

bshaffer:
According to the requirements of `key` in [the spec](https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-6-cache.md#definitions), we will need to base64-encode our URLs to make the keys safe (i.e. we are using illegal characters in the scope urls)

tmatsuo:
LGTM

bshaffer:
@stanley-cheung as you implemented the original `CacheInterface`, PTAL.

src/CacheTrait.php
```diff
@@ -62,7 +62,17 @@ private function setCachedValue($v)
...
-        $this->cache->set($key, $v);
+
+        $key = self::getValidKeyName($this->cacheConfig['prefix'].$fetcherKey);
+        $cacheItem = $this->cache->getItem($key);
+        $cacheItem->set($v);
+        $cacheItem->expiresAfter($this->cacheConfig['lifetime']);
+        return $this->cache->save($cacheItem);
+    }
+
+    public static function getValidKeyName($key)
```
kesar: why is this static?

stanley-cheung:
LGTM. Using PSR-6' `CacheItemPoolInterface` is a great idea, rather than rolling our own `CacheInterface`. Just FYI, I wasn't the original implementor of `CacheInterface`. That even predates me :P

src/CacheTrait.php
bshaffer: @kesar you're right, this function's visibility should match that of `getCachedValue` and `setCachedValue`

bshaffer:
@stanley-cheung  `CacheInterface` has now been banished back to the mysterious origins from whence it came.

PR #109, opened 2016-05-10
simplify travis matrix / composer using prefer-lowest flag

bshaffer:


tmatsuo:
The change itself looks good, but phplint is complaining...

bshaffer:
@tmatsuo PTAL - I ran `.php_cs` and added it to tests, and also upgraded the minimum guzzle version to `5.3` to fix failing tests on PHP 7.

tmatsuo:
LGTM

PR #110, opened 2016-05-10
adds token callback to AuthToken Middleware & Subscriber

bshaffer:
This is an attempt to resolve https://github.com/google/google-api-php-client/issues/901 and provide a more elegant way for https://github.com/google/google-api-php-client/pull/905 and https://github.com/google/google-api-php-client/pull/925


bshaffer:
@tmatsuo @SurferJeffAtGoogle PTAL

tmatsuo:
LGTM

PR #112, opened 2016-05-26
Expose keyfile

jdpedrie:
Hello!

This pull request is part of a solution for GoogleCloudPlatform/gcloud-php#32.

The goal of this change is to allow access to json key data even in cases when a key is not explicitly given to the credentials loader. So when a key is loaded from the `GOOGLE_APPLICATION_CREDENTIALS` environment variable, or from the well-known path, applications using this library will have access to that information. In the case of gcloud-php, we will use it to attempt to determine the project ID even when one is not explicitly provided.

### DIFF
--- a/src/Credentials/ServiceAccountCredentials.php
+++ b/src/Credentials/ServiceAccountCredentials.php
@@ -103,6 +103,8 @@ class ServiceAccountCredentials extends CredentialsLoader
             'sub' => $sub,
             'tokenCredentialUri' => self::TOKEN_CREDENTIAL_URI,
         ]);
+
+        $this->jsonKey = $jsonKey;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


jdpedrie:
hey @bshaffer, any chance you could take a look at this in the next couple of days? Thanks!

bshaffer:
@jdpedrie what if we made the existing methods more generic? Then you could just call those instead of saving/exposing it on the `credentials` subclasses?

```php
use Google\Auth\CredentialsLoader;

$jsonKey = CredentialsLoader::fromEnv() ?: CredentialsLoader::fromWellKnownFile();
```

This would break BC, but it's worth it for a better API.

jdpedrie:
I'm a fan of that! I tried to avoid BC breaks in my change, but if that's not an overriding concern I definitely prefer your solution.

bshaffer:
@jdpedrie I appreciate that! I think this is a worthwhile break, however.

PR #118, opened 2016-06-17
explicitly set AppEngine Flexible to use GCE credentials

bshaffer:
If a custom runtime is used in App Engine Flexible, or PHP is executing outside of `php-nginx` (in this environment, `SERVER_SOFTWARE` is [set explicitly](https://github.com/GoogleCloudPlatform/php-docker/blob/2be08b74359c6d5bfd54c058e86ada19b4bfb94b/php-nginx/fastcgi_params#L30)), the auth library tries to use `AppIdentityCredentials`, which fails because the [appengine php sdk](https://github.com/GoogleCloudPlatform/appengine-php-sdk) doesn't exist.

### DIFF
--- a/tests/ApplicationDefaultCredentialsTest.php
+++ b/tests/ApplicationDefaultCredentialsTest.php
@@ -175,6 +175,52 @@ class ADCGetMiddlewareTest extends \PHPUnit_Framework_TestCase
     }
 }
 
+class ADCGetCredentialsAppEngineTest extends BaseTest
+{
+    private $originalHome;
+    private $originalServiceAccount;
+
+    protected function setUp()
+    {
+        // set home to be somewhere else
+        $this->originalHome = getenv('HOME');
+        putenv('HOME=' . __DIR__ . '/not_exist_fixtures');
+
+        // remove service account path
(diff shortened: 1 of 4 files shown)


tests/ApplicationDefaultCredentialsTest.php
```diff
@@ -175,6 +175,52 @@ public function testSuccedsIfNoDefaultFilesButIsOnGCE()
...
+        $this->assertInstanceOf(
+            'Google\Auth\Credentials\AppIdentityCredentials',
+            ApplicationDefaultCredentials::getCredentials()
+        );
+    }
+
+    public function testAppEngineFlexible()
+    {
+        $_SERVER['SERVER_SOFTWARE'] = 'Google App Engine';
+        $_SERVER['GAE_VM'] = 'true';
```
tmatsuo: small nit: does it affect other tests?

tmatsuo:
LGTM with a minor nit, but please submit

tests/ApplicationDefaultCredentialsTest.php
bshaffer: The `$_SERVER` variable is reset for each test by phpunit, so this is still hermetic!

Issue #121, opened 2016-07-18
Implement Caching on a Credentials level


bshaffer:
Caching could be implemented on a Credentials level instead of a Middleware/Subscriber level.

PROS:
 - Caching happens closer to the actual token fetching, which will allow more libraries to take advantage of caching hands-free (see https://github.com/googleapis/gax-php/issues/26)
 - Straightforward refactoring

CONS:
 - Breaking BC (again)

`ScopedAccessToken` subscriber/middleware has caching. We could leave the caching logic in there.

tmatsuo:
I think I like it

bshaffer:
implemented in #123

PR #123, opened 2016-07-19
moves caching to wrapper class 'FetchAuthTokenCache' and adds Memory cache

bshaffer:
Addresses #121 

Highlights:
- Adds `FetchAuthTokenCache` wrapper class to implement caching on anything implementing `FetchAuthTokenInterface`.
- Removes caching from all Middleware / Subscriber classes
- Bypasses caching (by setting the key to null) for `AppIdentityCredentials` - Caching here is handled by the [AppIdentityService](https://github.com/GoogleCloudPlatform/appengine-php-sdk/blob/master/google/appengine/api/app_identity/AppIdentityService.php#L168)
- Removes calls to `getCacheKey` from Cache Trait - that function does not exist as part of the trait, and as a result was causing some unexpected problems. The cache key is now explicitly passed in.

### DIFF
--- a/src/FetchAuthTokenCache.php
+++ b/src/FetchAuthTokenCache.php
@@ -1,0 +1,108 @@
+<?php
+/*
+ * Copyright 2010 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+use Psr\Cache\CacheItemPoolInterface;
+
+/**
+ * A class to implement caching for any object implementing
+ * FetchAuthTokenInterface
+ */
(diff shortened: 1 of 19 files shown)


bshaffer:
@dwsupplee and @jdpedrie I would love it if you could take a look as well!

src/FetchAuthTokenCache.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return array the response
+     *
+     * @throws \Exception
+     */
+    public function fetchAuthToken(callable $httpHandler = null)
+    {
+        // Use the cached value if its available.
+        //
+        // TODO: correct caching; update the call to setCachedValue to set the expiry
+        // to the value returned with the auth token.
```
tmatsuo: Why this is marked TODO? Is it somewhat difficult to do that?

tmatsuo:
LGTM, I have one question but it can be left as TODO and you can file another PR.

src/FetchAuthTokenCache.php
bshaffer: Hmm, no I don't think so. We might be able to do it w/o breaking BC (i.e. add an optional expiry parameter)

bshaffer:
As discussed, we should add an in-memory implementation for PSR-6 ([for example](https://github.com/php-fig/cache-util)) to this library to use as a default caching mechanism.

This can be consumed by googleapis/gax-php and googlecloudplatform/gcloud-php

dwsupplee:
I've opened https://github.com/google/google-auth-library-php/pull/126 against this PR which includes an in-memory implementation. Let me know what you think :).

dwsupplee:
I consent

bshaffer:
yeah it doesn't seem to work quite right in this scenario 😄

PR #126, opened 2016-07-25
add simple in-memory cache implementation

dwsupplee:


src/Cache/Item.php
```diff
@@ -0,0 +1,123 @@
...
+
+        return $this;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAt($expiration)
+    {
+        $this->expiration = $expiration;
```
jdpedrie: Should this value be checked for an instance of `DateTimeInterface` in the method? Will the comparison on line 84 fail or be incorrect?

dwsupplee: Since we still need to support 5.4 and `DateTimeInterface` is a feature added in 5.5 I thought it would be nice to leave the flexibility here. This allows users on 5.5 to pass in a DateTimeImmutable or custom DateTime implementation with no issues. The trade off is exactly what you brought up, however.

I'm open to strictly requiring `DateTime` if you don't think the flexibility is worth the trade off. Let me know what you think.

dwsupplee: WDYT about this? https://github.com/google/google-auth-library-php/pull/126/files#diff-5d21a48382f53b7e7fd11224f21e332aR101

jdpedrie: :+1:

src/Cache/InvalidArgumentException.php
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth\Cache;
+
+use Psr\Cache\InvalidArgumentException as PsrInvalidArgumentException;
+
+class InvalidArgumentException extends \InvalidArgumentException implements PsrInvalidArgumentException
```
bshaffer: -_-

bshaffer:
@dwsupplee this looks great!

src/Cache/InvalidArgumentException.php
tmatsuo: I wish there was a default concrete implementation provided by `Psr\Cache`, but presumably it's not the case.

src/Cache/Item.php
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAt($expiration)
+    {
+        if ($expiration instanceof \DateTimeInterface) {
+            $this->expiration = $expiration;
+        } elseif ($expiration instanceof \DateTime) {
```
tmatsuo: Nit: The `elseif` clause does the same thing as the first `if`. Does it make sense to consolidate them?

src/Cache/Item.php
```diff
@@ -0,0 +1,129 @@
...
+            $this->expiration = $expiration;
+        } else {
+            $this->expiration = null;
```
tmatsuo: When you pass invalid value, like `string`, it set the indefinite expiration. I would return the `InvalidArgumentException` in that case for safety. WDYT?

src/Cache/Item.php
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * {@inheritdoc}
+     */
+    public function expiresAfter($time)
+    {
+        if (is_numeric($time)) {
+            $this->expiration = new \DateTime("now + $time seconds");
+        } elseif ($time instanceof \DateInterval) {
+            $this->expiration = (new \DateTime())->add($time);
+        } else {
```
tmatsuo: Why not just making this method a wrapper for `expireAt`?

src/Cache/InvalidArgumentException.php
dwsupplee: That is correct, unfortunately all that is offered is an interface that must be implemented.

src/Cache/Item.php
dwsupplee: There's actually a reason for this :). The `DateTimeInterface` is only available in 5.5+, the `DateTime` check is for users on 5.4 who don't have access to that interface.

src/Cache/MemoryCacheItemPool.php
```diff
@@ -0,0 +1,154 @@
...
+     */
+    public function getItem($key)
+    {
+        return current($this->getItems([$key]));
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getItems(array $keys = [])
```
tmatsuo: I wish we could use generator. Is it because we need to support older versions of PHP?

src/Cache/MemoryCacheItemPool.php
```diff
@@ -0,0 +1,154 @@
...
+     */
+    public function deleteItem($key)
+    {
+        return $this->deleteItems([$key]);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function deleteItems(array $keys)
```
tmatsuo: This method can lead to partial deletion. Is it OK?

src/Cache/Item.php
dwsupplee: I tried to stick as closely to the interfaces defined by [psr6](https://github.com/php-fig/cache) as possible, as long as we don't mind straying from that definition I don't mind adding that.

tmatsuo: The scenario I'm worrying is this.

Someone call `expiresAt` with a string like `2016-10-31`, and it will persist forever, which may be surprising to the user.

tmatsuo:
I have put some comments on this, overall it looks nice :)

src/Cache/Item.php
dwsupplee: Hmmm, I'm having a hard time seeing the value in that. Could you explain further how you would see that working?

src/Cache/MemoryCacheItemPool.php
dwsupplee: Unfortunately yes, generators are a 5.5 feature. I am curious though, why would you prefer a generator instead of an array here?

src/Cache/Item.php
dwsupplee: That makes a lot of sense. @jdpedrie @bshaffer what do you guys think?

src/Cache/MemoryCacheItemPool.php
dwsupplee: Yes, I believe so. `clear` can be used to do a full deletion.

src/Cache/Item.php
tmatsuo: If it is just a wrapper, then we can:
Probably it's just a personal preference.

```php
public function expiresAfter($time)) 
{
    if (is_numeric($time)) {
        $expire = new \DateTime("now + $time seconds");
    } elseif ($time instanceof \DateInterval) {
        $expire = (new \DateTime())->add($time);
    } elseif  ($time === null) {
        // Optional for default value
        $expire = (new \DateTime())->add(self::DEFAULT_CACHE_LIFETIME);
    } else {
        // I think we should throw exception here although it's not defined in the interface
    }
    return $this->expiresAt($expire);
}
```

src/Cache/MemoryCacheItemPool.php
tmatsuo: For example, if the pool has the key `foo`, `bar`, and `baz` and you call this method with `bar`, `bogus`, and `foo`, then the method will throw an exception, but the result is somewhat inconsistent. The `bar` is deleted, but `foo` is not.

src/Cache/MemoryCacheItemPool.php
tmatsuo: Because a big array can exhaust your memory

src/Cache/MemoryCacheItemPool.php
dwsupplee: Ah, I follow now. Excellent point. Would it make more sense to have two loops? One which checks the keys, and another which actual performs the deletion?

tmatsuo: Yeah I prefer having 2 loops in this case. The tricky part is that, keys that passed the first loop still might lead to an exception in the second loop. Just omitting the check in the second loop is likely enough.

tmatsuo: Sorry I misread the code, we can simply have 2 loops. The second loop doesn't need to validate still.

dwsupplee: @tmatsuo How does this look? https://github.com/google/google-auth-library-php/pull/126/files#diff-c744bce1955390e9dc46f3ed4bcf5265R95

tmatsuo: Looks good, if you can add a test for the scenario I mentioned, that would be perfect :)

src/Cache/Item.php
tmatsuo: I can live with the current code though as long as the interface doesn't change. I think we should ask to change the interface.

tmatsuo: http://www.php-fig.org/psr/psr-6/#error-handling says:

> While caching is often an important part of application performance, it should never be a critical part of application functionality. Thus, an error in a cache system SHOULD NOT result in application failure. For that reason Implementing Libraries MUST NOT throw exceptions other than those defined by the interface, and SHOULD trap any errors or exceptions triggered by an underlying data store and not allow them to bubble.

So we should stick to the interface definition now.

src/Cache/MemoryCacheItemPool.php
dwsupplee: You've got it :) https://github.com/google/google-auth-library-php/pull/126/files#diff-5adad912143186cc944bd009d1fd3f3dR115

tests/Cache/MemoryCacheItemPoolTest.php
```diff
@@ -0,0 +1,189 @@
...
+    {
+        $keys = ['item1', '{item2}'];
+
+        foreach ($keys as $key) {
+            $this->saveItem($key, 'value');
+        }
+
+        $this->pool->deleteItems($keys);
+
+        $this->assertTrue($this->pool->hasItem($keys[0]));
```
bshaffer: Looks like this test will never reach this assert. We should remove the `@expectedException` and put `deleteItems` in a `try/catch` block.

tmatsuo: Good catch, also doesn't saveItem throw an exception too?

src/Cache/MemoryCacheItemPool.php
jdpedrie: Is that a big concern in this case? We should hope that this cache will only hold a few tokens at a time, and only for a short period of time (i.e. until the script ends).

It would be a problem if someone decided to use this as their cache implementation, but I doubt that would happen, and it is clearly not a great idea.

tests/Cache/MemoryCacheItemPoolTest.php
dwsupplee: You guys are awesome. Thanks for catching this, sorry for the oversight.

src/Cache/MemoryCacheItemPool.php
tmatsuo: Yeah as long as we need to support PHP 5.4 I can live with the current code.

This raises another discussion. Do we still need to support PHP 5.4, or PHP 5.5?

src/Cache/Item.php
jdpedrie: @tmatsuo I don't agree with this conclusion. The only reason passing `2016-03-10` is NOT an execution-halting error is because PHP versions less than 5.5 (which we still need to support) do not include `DateTimeInterface`. We would not say that it is incorrect to use that typehint if it was available in order to prevent the cache system from erroring.

Internal errors, where a storage system is unavailable or some other problem prevents a well-formed request from caching should be silenced as the PSR states, but errors in the implementing code, errors that would normally halt execution (failing parameter type constraint) SHOULD raise an error so that they can be fixed!

Hopefully the distinction I'm trying to make is clear. I can provide examples if not. :)

tests/Cache/MemoryCacheItemPoolTest.php
dwsupplee: How does this look?

 https://github.com/google/google-auth-library-php/pull/126/files#diff-5adad912143186cc944bd009d1fd3f3dR113

src/Cache/Item.php
jdpedrie: We should document that so later maintainers understand the reasoning and don't undo it.

dwsupplee: Sure. Updating with a comment.

src/Cache/Item.php
jdpedrie: To take this one step further, other libraries do throw exceptions from `expiresAt()` in the case of an invalid argument:

https://github.com/tedious/Stash/blob/master/src/Stash/Item.php#L365
https://github.com/symfony/cache/blob/master/CacheItem.php#L75

tmatsuo: @jdpedrie

I think we're on the same boat. Ultimately, I think we should raise an `InvalidArgumentException`, but as long as the PSR says `MUST NOT`, we should obey that.

I hope I don't misunderstand you.

tmatsuo:
LGTM

src/Cache/Item.php
dwsupplee: @jdpedrie I think that is a fair statement - and ultimately I agree it should be throwing an exception here. However I with agree with @tmatsuo that we should be sticking to the PSR.

As a counter example take a look at an implementation by @crell who helped form the PSR.

https://github.com/php-fig/cache-util/blob/master/src/BasicCacheItemTrait.php#L69

jdpedrie: I gave the section of PSR-6 you quoted a second close reading, and I think that your understanding of it is correct.

That said, I think this aspect of the document is poorly conceived. With the knowledge that other implementors deviate in this case, and the belief that this could be a major source of confusion and difficulty in debugging, I think that we should throw an exception anyways!

A better wording of the PSR probably should have created a distinction between errors bubbling up from the underlying system and errors in the way users call the methods. 👎

jdpedrie: @dwsupplee Perhaps some sort of E_USER_WARNING or E_USER_NOTICE then rather than an exception?

src/Cache/MemoryCacheItemPool.php
jdpedrie: PHP 5.5 is no longer maintained. I'd be a fan of seeing us move towards 5.6, but I do not know what other concerns need to be addressed prior to that happening. I do think it's very reasonable to ask users still on PHP 5.4 to start thinking about upgrading.

dwsupplee: That's a great question. I am all for moving us forward to more secure and updated versions.

We require 5.5+ in gcloud-php currently, but [google-api-php-client](https://github.com/google/google-api-php-client) is still on 5.4.

Do we have matrixes which provide a guideline for how far back we should provide support? I feel like this may warrant an issue or discussion outside of this PR.

src/Cache/Item.php
tmatsuo: @jdpedrie

I think the most correct way is to change the psr-6 and obey the new interface, which throws `InvalidArgumentException`. WDYT?

dwsupplee: Or even better see if this can be bumped up to 5.5 https://github.com/php-fig/cache/blob/master/composer.json#L13 and allow us to typehint the interface with DateTimeInterface :).

It appears all of the PSR interfaces are set at 5.3. It seems like it is worth opening a discussion to move that bar forward.

src/Cache/InvalidArgumentException.php
Crell: There is the [cache-utils](https://packagist.org/packages/fig/cache-util) package, which is intended as a tool to house the "common boilerplate" between different implementations.  PRs to add common stuff like that are welcome.

src/Cache/Item.php
Crell: We debated using DateTimeInterface in the type hint directly.  In the end we decided against it as many many projects were still on older PHP versions, and that would have been the ONLY post-5.3 feature in the spec.

PSRs don't have any fixed PHP version they must all rely on, but we do generally try to support as broad a segment of the market as possible for obvious reasons.

If a user passes a non-DateTime/DateTimeImmutable object to the method, it's not expected to work.  Using an assertion as in the utility trait is the closest approximation of type checking, and what I would recommend.

Crell:
Jumping in here because I was tagged...  Yes, I am/was the editor for PSR-6.

There appears to be a lot of code here that's duplicated with the cache-utils library.  Is there a reason to not leverage that and reduce code duplication?  (PRs against the utility library are also welcome.)

src/Cache/Item.php
tmatsuo: @Crell Thanks for chiming in. Does it make sense to change `expireAt` and `expireAfter` throwing `InvalidArgumentException`? or Do you prefer asserting like you said? I think exception is more generic.

dwsupplee:
Thanks for joining in on the conversation :). We did have a quick chat about that before starting any work and decided we wanted something we would have explicit control over.

bshaffer:
@Crell We love what you have in `cache-utils`. However, our requirements are so lightweight we'd rather not add the complexity of another dependency. Also, `cache-utils` does not have any stable tags.

src/Cache/InvalidArgumentException.php
bshaffer: I advocate adding an Exception like this one (`Fig\Cache\InvalidArgumentException`) to `cache-utils`.  I added a bug for this in https://github.com/php-fig/cache-util/issues/4

src/Cache/Item.php
Crell: Assertions and Errors (PHP 7) are for cases where a bug indicates the developer screwed up, and there's an error in the code itself.  Exceptions are for when the context is buggy, eg, user-input, DB goes away, etc.

If someone passes a non-DateTimeInterface-equivalent object, that's entirely on the developer to have done wrong so an assertion is appropriate.  It's the closest equivalent to what would happen had we typed the method against DateTimeInterface (with the resulting hard-PHP-5.5 dependency).

Crell:
Stable tags are easy enough to add as soon as it reaches a stable-ish point. :smile: The main thing missing, frankly, is other projects (like this one) testing it out to verify that it has enough in it that is non-buggy.  If it gets the job done I'm happy to start adding tags if it would be helpful.

A completely in-memory cache is an odd use of PSR-6, I grant, but for that matter it's also not at all specific to this library so could be its own stand-alone library, no?

src/Cache/MemoryCacheItemPool.php
bshaffer: Due to the large number of consumers of this library, we need to continue to support 5.4 for the foreseeable future as we are trying to drive Google adoption and not PHP 5.5 adoption.

src/Cache/Item.php
tmatsuo: Ok, I can live with assertions, but the behavior depends on the PHP configuration (`assert.bail`, `assert.warning`, `error_reporting`). I still prefer exceptions in this case.

> Exceptions are for when the context is buggy, eg, user-input, DB goes away, etc.
So why do you have `InvalidArgumentException`?

bshaffer:
@Crell  I agree a stand-alone library would be best

tests/Cache/MemoryCacheItemPoolTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $this->assertTrue($this->pool->save($item));
+
+        return $item;
+    }
+
+    public function testGetsFreshItem()
+    {
+        $item = $this->pool->getItem('item');
+        
+        $this->assertInstanceOf(Item::class, $item);
```
bshaffer: We need to change this, or skip in PHP 5.4. [Travis is failing](https://travis-ci.org/google/google-auth-library-php/jobs/147305702#L244)

tests/Cache/MemoryCacheItemPoolTest.php
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    public function testGetsExistingItem()
+    {
+        $key = 'item';
+        $value = 'value';
+        $this->saveItem($key, $value);
+        $item = $this->pool->getItem($key);
+
+        $this->assertInstanceOf(Item::class, $item);
```
bshaffer: Also here

tests/Cache/MemoryCacheItemPoolTest.php
```diff
@@ -0,0 +1,190 @@
...
+        $this->assertTrue($item->isHit());
+    }
+
+    public function testGetsMultipleItems()
+    {
+        $keys = ['item1', 'item2'];
+        $items = $this->pool->getItems($keys);
+
+        $this->assertEquals($keys, array_keys($items));
+        $this->assertContainsOnlyInstancesOf(Item::class, $items);
```
bshaffer: And here

tests/Cache/MemoryCacheItemPoolTest.php
jdpedrie: `::class` was added in PHP 5.5. :sad:

dwsupplee: Fixed

jdpedrie:
@Crell to add a bit of history to the reasoning behind using PSR-6 here, we need to cache tokens just enough so that we're not renegotiating tokens multiple times within a single request/response application run. As you alluded to, using PSR-6 here is a bit of overkill, but we also wanted to provide a common interface for users to override the caching system with one they already have in their application. So if you're using memcached via Stash, you can use that in place of this simple cache implementation.

Essentially, we're adding some extra weight where normally one would expect a static or something in order to allow users additional flexibility and control where their application warrants it.

Crell:
That makes sense, yes.  In which case, collaborating on the shared util library also makes sense, IMO, as it would likely be shared by any other implementation someone dropped in.

src/Cache/Item.php
tmatsuo: If it is tough to change the interface, can you also add a recommendation for having assertion (instead of throwing exception) in the psr? Seems like there are some libraries which throws exceptions.

dwsupplee: The [PHP docs](http://us2.php.net/manual/en/function.assert.php) for assert state:

> **Assertions should be used as a debugging feature only.** You may use them for sanity-checks that test for conditions that should always be TRUE and that indicate some programming errors if not or to check for the presence of certain features like extension functions or certain system limits and features.

>**Assertions should not be used for normal runtime operations like input parameter checks.** As a rule of thumb your code should always be able to work correctly if assertion checking is not activated.

If we do use assertions there isn't a guarantee they are being handled - for that reason I am not in favor of them.

Crell: Well we can't change the spec itself now, as PSRs are fixed.  But that does sound like valid stuff to include in an Errata.

tmatsuo: @Crell Thanks! So my temporary conclusion is to add an assertion for now, create a bug to change it to exception once the errata comes out.

WDYT?

Crell: Type checks are a form of assertion, so they're a reasonable alternative.  A bad type check should only happen on developer error.  If you're passing a user input string directly to a method asking for a typed object, that's unquestionably a developer error.  It will still break with assertions turned off, just with a much less helpful error message.

If anything, the errata we add would recommend an assertion, not an exception, as the exception would be a spec change.  But that's something to hash out on the FIG list.

bshaffer: We should issue a PHP warning, as PSR-6 is against throwing an exception, and the PHP docs are against using `assert`.

```php
trigger_error($error_msg, E_USER_WARNING)
```

bshaffer:
@Crell because this library is widely used, we've already experienced a few nightmare dependency scenarios (re Guzzle). We are trying to include as few dependencies as possible. DRY is nice, but not as nice as your application actually being able to install. 

As an aside, it would be great for composer to have support for "any libraries with tag X, and default to some-repo-with-tag-X". I could then say "This library supports anything with PSR-6, but go ahead and add `psr6-utils` unless another one is specified". Maybe if we containerize our dependencies and add container support in composer this would be possible. Or it would be insane, I'm not quite sure.

Crell:
I think there's some advanced composer-fu for that, but I'm not certain.  (@weierophinney can speak to that better.)  However, cache-utils is not a complete implementation nor does it intend to be.  It's a tool to make building implementations require less boilerplate.

bshaffer:
I think we should make these classes `final`. Thoughts @dwsupplee @jdpedrie and @tmatsuo ?

src/Cache/Item.php
Crell: FIG discussion of the recommended handling here:

https://groups.google.com/forum/#!msg/php-fig/RsVcQ4w5IHw/WSb6m_zlAgAJ

(Per FIG conventions, please discuss on the list, not on the PR, except for small code/grammar tweaks.  Thanks.)

jdpedrie:
@bshaffer what is the reasoning for making them final?

bshaffer:
To make it clear these classes should not be used out of the context of this library. This will also prevent subclassing them.

dwsupplee:
While I am not against using final here, I am curious how it helps to make it clear one shouldn't use these classes outside of this context. I don't think I would ever assume that having seen the final keyword. You could still use the implementations wherever you wanted regardless of whether they are final or not, correct?

bshaffer:
You're right, it just prevents BC issues caused by subclassing. I have seen the `final` keyword on classes that are supposed to be private, but you're right it doesn't really mean that.

nicolas-grekas:
Why not using symfony/cache instead or creating a new implementation that you'll need to maintain?

bshaffer:
Hi @nicolas-grekas! welcome to the thread 😄 

We only want to implement a default lightweight in-memory caching implementation, and did not want to overcomplicate things with a large dependency / secondary caching implementation. Our users are welcome to use `symfony/cache` instead, as this library is `PSR-6`-compliant.

tmatsuo:
LGTM
We can implement the type checking on `expireAt` once PHP-fig have a conclusion.

tmatsuo:
LGTM!

PR #131, opened 2016-08-01
Do not overwrite refresh token with null

michaelbausor:
Prevent overwriting the refreshToken with null when no refresh token is returned in fetchAuthToken method
Fixes https://github.com/google/google-auth-library-php/issues/130

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -491,11 +491,6 @@ class OAuth2 implements FetchAuthTokenInterface
 
         $response = $httpHandler($this->generateCredentialsRequest());
         $credentials = $this->parseTokenResponse($response);
-        // We expect the refresh token returned to be unset when the grant
-        // type of the request was refresh_token, and the current refresh
-        // token, used in the request, has not expired. In that case, we
-        // should not overwrite the current refesh token with null
-        $credentials = $this->addRefreshToken($credentials);
         $this->updateToken($credentials);
 
         return $credentials;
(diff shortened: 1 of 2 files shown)


bshaffer:
A cleaner solution would be to change updateToken to not merge in a null refresh_token, i.e.:

```php
    public function updateToken(array $config)
    {
        // ** I removed "refresh_token" from here
        $opts = array_merge([
            'extensionParams' => [],
            'access_token' => null,
            'id_token' => null,
            'expires' => null,
            'expires_in' => null,
            'expires_at' => null,
            'issued_at' => null,
        ], $config);

        $this->setExpiresAt($opts['expires']);
        $this->setExpiresAt($opts['expires_at']);
        $this->setExpiresIn($opts['expires_in']);
        // By default, the token is issued at `Time.now` when `expiresIn` is set,
        // but this can be used to supply a more precise time.
        if (!is_null($opts['issued_at'])) {
            $this->setIssuedAt($opts['issued_at']);
        }

        $this->setAccessToken($opts['access_token']);
        $this->setIdToken($opts['id_token']);
        // ** I added this conditional here!
        if (isset($opts['refresh_token'])) {
            $this->setRefreshToken($opts['refresh_token']);
        }
    }
```

michaelbausor:
I had something very similar at first as well. I decided against doing it this way because it means that updateToken cannot set the refresh token to null under any circumstances, so someone calling updateToken manually would also need to call setRefreshToken separately.

If you think this isn't an issue, then I definitely agree that your proposed solution is cleaner.

Separately, it seems as though the 'expires' setting is doing no work at all - it is always overwritten by 'expires_at'. Perhaps I can remove this at the same time?

bshaffer:
Just use `array_has_key` instead of `isset` and a caller will be able to set the refresh token to null.

Yes, we should definitely remove `expires`. This won't break BC since it's already being overwritten.

michaelbausor:
PTAL

src/OAuth2.php
```diff
@@ -600,7 +597,12 @@ public function updateToken(array $config)
 
         $this->setAccessToken($opts['access_token']);
         $this->setIdToken($opts['id_token']);
-        $this->setRefreshToken($opts['refresh_token']);
+        // We expect refresh_token to be unset when an access token is obtained
```
bshaffer: sorry, this comment is very hard to read. How about

> The refresh token should only be updated if a value is explicitly passed in,
> as some access token responses do not include a refresh token.

michaelbausor: Updated, PTAL

bshaffer:
Thank you for working on this, it looks great!

michaelbausor:
No worries :-)

PR #133, opened 2016-08-26
Change authorization header key to lowercase

stanley-cheung:
For issue grpc/grpc#7881

A recent GFE reloadable flag rollout caused a few Cloud APIs to start rejecting some requests from gRPC PHP clients because in some cases, the 'Authorization' header key for the bearer token got directly sent to the server, and it violates HTTP/2 standards. I am hoping we can change the header key to lower case.

### DIFF
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -30,7 +30,7 @@ abstract class CredentialsLoader implements FetchAuthTokenInterface
     const ENV_VAR = 'GOOGLE_APPLICATION_CREDENTIALS';
     const WELL_KNOWN_PATH = 'gcloud/application_default_credentials.json';
     const NON_WINDOWS_WELL_KNOWN_PATH_BASE = '.config';
-    const AUTH_METADATA_KEY = 'Authorization';
+    const AUTH_METADATA_KEY = 'authorization';
 
     /**
      * @param string $cause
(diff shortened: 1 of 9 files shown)


bshaffer:
This is confusing, as every example I've seen uses uppercase-first `Authorization`:

- [Oauth Spec](https://tools.ietf.org/html/rfc6750#section-2.1)
- [Authorization Header Spec](https://tools.ietf.org/html/rfc2617#section-2)

I do not see how this could be an issue in the client.

stanley-cheung:
Please see http://httpwg.org/specs/rfc7540.html#HttpHeaders

"Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared in a case-insensitive fashion. However, header field names MUST be converted to lowercase prior to their encoding in HTTP/2. A request or response containing uppercase header field names MUST be treated as malformed (Section 8.1.2.6)."

bshaffer:
Shouldn't that transformation be happening in the underlying library? I can see this being an ongoing issue if the clients are expected to lowercase all their headers specifically for HTTP/2

stanley-cheung:
Yes we are also doing the transformation and error checking on the gRPC side. But I thought that since for HTTP1.1 it's case insensitive anyways, we might as well always lowercase the key in the auth library?

stanley-cheung:
Ping?

PR #138, opened 2016-10-20
addresses issues in some caching implementations where data is returned for expired items

bshaffer:
see https://github.com/tedious/Stash/issues/327


tmatsuo:
LGTM. Do you want to add a test?

bshaffer:
@tmatsuo I'm leaning towards implementing https://github.com/google/google-api-php-client/pull/1085 instead, as the above call to `isHit` should not be required, and is only needed for https://github.com/tedious/Stash/issues/327

PR #139, opened 2016-10-21
fixes cache-key stripping to be in line with psr-6 spec

bshaffer:


src/CacheTrait.php
```diff
@@ -67,6 +67,6 @@ private function getFullCacheKey($key)
         $key = $this->cacheConfig['prefix'] . $key;
 
         // ensure we do not have illegal characters
-        return str_replace(['{', '}', '(', ')', '/', '\\', '@', ':'], '-', $key);
+        return preg_replace('|[^a-zA-Z0-9_\.! ]|', '', $key);
```
tmatsuo: why there is a space after the ecxclamation mark?

bshaffer: Looking through [the spec](http://www.php-fig.org/psr/psr-6/#definitions), there space is not in the list of allowed keys. However, the spec does not forbid the space, it only forbids characters in this list: `{}()/\@:`.

Because this library is just a consumer of other caching libraries, I think it's best to remove this logic all together, and add it to `google-api-php-client`.

PR #142, opened 2016-12-21
fix the cURL:60 SSL cert issue

peter279k:
# Change log
- add the pem file which id from cURL official website.
- fix the cURL:60 SSL certificate issue.


bshaffer:
Hi @peter279k! Thank you for your contribution.

We should not have to supply our own pem file by default, as guzzle supports locating the proper pem file. I believe a better fix would be to allow you to pass in your own options to [`OAuth2::fetchAuthToken`] like this:

```php
    /**
     * Fetches the auth tokens based on the current state.
     *
     * @param callable $httpHandler callback which delivers psr7 request
     *
     * @return array the response
     */
    public function fetchAuthToken(callable $httpHandler = null, array $options = [])
    {
        if (is_null($httpHandler)) {
            $httpHandler = HttpHandlerFactory::build();
        }
        $response = $httpHandler($this->generateCredentialsRequest(), $options);
        $credentials = $this->parseTokenResponse($response);
        $this->updateToken($credentials);
        return $credentials;
    }
```

What do you think?

peter279k:
Hi @bshaffer , I think we can consider the following two points:

- The user should add or customize own pem file for SSL certificate.
- How about providing the default pem file? So that it can make user easily use this library.

bshaffer:
Thanks for this! Can you explain here or link to material explaining why this library should have its own pem file and why this is easier for the user?

Thanks!

peter279k:
Hi @bshaffer , here is my answer:
- There are some SDK packages will support their certs by default. For example, the [Facebook-SDK](https://github.com/facebook/php-graph-sdk/blob/master/src/Facebook/HttpClients/certs/DigiCertHighAssuranceEVRootCA.pem) do this.
- The users use this package easily because they just download the package and run their code.They even don't care about the pem files and SSL certificate issues.

bshaffer:
Closing in favor of #165

PR #143, opened 2017-01-24
Fix base_url option

themightychris:
The option is called `base_uri`: http://docs.guzzlephp.org/en/latest/quickstart.html


themightychris:
I signed it!

PR #153, opened 2017-06-08
Create composer.json

JurRutten:
update to guzzlehttp/psr7 1.3


JurRutten:
I signed it!

bshaffer:
`~1.2` will install 1.3 (and at the time of this comment, `1.4.2`), so this change is not necessary unless there was a change in 1.3 that this library relies on.

PR #156, opened 2017-06-27
Modified how the home directory is determined using proper cross plat…

logistiker:
…form methods for PHP since Apache sets the HOME environment variable incorrectly or not at all.


logistiker:
I signed it!

bshaffer:
It would be better to explicitly set your credentials by setting `GOOGLE_APPLICATION_CREDENTIALS` as the path to your credentials file. Looking in the `$HOME` directory for the `gcloud` file is for local development / CLI use only.

Additionally, this solution does not work on windows, as `posix_getpwuid` doesn't exist.

Issue #148, opened 2017-05-01
Implement a Simple Auth Client


bshaffer:
It would be great to implement a simple auth client similar to the `Google_Client` class that allows the user to [make http calls directly](https://github.com/google/google-api-php-client#making-http-requests-directly) with minimum setup, i.e.:

```php
use Google\Auth\Client;

$client = new Client();
$client->useApplicationDefaultCredentials();
$http = $client->authorize();

$http->get('drive/v2/files');
```

This could include a lot of the auth-related code in `Google_Client`, i.e. the OAuth flows to exchange auth codes and refresh tokens for access tokens.

bshaffer:
For simplicity, another method would be to implement a method (similar to `HttpHandlerFactory::build`) to return a client from a set of credentials, ie

```php
class CredentialsLoader
{
    /// ...
    public static function buildHttpClient(CredentialsLoader $credentials, ClientInterface $client = null)
    {
        // ...
        return $client; // authorized Guzzle client
    }
}
```

PR #162, opened 2017-08-10
adds makeHttpClient from credentials

bshaffer:
see #148 

cc @michaelbausor, @ryanmats

Is this useful?

Usage:

```php
$credentials = GCECredentials();
$client = ClientLoader::makeHttpClient($credentials, [
    'base_url' => 'https://www.googleapis.com/taskqueue/v1beta2/projects/'
]);
$response = $client->get('myproject/taskqueues/myqueue');
```


ryanmats:
yes, this would be helpful for the HTTP auth samples.

PR #164, opened 2017-08-17
Modified how the home directory in posix compliant environments is de…

logistiker:
…termined using posix methods for PHP since Apache sets the HOME environment variable incorrectly or not at all in posix environments.


logistiker:
This leaves in tact the current code for windows why replacing non windows code with posix compliant calls because they will always find the user home directory whereas getenv('HOME') will not.

bshaffer:
@logistiker this is failing in every travis build, so it does not appear to be a safe or equivalent change.

PR #165, opened 2017-08-25
fix the cURL:60 SSL cert issue

bshaffer:
See #142


tmatsuo:
I'm not sure if bundling the cacert.pem is a good idea. This file needs to be updated whenever the list of authorities changes, right?

bshaffer:
yeah, that's a good point. I am curious if anyone else really needs this. I've seen it done in other repositories, but as guzzle itself doesn't include a pem, I am thinking it is not necessary.

tmatsuo:
I think it's better to defer it (the update of the file) to the operating systems (apt-get, etc).

PR #166, opened 2017-08-25
Allow phpunit to set the home directory

logistiker:
Made an additional modification to allow setting the home directory when the credential loader is being run in the phpunit environment since posix functions cannot be mocked in a unit testing environment.  Note that you can't test everything in a unit testing environment and expect this code will work in all environments.  The bug this is fixing happens when this code is run on apache linux and could only be caught with an integration test.

### DIFF
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -89,8 +89,13 @@ abstract class CredentialsLoader implements FetchAuthTokenInterface
      */
     public static function fromWellKnownFile()
     {
-        $rootEnv = self::isOnWindows() ? 'APPDATA' : 'HOME';
-        $path = [getenv($rootEnv)];
+        if (self::isOnWindows()) {
+            $rootEnv = 'APPDATA';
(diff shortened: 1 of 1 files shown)


bshaffer:
A better way to do this would be to temporarily set the home directory to something else. See [this test](https://github.com/google/google-auth-library-php/blob/0420299c61f2b6eef1da465342a1b66b34b36c13/tests/ApplicationDefaultCredentialsTest.php#L29) for an example.

logistiker:
Your comment to temporarily set the home directory to something else doesn't make any sense.  It's not possible to override what the posix functions return.  You have unit tests which directly test the fromWellKnownFile method: 

```
    public function testIsNullIfFileDoesNotExist()
    {
        putenv('HOME=' . __DIR__ . '/../not_exists_fixtures');
        $this->assertNull(
            ServiceAccountCredentials::fromWellKnownFile()
        );
    }
```

Those will always fail because the posix functions will always get the true home directory for the user running phpunit.

Issue #149, opened 2017-05-16
Implement IAP/JWT client


nohn:
Please implement the functionality required to connect to IAP protected services like in the Python Auth Client: https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/iap/make_iap_request.py

bshaffer:
@nohn Can you explain what the missing auth is?

nohn:
The basic documentation is on https://cloud.google.com/iap/docs/authentication-howto. There a bit more on https://cloudplatform.googleblog.com/2017/04/Getting-started-with-Cloud-Identity-Aware-Proxy.html. It seem like at least the target_audience claim is missing, but I'm not sure, if that's all.

b4b4r07:
+1

bshaffer:
We are working on this! cc @ryanmats

b4b4r07:
When will it be implemented?

bshaffer:
@b4b4r07 Is the request to support the field `target_audience` in the JWT payload? I haven't tested it out, but it may be possible through something like this: https://gist.github.com/bshaffer/35fc2fb65df200c5e3a201c70c466c8b

b4b4r07:
Thank you. Does that script really work? I tried but I got an error. As far as I can see, making IAP request needs to refresh token. However that script doesn't do that.

bshaffer:
It won't work until #171 is merged, and then you will be able to do the following:

```php
$oauth->setAdditionalClaims([
    'target_audience' => $clientID,
]);
```

I've updated the sample to reflect this, but if you could pull down the branch and test, that would be awesome! You can do so by changing `composer.json` to look something like this (and then running `composer update`):

```json
{
    "required": {
        "google/auth": "dev-additional-claims as v1.1"
    }
}
```

bshaffer:
If someone on this thread could confirm this change to the auth library enables a successful IAP call, I will get the change merged and a new version tagged.

nohn:
```
$ php test.php
PHP Fatal error:  Uncaught Error: Class 'Google\Auth\OAuth2' not found in /home/project/iaphp/test.php:18
Stack trace:
#0 {main}
  thrown in /home/project/iaphp/test.php on line 18
```

```
$ cat composer.json
{
    "required": {
        "google/auth": "dev-additional-claims as v1.0.1"
    },
    "require": {
        "google/auth": "dev-additional-claims as v1.0.1"
    }
}

```

[...]

PR #171, opened 2017-09-29
adds support for additional claims in JWT tokens

bshaffer:
see #149


bshaffer:
cc @ryanmats

bshaffer:
@ryanmats were you able to get IAP working with this change?

bshaffer:
cc @jonparrott

PR #172, opened 2017-10-03
Accept a default httpHandler at the construct level in FetchAuthToken

RyanGordon:
I've been trying to make this work:

```php
$httpHandler = HttpHandlerFactory::build(new GuzzleHttp\Client([
    'timeout' => 1.0,
    'connect_timeout' => 0.3,
    'read_timeout' => 0.7
]);
$client = new SpannerClient([
    'authHttpHandler' => $httpHandler
]);
```

From a high level this doesn't work because in [gax-php GrpcCredentialsHelper.php](https://github.com/googleapis/gax-php/blob/master/src/GrpcCredentialsHelper.php#L133) it does not pass in the `authHttpHandler` function and at a higher level, none of the options get passed into GrpcCredentialsHelper because all of those get set via [getGaxConfig](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Spanner/Connection/Grpc.php#L153) and include none of the higher level configuration.

This fix works by setting the httpHandler in the construction of [FetchAuthTokenCache](https://github.com/GoogleCloudPlatform/google-cloud-php/pull/692/files#diff-d24fa15bc71079d398ddee7a482091ccR153).

Related: https://github.com/GoogleCloudPlatform/google-cloud-php/pull/692

### DIFF
--- a/src/FetchAuthTokenCache.php
+++ b/src/FetchAuthTokenCache.php
@@ -42,10 +42,16 @@ class FetchAuthTokenCache implements FetchAuthTokenInterface
      */
     private $cache;
 
+    /**
+     * @var callable
+     */
+    private $httpHandler;
+
     public function __construct(
         FetchAuthTokenInterface $fetcher,
         array $cacheConfig = null,
-        CacheItemPoolInterface $cache
+        +        CacheItemPoolInterface $cache,
+        +        callable $httpHandler = null
     ) {
         $this->fetcher = $fetcher;
         $this->cache = $cache;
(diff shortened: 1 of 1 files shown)


src/FetchAuthTokenCache.php
```diff
@@ -42,17 +42,24 @@ class FetchAuthTokenCache implements FetchAuthTokenInterface
      */
     private $cache;
 
+    /**
+     * @var callable
```
dwsupplee: Really minor nit: `callable|null`

RyanGordon: Thanks, Fixed!

bshaffer:
Closing, see comments on https://github.com/GoogleCloudPlatform/google-cloud-php/pull/692

PR #174, opened 2017-11-04
Updated to PHPUnit 5.7

gabriel-caruso:
I've update PHPUnit to `5.7.*`, and dropped support to PHP versions `5.4` and `5.5`, as [PHPUnit 5 requires PHP `5.6`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-5.0.md#removed-1).

### Why dropped support for PHP versions `5.4` and `5.5`?
These versions are [no longer supported by PHP](http://php.net/supported-versions.php).

### Why PHPUnit 5.7 and not PHPUnit 6?
Because [PHPUnit 6 requires PHP 7](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#removed), and we will need to work on that.

But, I use [PSR-1](http://www.php-fig.org/psr/psr-1/#namespace-and-class-names) while extending PHPUnit TestCase class. This will help us when to migrate to PHPUnit 6, that [no longer support snake case namespaces](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ branches:
 sudo: false
 
 php:
-  - 5.4
-  - 5.5
   - 5.6
   - 7.0
   - 7.1
--- a/composer.json
+++ b/composer.json
@@ -6,7 +6,7 @@
   "homepage": "http://github.com/google/google-auth-library-php",
   "license": "Apache-2.0",
   "require": {
-    "php": ">=5.4",
+    "php": ">=5.6",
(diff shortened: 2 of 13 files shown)


bshaffer:
Thank you for your Pull Request! Unfortunately, even though PHP 5.4 and PHP 5.5 are no longer supported, we still have users on those versions. So unless there is a good reason to update it (a better reason than upgrading our PHPUnit version), we are going to continue supporting these versions of PHP.

PR #175, opened 2017-11-07
Use PHPUnit\Framework\TestCase instead of PHPUnit_Framework_TestCase

carusogabriel:
A simple version of #174.

Now, I'm just updating to [`PHPUnit 4.8.36`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-4.8.md#4836---2017-06-21), keep all PHP versions support, and use the `PHPUnit\Framework\TestCase` notation instead of `PHPUnit_Framework_TestCase` while extending our TestCases. This will help us migrate to PHPUnit 6, that [no longer support snake case class names](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#changed-1).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "psr/cache": "^1.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "3.7.*",
+    "phpunit/phpunit": "^4.8.36",
     "friendsofphp/php-cs-fixer": "^1.11"
   },
   "autoload": {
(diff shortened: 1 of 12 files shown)


jdpedrie:
@bshaffer fwiw, I've been testing out `"phpunit/phpunit": "^4.8|^5.0"` and it's worked well. If you're running PHP 7, you get the improved error handling (because of how PHP7 reorganized the exception hierarchy, older PHPUnit doesn't report errors very well), and you maintain backwards compatibility with the old `PHPUnit_Framework_TestCase`. So long as you don't use features introduced in 5.x, you should be good.

bshaffer:
This looks great! The only issue I'm encountering is [a strange one on PHP 7.2](https://travis-ci.org/google/google-auth-library-php/jobs/308005201) when `composer update --prefer-lowest` is used. Because of this, I've changed things up a little in #176 so that `phpunit` is installed after `--prefer-lowest`, so that the prefer lowest option does not apply.

I'll merge this PR once we get #176 in

gabriel-caruso:
@bshaffer Ok, thanks for the feedback.

PR #177, opened 2017-11-29
Fix undefined variable issue

seferov:


seferov:
I signed it!

src/Cache/Item.php
```diff
@@ -134,7 +134,7 @@ public function expiresAfter($time)
         } else {
             $message = 'Argument 1 passed to %s::expiresAfter() must be an ' .
                        'instance of DateInterval or of the type integer, %s given';
-            $error = sprintf($message, get_class($this), gettype($expiration));
+            $error = sprintf($message, get_class($this), gettype($this->expiration));
```
bshaffer: This should actually be `gettype($time)`, as the error message is referring to the argument type.

PR #178, opened 2017-12-06
Refactoring tests

carusogabriel:
I've refactored some tests, using:
- `assertFalse` instead of `assertTrue` with `!` before the condition;
- `assertInternalType`, avoiding using `is_*` functions;
- `assertGreaterThan` for mathematical comparisons;
- `assertArrayHasKey` and `assertArrayNotHasKey` instead of `array_key_exists` and `isset` functions;
- `assertEquals` and `assertNotEquals` instead of comparisons.

### DIFF
--- a/tests/Credentials/ServiceAccountCredentialsTest.php
+++ b/tests/Credentials/ServiceAccountCredentialsTest.php
@@ -294,14 +294,16 @@ class SACFetchAuthTokenTest extends TestCase
             $testJson
         );
         $update_metadata = $sa->getUpdateMetadataFunc();
-        $this->assertTrue(is_callable($update_metadata));
+        $this->assertInternalType('callable', $update_metadata);
 
         $actual_metadata = call_user_func($update_metadata,
             $metadata = array('foo' => 'bar'),
             $authUri = null,
             $httpHandler);
-        $this->assertTrue(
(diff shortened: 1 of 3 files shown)


tests/Credentials/ServiceAccountCredentialsTest.php
```diff
@@ -495,15 +496,16 @@ public function testNoScopeAndNoAuthUri()
...
             $metadata = array('foo' => 'bar'),
             $authUri = null);
         // no access_token is added to the metadata hash
         // but also, no error should be thrown
-        $this->assertTrue(is_array($actual_metadata));
-        $this->assertTrue(
-            !isset($actual_metadata[CredentialsLoader::AUTH_METADATA_KEY]));
+        $this->assertInternalType('array', $actual_metadata);
+        $this->assertFalse(
+            isset($actual_metadata[CredentialsLoader::AUTH_METADATA_KEY])
```
bshaffer: Would this be better as 

```php
 $this->assertArrayNotHasKey(CredentialsLoader::AUTH_METADATA_KEY, $actual_metadata);
```

carusogabriel: Good one! Gonna change it and search for similars

PR #181, opened 2017-12-23
Fix bug when merging headers on Guzzle5 http handler

erictinocopro:
This will avoid removing already set headers with headers from options (like encryption)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## v1.2.1 (23/12/2017)
+
+### Changes
+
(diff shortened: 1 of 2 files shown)


erictinocopro:
I signed it!

erictinocopro:
I signed it!

bshaffer:
@erictinocopro thank you for this, but as we have not tagged `1.2.1` yet, please remove your entry to the CHANGELOG and I will add it when we tag a new version.

Also, it would be great to have a test!

bshaffer:
See #186

PR #183, opened 2018-01-18
Pass correct parameter to AuthTokenMiddleware

kaimallea:
`AuthTokenMiddleware` expects a `$httpCallable`, but is being passed `$cacheConfig`.


bshaffer:
Good catch! Thank you for finding this! I've added this to #184.

Issue #188, opened 2018-03-13
GAE_VM is a deprecated environment variable, so GCECredentials::onAppEngineFlexible returns false.


patrick-yi-82:
I've been having trouble getting default credential from the app engine, and found out that GAE_VM is a deprecated environment variable. (according to https://cloud.google.com/appengine/docs/flexible/python/upgrading#environment_variable_changes)

As a workaround, I am using GAE_SERVICE to check if I am on a GAE, however I felt this issue should be mentioned so the auth library can be updated.

https://github.com/google/google-auth-library-php/blob/0420299c61f2b6eef1da465342a1b66b34b36c13/src/Credentials/GCECredentials.php#L111

bshaffer:
Thank you for this! Looks like you're right, we need to update this right away. I found the environment variable `GAE_INSTANCE` which on App Engine Flexible is automatically prefixed with `aef-`. This is a pretty reliable way to handle this check.

tmatsuo:
I would not rely on the fact that the instance names are prefixed with `aef-`. Unless this is a written contract, it may change any time.

bshaffer:
Unfortunately, this is the best solution we have at this time. There are no environment variables in Flex which are not available in other App Engine environments (or planned to be). And Flex is being turned down, so pushing an update to the runtime, while possible, is not encouraged.

bshaffer:
@tmatsuo what do you think we should use instead?

tmatsuo:
@bshaffer I would ask the PM or the tech lead for the best option :)

bshaffer:
Fixed in [`v1.3.0`](https://github.com/google/google-auth-library-php/releases/tag/v1.3.0)

PR #189, opened 2018-03-14
Fixes usage of deprecated env var for GAE Flex

bshaffer:
See #188


tmatsuo:
The only concern is that this change assumes the instance names are all prefixed with `aef-`. If this is in the Flex contract, it's good, but in case it's not in the contract, maybe we can just check the existence of the envvar?

bshaffer:
The concern is that `GAE_INSTANCE` is present on other versions of App Engine (e.g. Standard and future versions). 

How do you feel about this being a stop-gap until we can implement a more robust solution (in the runtime itself)?

tmatsuo:
Ok

PR #190, opened 2018-03-29
fix - guzzlehttp/psr7 dependency version definition

ait-sd:
Google api client relies on guzzlehttp/psr7:^1.2


ait-sd:
I signed it!

tmatsuo:
Those are basically the same, but having the same constraints is better. Thanks

PR #193, opened 2018-04-11
Removed useless else after return

alexmanno:
Removed useless else after return for more readability

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -547,7 +547,7 @@ class OAuth2 implements FetchAuthTokenInterface
             return $res;
         }
 
-// Assume it's JSON; if it's not throw an exception
+        // Assume it's JSON; if it's not throw an exception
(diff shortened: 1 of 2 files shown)


src/OAuth2.php
```diff
@@ -543,14 +545,14 @@ public function parseTokenResponse(ResponseInterface $resp)
...
             return $res;
-        } else {
-            // Assume it's JSON; if it's not throw an exception
-            if (null === $res = json_decode($body, true)) {
-                throw new \Exception('Invalid JSON response');
-            }
+        }
 
-            return $res;
+// Assume it's JSON; if it's not throw an exception
```
bshaffer: This needs to be indented

bshaffer:
cc @googlebot

alexmanno:
@bshaffer Fixed! I also have signed the CLA

PR #194, opened 2018-04-14
RFC: Add ClientCredentials class

michaelbausor:
For now this is just copying the AuthWrapper class from https://github.com/googleapis/gax-php/blob/master/src/ApiCore/AuthWrapper.php. We should discuss what features are appropriate for the Auth library, and which are best left in the client libraries. The way I see the features at the moment is:
### High value
- Hold onto an `authHttpHandler` object so it doesn't need to always be supplied to `fetchAuthToken()`
- Always check `getLastReceivedToken()` for an unexpired token before calling `fetchAuthToken()`
### Medium value
- Potential for simplified construction methods, in particular to help with using caching. E.g. we could add a factory method that uses the System V cache
- Functions to get bearer string and gRPC credentials callback

### Still to resolve
- What exceptions should we throw? Do we want to be throwing `Exception`?
- Is this just going to confuse things? Should we instead scrap the factory methods, and ask people to use `ApplicationDefaultCredentials::<X>` and then pass it into a `new ClientCredentials(...)` call?
- What should we name this class? I chose `ClientCredentials` instead of `AuthWrapper`, but I'm not sure that is much of an improvement.

### DIFF
--- a/src/ClientCredentials.php
+++ b/src/ClientCredentials.php
@@ -1,0 +1,247 @@
+<?php
+/*
+ * Copyright 2018, Google Inc.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *     * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
@bshaffer have you had a chance to check this one out?  Looks like it slipped through the cracks :)

src/ClientCredentials.php
```diff
@@ -0,0 +1,247 @@
...
+                return '';
+            }
+        }
+        return $token['access_token'];
+    }
+
+    private static function isValid($token)
+    {
+        return is_array($token)
+            && array_key_exists('access_token', $token);
```
wolzogen:
```php
return is_array($token) && isset($token['array_key_exists']);
```

src/ClientCredentials.php
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    private static function isExpired($token)
+    {
+        return !(self::isValid($token)
+            && array_key_exists('expires_at', $token)
```
wolzogen:
```php
&& isset($token['expires_at'])
```

src/ClientCredentials.php
```diff
@@ -0,0 +1,247 @@
...
+            'scopes'            => null,
+            'authHttpHandler'   => null,
+            'enableCaching'     => true,
+            'authCache'         => null,
+            'authCacheOptions'  => null,
+        ];
+        $authHttpHandler = $args['authHttpHandler'] ?: self::buildHttpHandlerFactory();
+        $authCacheOptions = $args['authCacheOptions'];
+        if ($args['enableCaching']) {
+            $authCache = $args['authCache'] ?: new MemoryCacheItemPool();
```
wolzogen:
```php
$authCache = $args['authCache'] ?: new MemoryCacheItemPool;
```

src/ClientCredentials.php
```diff
@@ -0,0 +1,247 @@
...
+            'enableCaching'     => true,
+            'authCache'         => null,
+            'authCacheOptions'  => [],
+            'authHttpHandler'   => null,
+        ];
+
+        $loader = CredentialsLoader::makeCredentials($authConfig['scopes'], $keyFile);
+
+        if ($authConfig['enableCaching']) {
+            $authCache = $authConfig['authCache'] ?: new MemoryCacheItemPool();
```
wolzogen:
```php
$authCache = $authConfig['authCache'] ?: new MemoryCacheItemPool;
```

michaelbausor:
Closing stale PR

PR #195, opened 2018-06-05
Add retry and increase timeout for GCE credentials

ZhouyihaiDing:
It seems 0.3s timeout is not enough and cause flaky sometimes: https://github.com/grpc/grpc/issues/12490

The PR is imitating [what Java does](https://github.com/google/google-auth-library-java/blob/cbc6c1062d5fe1920c6a641a4d63769b56267eb0/oauth2_http/java/com/google/auth/oauth2/ComputeEngineCredentials.java#L76), who has retry and 500ms timeout.

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -70,6 +70,19 @@ class GCECredentials extends CredentialsLoader
     const FLAVOR_HEADER = 'Metadata-Flavor';
 
     /**
+     * Note: the explicit `timeout` and `tries` below is a workaround. The underlying
+     * issue is that resolving an unknown host on some networks will take
+     * 20-30 seconds; making this timeout short fixes the issue, but
+     * could lead to false negatives in the event that we are on GCE, but
+     * the metadata resolution was particularly slow. The latter case is
+     * "unlikely" since the expected 4-nines time is about 0.5 seconds.
+     * This allows us to limit the total ping maximum timeout to 1.5 seconds
+     * for developer desktop scenarios.
+     */
+    const MAX_COMPUTE_PING_TRIES = 3;
+    const COMPUTE_PING_CONNECTION_TIMEOUT_S = 0.5;
+
+    /**
      * Flag used to ensure that the onGCE test is only done once;.
      *
      * @var bool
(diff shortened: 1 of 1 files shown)


src/Credentials/GCECredentials.php
```diff
@@ -126,28 +139,32 @@ public static function onGce(callable $httpHandler = null)
...
+
+                return $resp->getHeaderLine(self::FLAVOR_HEADER) == 'Google';
+            } catch (ClientException $e) {
+                // error_log("ClientException", 0);
+            } catch (ServerException $e) {
+                // error_log("ServerException", 0);
+            } catch (RequestException $e) {
+                // error_log("RequestException", 0);
+            }
+            $httpHandler = HttpHandlerFactory::build();
```
ZhouyihaiDing: Need to build a new httpHandler instead of using the old one, otherwise it will throw `OUTBOUND Exception`
https://travis-ci.org/google/google-auth-library-php/builds/388429882?utm_source=github_status&utm_medium=notification

However, if `$httpHandler` is not null and passed into this function. Should we make a copy of that here?

tmatsuo: I'm ok with the current approach

src/Credentials/GCECredentials.php
```diff
@@ -126,28 +139,32 @@ public static function onGce(callable $httpHandler = null)
...
+                // the metadata resolution was particularly slow. The latter case is
+                // "unlikely".
+                $resp = $httpHandler(
+                    new Request('GET', $checkUri),
+                    ['timeout' => self::COMPUTE_PING_CONNECTION_TIMEOUT_S]
+                );
+
+                return $resp->getHeaderLine(self::FLAVOR_HEADER) == 'Google';
+            } catch (ClientException $e) {
+                // error_log("ClientException", 0);
```
tmatsuo: Why do we have `error_log` commented out? Can we just remove them, or un-comment if it does make sense?

src/Credentials/GCECredentials.php
ZhouyihaiDing: Thanks!

src/Credentials/GCECredentials.php
ZhouyihaiDing: Done. Removed the comment.

src/Credentials/GCECredentials.php
```diff
@@ -126,28 +139,29 @@ public static function onGce(callable $httpHandler = null)
...
+                // could lead to false negatives in the event that we are on GCE, but
+                // the metadata resolution was particularly slow. The latter case is
+                // "unlikely".
+                $resp = $httpHandler(
+                    new Request('GET', $checkUri),
+                    ['timeout' => self::COMPUTE_PING_CONNECTION_TIMEOUT_S]
+                );
+
+                return $resp->getHeaderLine(self::FLAVOR_HEADER) == 'Google';
+            } catch (ClientException $e) {
```
jdpedrie: Could we replace all these catches with a single reference to [`GuzzleHttp\Exception\GuzzleException`](https://github.com/guzzle/guzzle/blob/master/src/Exception/GuzzleException.php)?

```php
catch (GuzzleException $e) {}
```

ZhouyihaiDing: I tried to catch `GuzzleException` or even the `Exception`, but it seems the Exception is not catched and break the test. https://travis-ci.org/google/google-auth-library-php/jobs/394324374

jdpedrie: If you catch `Exception`, be sure to add a backslash: `\Exception`. Odd that `GuzzleException` didn't work. You added the `use` statement at the top of the class, right?

ZhouyihaiDing: I added the `use GuzzleHttp\Exception\GuzzleException;` and removed 
```
use GuzzleHttp\Exception\ClientException;
use GuzzleHttp\Exception\RequestException;
use GuzzleHttp\Exception\ServerException;
```

tmatsuo: @jdpedrie Gentle ping. How does it look now?

PR #202, opened 2018-07-23
Only emits the warning for gcloud credentials

tmatsuo:


src/Credentials/UserRefreshCredentials.php
```diff
@@ -80,16 +85,22 @@ public function __construct(
...
-            . 'continues to use end user credentials from Cloud SDK, you might '
-            . 'receive a "quota exceeded" or "API not enabled" error. For '
-            . 'more information about service accounts, see '
-            . 'https://cloud.google.com/docs/authentication/.',
-            E_USER_WARNING
-        );
+        if ($jsonKey['client_id'] === self::CLOUD_SDK_CLIENT_ID
+            && getenv(self::SUPPRESS_CLOUD_SDK_CREDS_WARNING_ENV) !== 'true') {
+            trigger_error(
+                'Your application has authenticated using end user credentials '
```
dwsupplee: Just a small nit, is it possible to use the same indentation style as the original revision?

tmatsuo:
@dwsupplee I introduced an env var to suppress the warning. PTAL

src/Credentials/UserRefreshCredentials.php
```diff
@@ -80,16 +85,22 @@ public function __construct(
...
-            'Your application has authenticated using end user credentials '
-            . 'from Gooogle Cloud SDK. We recommend that most server '
-            . 'applications use service accounts instead. If your application '
-            . 'continues to use end user credentials from Cloud SDK, you might '
-            . 'receive a "quota exceeded" or "API not enabled" error. For '
-            . 'more information about service accounts, see '
-            . 'https://cloud.google.com/docs/authentication/.',
-            E_USER_WARNING
-        );
+        if ($jsonKey['client_id'] === self::CLOUD_SDK_CLIENT_ID
```
dwsupplee: Do you think adding a unit test for this condition adds any value?

tests/Credentials/UserRefreshCredentialsTest.php
```diff
@@ -110,6 +110,15 @@ public function testInitalizeFromAFile()
             new UserRefreshCredentials('scope/1', $keyFile)
         );
     }
+
+    public function testGcloudWarning()
+    {
+        putenv('SUPPRESS_GCLOUD_CREDS_WARNING=false');
+        $keyFile = __DIR__ . '/../fixtures2' . '/gcloud.json';
+        $this->assertNotNull(
```
dwsupplee: I believe this test should expect `PHPUnit_Framework_Error_Warning` is thrown

tmatsuo: @dwsupplee 
Yeah I wanted to make sure the test fails with that revision, now it has correct tests. PTAL

dwsupplee: Gotcha :)

tests/Credentials/UserRefreshCredentialsTest.php
```diff
@@ -110,6 +110,27 @@ public function testInitalizeFromAFile()
             new UserRefreshCredentials('scope/1', $keyFile)
         );
     }
+
+    /**
+     * @expectedException \PHPUnit\Framework\Error\Warning
```
dwsupplee: If I remember correctly depending on the version of PHPUnit installed this could not exist. There wasn't a class added for forwards compatibility like `TestCase`.

tests/Credentials/UserRefreshCredentialsTest.php
```diff
@@ -110,6 +110,27 @@ public function testInitalizeFromAFile()
...
+     */
+    public function testGcloudWarning()
+    {
+        putenv('SUPPRESS_GCLOUD_CREDS_WARNING=false');
+        $keyFile = __DIR__ . '/../fixtures2' . '/gcloud.json';
```
dwsupplee: No need for string concatenation

tests/Credentials/UserRefreshCredentialsTest.php
```diff
@@ -110,6 +110,27 @@ public function testInitalizeFromAFile()
...
+        $this->assertNotNull(
+            new UserRefreshCredentials('scope/1', $keyFile)
+        );
+    }
+
+    public function testValid3LOauthCreds()
+    {
+        putenv('SUPPRESS_GCLOUD_CREDS_WARNING=false');
+        $keyFile = __DIR__ . '/../fixtures2' . '/valid_oauth_creds.json';
```
dwsupplee: ditto

tests/Credentials/UserRefreshCredentialsTest.php
tmatsuo: Done

tests/Credentials/UserRefreshCredentialsTest.php
tmatsuo: Done

PR #208, opened 2018-09-10
Add support for InsecureCredentials

michaelbausor:
Adds support for insecure credentials, which can be used when calling APIs that don't require credentials (e.g. public buckets), emulators, or test services.

cc @tmatsuo @bshaffer @dwsupplee @jdpedrie

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 *~
 vendor
 composer.lock
+
+# IntelliJ
+.idea
+*.iml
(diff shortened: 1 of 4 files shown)


michaelbausor:
@dwsupplee I don't have merge access - if you do, can you merge? Otherwise, tagging @bshaffer and @tmatsuo

dwsupplee:
Done!

michaelbausor:
I think `EmptyCredentials` or `AnonymousCredentials` would have been fine too - do you think that would have been preferable to `InsecureCredentials`?

PR #212, opened 2018-11-01
Cache item expiration correctly calculated when DST shifts

ged15:
Hello,

During the latest DST shift we had in Europe on 28th October we experienced problems with our long-running processes failing to authenticate with GCP PubSub.

I think it was caused by a comparison that does not take the DST into account.

I split this PR into 2 commits - the failing test that demonstrates the problem we encountered and the fix.

Since manipulating time in PHP using built-in `DateTime` classes is not possible, I introduced a clock abstraction. I don't like the singleton design of it but it seemed like the simplest way to solve the problem while also keeping BC.

### DIFF
--- a/src/Cache/Item.php
+++ b/src/Cache/Item.php
@@ -81,7 +81,8 @@ final class Item implements CacheItemInterface
             return true;
         }
 
-        return Clock::now()->getTimestamp() < (clone $this->expiration)->getTimestamp();
+        return Clock::now()->setTimezone(new \DateTimeZone('UTC'))->getTimestamp()
+            < (clone $this->expiration)->setTimezone(new \DateTimeZone('UTC'))->getTimestamp();
     }
 
     /**
--- a/src/Cache/Item.php
+++ b/src/Cache/Item.php
@@ -81,8 +81,10 @@ final class Item implements CacheItemInterface
             return true;
         }
 
+        $expiration = clone $this->expiration;
+
         return Clock::now()->setTimezone(new \DateTimeZone('UTC'))->getTimestamp()
-            < (clone $this->expiration)->setTimezone(new \DateTimeZone('UTC'))->getTimestamp();
+            < $expiration->setTimezone(new \DateTimeZone('UTC'))->getTimestamp();
     }
 
     /**
--- a/src/Cache/Item.php
+++ b/src/Cache/Item.php
@@ -81,7 +81,7 @@ final class Item implements CacheItemInterface
             return true;
         }
 
-        return new \DateTime() < $this->expiration;
+        return Clock::now()->getTimestamp() < (clone $this->expiration)->getTimestamp();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


ged15:
Anybody looking into merging this?

src/Cache/Item.php
```diff
@@ -125,18 +128,18 @@ public function expiresAt($expiration)
      */
     public function expiresAfter($time)
     {
-        if (is_int($time)) {
-            $this->expiration = new \DateTime("now + $time seconds");
```
dwsupplee:
```php
$this->expiration = (new \DateTime())
    ->add(new \DateInterval("PT{$time}S"));
```

src/Cache/Item.php
```diff
@@ -81,7 +81,10 @@ public function isHit()
             return true;
         }
 
-        return new \DateTime() < $this->expiration;
```
dwsupplee:
```php
return (new \DateTime)->getTimestamp() < $this->expiration->getTimestamp();
```

dwsupplee:
Friendly ping @ged15

ged15:
Hi @dwsupplee,

Thank you for your patience and not closing this PR! As we are approaching DST shift again, it becomes crucial to fix this.

> <...> I've commented on two lines which in my testing should resolve the issue without the need for the `Clock` abstraction or any cloning of objects. What do you think of just making the changes to the two proposed lines instead?

I think it is an easier solution indeed and I considered using it initially. However, writing tests becomes more difficult then. I am good with also removing the abstraction and the test. WDYT?

ged15:
Ping @dwsupplee

dwsupplee:
I really like having the test. What if we put a `getDateTime` protected method that wraps the instantiation of `\DateTime` on the `Item` implementation? Then we could mock it out in the test still. Does that seem like a reasonable compromise?

ged15:
> What if we put a getDateTime protected method that wraps the instantiation of \DateTime on the Item implementation?

This would couple the test and the implementation severely and, in my opinion, would be a worse solution than the abstraction.

Another approach would be to use the [UOPZ](https://github.com/krakjoe/uopz) PHP extension in the test environment to override the `Datetime` class totally. But is the `Clock` abstraction really that much of a problem to warrant that? 🙂

dwsupplee:
I only push back on the abstraction because my aim is to reduce the amount of surface we expose to users and need to worry about maintaining in the future. I agree an extension would be too much here as well.

If the protected method doesn't feel right to you, I wouldn't mind going back to the idea of dropping the test case then.

@jdpedrie do you have any thoughts?

jdpedrie:
@ged15 thanks for the good conversation, and appreciate your effort at finding the best implementation. :)

I'm in the protected method overridden in the test. I recognize it may not be perfectly ideal, but it's an [existing](https://github.com/googleapis/google-cloud-php/blob/faeb39adde5d12c4225280a1dd9cd5c41e97a667/Bigtable/src/Mutations.php#L134) [convention](https://github.com/googleapis/google-cloud-php/blob/faeb39adde5d12c4225280a1dd9cd5c41e97a667/Core/src/Compute/Metadata/Readers/StreamReader.php#L85) throughout the libraries we maintain, and it's not caused any trouble.

We've used `@internal` too, but I prefer options that simply don't expand the public API, regardless of how we try to document around it.

ged15:
@dwsupplee @jdpedrie I could not figure out a way to override the method in the test while also keeping the test and code understandable. I understand this is not ideal but I see this as the least of all evils way forward. Hence I've opened #246. It should fix the DST shift problem by explicitly setting timezone to UTC.

PR #217, opened 2018-12-05
Fix failing tests

dwsupplee:
It isn't clear to me how these tests ever passed previously. 😆


dwsupplee:
Looking into the 5.4 error now.

Issue #135, opened 2016-09-27
Add method for detecting a default Project ID


bshaffer:
The python auth library has support for detecting the current `Project ID` in several different ways ([see here](https://google-auth.readthedocs.io/en/latest/_modules/google/auth/_default.html#default). If we added support for this in the PHP auth library, it could be used in `google-cloud-php`'s [`ClientTrait`](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/ClientTrait.php).

This has the advantages of DRY, and would allow us to add the method of retrieving the project ID from the gcloud command line tool and from App Engine standard, neither of which are currently supported by `google-cloud-php`.

JamesVanWaza:
@bshaffer what's the solution for the issue.

jdpedrie:
@JamesVanWaza #230 will address this. Look for it in a near-future release of the Auth library.

PR #220, opened 2019-02-18
Project Id Provider

danielgsims:
See #135

### DIFF
--- a/src/ProjectIdProvider.php
+++ b/src/ProjectIdProvider.php
@@ -1,0 +1,181 @@
+<?php
+/*
+ * Copyright 2019 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+use DomainException;
+use Google\Auth\Credentials\GCECredentials;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
+use GuzzleHttp\Psr7\Request;
+
+/**
+ * ProjectIdProvider encapsulates the behavior for determining a google-cloud project id.
+ *
+ * This class allows you to dynamically determine the application's project id. For example:
+ *
+ *   $provider = new Google\Auth\ProjectIdProvider();
+ *   $projectId = $provider->getProjectId();
+ *
+ */
+class ProjectIdProvider
+{
+    /**
+     * @var callable
+     */
+    const METADATA_SERVICE = 'http://169.254.169.254/computeMetadata/v1/project/project-id';
+
+    /**
+     * Determines the project id for the current project. If one cannot be determined, a DomainException
+     * is thrown. The project is determined from the following locations respectively:
+     *
+     *   1. A Credentials file supplied via the GOOGLE_APPLICATION_CREDENTIALS env variable
+     *   2. The default application credentials
+     *   3. The Google Cloud SDK configuration command
+     *   4. The App Engine Standard env variable
+     *   5. The Google Cloud Compute Engine Meta Data service
+     *
+     * @param callable|null $httpHandler
+     * @return string The current project's id.
+     * @throws DomainException
+     */
+    public static function getProjectId(callable $httpHandler = null)
+    {
+        // List of sources for determining a project id
+        $sources = [
+            'self::fromApplicationDefaultCredentials' => [],
+            'self::fromSdk' => [],
+            'self::fromAppEngineStandard' => [],
+            'self::fromComputeEngineMetaData' => [$httpHandler],
+        ];
+
+        // Look through all of the sources until we find a project id
+        foreach ($sources as $callable => $args) {
+            $id = call_user_func_array($callable, $args);
+            if ($id !== null) {
+                return $id;
+            }
+        }
+
+        throw new DomainException('Could not determine project id');
+    }
+
+    /**
+     * Check for a project id in the Key File specified in the GOOGLE_APPLICATION_CREDENTIALS
+     * environment variable
+     *
+     * @return string|null
+     */
+    private static function fromApplicationDefaultCredentials()
+    {
+        $credentials = CredentialsLoader::fromEnv() ?: CredentialsLoader::fromWellKnownFile();
+
+        if (!isset($credentials['project_id'])) {
+            return null;
+        }
+
+        return $credentials['project_id'];
+    }
+
+    /**
+     * Attempts to retrieve a project id from the gcloud sdk
+     *
+     * @return string|null
+     */
+    private static function fromSdk()
+    {
+        $command = 'gcloud config config-helper --format json';
+
+        // exec the command
+        $output = static::execute($command);
+
+        // if we didn't receive output, return null
+        if (!$output) {
+            return null;
+        }
+
+        // the output should be json. decode it and check for errors
+        $config = json_decode($output, true);
+
+        if (json_last_error() != JSON_ERROR_NONE) {
+            return null;
+        }
+
+        // the configuration should contain a project id at the specified path
+        if (isset($config['configuration']['properties']['core']['project'])) {
+            return $config['configuration']['properties']['core']['project'];
+        }
+    }
+
+    /**
+     * Check for the project id available in App Engine Standard Environments
+     *
+     * @return string|null
+     */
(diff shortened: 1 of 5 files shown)


danielgsims:
@dwsupplee @jdpedrie - I don't want to forget about this one. Thanks!

dwsupplee:
Thanks for the ping @danielgsims :) - this is on my agenda for the week

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+        $httpHandler = $this->httpHandler;
+
+        if (!$httpHandler) {
+            $httpHandler = HttpHandlerFactory::build();
+        }
+
+        // Make a request to the meta data service to get project id
+        // If we can't talk to the service and get a request exception, return null
+        try {
+            $response = $httpHandler(
```
dwsupplee: I think we'll need to follow the same pattern used in `GCECredentials` for talking to the metadata server, see [here](https://github.com/googleapis/google-auth-library-php/blob/master/src/Credentials/GCECredentials.php#L72-L83) for a little more detail.

tests/ProjectIdProviderTest.php
```diff
@@ -0,0 +1,215 @@
+<?php
```
dwsupplee: We should add copyright for tests too

tests/ProjectIdProviderTest.php
```diff
@@ -0,0 +1,215 @@
...
+
+namespace Google\Auth\Tests;
+
+use Google\Auth\CredentialsLoader;
+use Google\Auth\ProjectIdProvider;
+use PHPUnit\Framework\TestCase;
+
+class ProjectIdProviderShellMock extends ProjectIdProvider
```
dwsupplee: If you'd like to stick with testing private methods, WDYT about a little helper on the shell class you have here which will encapsulate the reflection/setAccessibility calls?

tests/ProjectIdProviderTest.php
```diff
@@ -0,0 +1,215 @@
...
+        // Use the dev credentials to ensure the default app credentials aren't loaded
+        putenv('HOME=' . __DIR__ . '/fixtures/alternative_defaults');
+
+        $provider->getProjectId();
+
+        // reset HOME env
+        putenv('HOME=' . $home);
+    }
+
+    public function testShellExecWrapper()
```
dwsupplee: This doesn't look to be used

jdpedrie:
friendly reminder on this, @danielgsims. :)

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+ */
+
+namespace Google\Auth;
+
+use DomainException;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
+use GuzzleHttp\Psr7\Request;
+
+/**
+ * ProjectIdProvider encapsulates the behavior for determining a google-cloud project id.
```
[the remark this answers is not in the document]
danielgsims: Sorry, for clarity, do you want me to change the doc block to `project ID` or the class name to `ProjectIDProvider` or both?

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+ *   $projectId = $provider->getProjectId();
+ *
+ */
+class ProjectIdProvider
+{
+    /**
+     * @var callable
+     */
+    private $httpHandler;
+    const METADATA_SERVICE = 'http://169.254.169.254/computeMetadata/v1/project/project-id';
```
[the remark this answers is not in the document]
danielgsims: That sounds good to me!

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+     *   1. A Credentials file supplied via the GOOGLE_APPLICATION_CREDENTIALS env variable
+     *   2. The default application credentials
+     *   3. The Google Cloud SDK configuration command
+     *   4. The App Engine Standard env variable
+     *   5. The Google Cloud Compute Engine Meta Data service
+     *
+     * @return string The current project's id.
+     * @throws DomainException
+     */
+    public function getProjectId()
```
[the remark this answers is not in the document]
danielgsims: I noticed how the rest of the library used statics after I pulled this logic in from my test application. I think that's definitely a great refactor.

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+
+        // Look through all of the sources until we find a project id
+        foreach ($sources as $callable) {
+            $id = $callable();
+            if ($id !== null) {
+                return $id;
+            }
+        }
+
+        // If a project id could not be found, throw an exception
```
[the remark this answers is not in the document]
danielgsims: Sounds good :)

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+
+        // if we didn't receive output, return null
+        if (!$output) {
+            return null;
+        }
+
+        // the output should be json. decode it and check for errors
+        $config = json_decode($output, true);
+
+        if (json_last_error() != JSON_ERROR_NONE) {
```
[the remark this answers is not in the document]
danielgsims: Good eye

src/ProjectIdProvider.php
jdpedrie: For what it's worth, GCECredentials doesn't apply those request settings to the real metadata calls. The retry count and timeout is used for the preflight request to check whether the application is on compute engine.

I think that rather than adding the timeout, this PR should call `GCECredentials::onGce()` prior to making the call to get the project ID. That way, future tweaks to make that method more reliable will cascade down to the method without us needing to keep things in sync.

tests/ProjectIdProviderTest.php
jdpedrie: Can you put test classes at the bottom of the file? You'll also need to wrap them in phpcs ignores:

```php
//@codingStandardsIgnoreStart
$stuff;
//@codingStandardsIgnoreEnd
```

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+        throw new DomainException('Could not determine project id');
+    }
+
+    /**
+     * Check for a project id in the Key File specified in the GOOGLE_APPLICATION_CREDENTIALS
+     * environment variable
+     *
+     * @return string|null
+     */
+    private function fromKeyFilePath()
```
jdpedrie: Both this method and the next are part of the [ApplicationDefaultCredentials strategy](https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application). In my mind, this name implies a user-supplied keyFilePath.

Perhaps `fromADCEnv` and `fromADCWellKnownFile`, or combine the two into `fromApplicationDefaultCredentials` and use the same strategy as demonstrated in `ApplicationDefaultCredentials::getCredentials()`?

```php
$jsonKey = CredentialsLoader::fromEnv()
    ?: CredentialsLoader::fromWellKnownFile();
```

src/ProjectIdProvider.php
```diff
@@ -0,0 +1,206 @@
...
+     *
+     * @return string|null
+     */
+    private function fromAppEngineStandard()
+    {
+        // app engine should contain the app id environment variable
+        $appId = getenv('APPLICATION_ID');
+
+        // if we didn't find it, return null
+        if (!$appId) {
```
jdpedrie: How about using `AppIdentityCredentials::onAppEngine()` instead?

src/ProjectIdProvider.php
dwsupplee: > this PR should call GCECredentials::onGce() prior to making the call to get the project ID.

👍

dwsupplee:
Friendly ping @danielgsims

src/ProjectIdProvider.php
danielgsims: I didn't apply this change in the last revision. I needed the env var anyway to parse the app id. Do you think I should add onAppEngine as an initial check?

src/ProjectIdProvider.php
danielgsims: Applied this change.

src/ProjectIdProvider.php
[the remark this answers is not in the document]
danielgsims: Since this wasn't used elsewhere, I just used an inline string, unless you'd prefer a class const

src/ProjectIdProvider.php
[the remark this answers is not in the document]
danielgsims: @dwsupplee - I think this is my last outstanding question here.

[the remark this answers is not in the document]
dwsupplee: Just the doc block please

src/ProjectIdProvider.php
jdpedrie: `APPLICATION_ID` seems rather generic; would this behave properly if someone were to set it themselves in a form different from the way App Engine uses it?

danielgsims: You're right, it would definitely be wonky. I added the App Engine check!

jdpedrie:
Closed in favor of #230.

PR #221, opened 2019-03-27
Add support for signing strings with a Credentials instance.

jdpedrie:
This change adds `Google\Auth\SignBlobInterface`, which provides a method for signing arbitrary bytes using a credentials instance.

The signing strategy depends on the credentials type.

* AppIdentityCredentials uses the AppIdentityService.
* GCECredentials uses [IAM signing](https://cloud.google.com/iam/credentials/reference/rest/v1/projects.serviceAccounts/signBlob).
* ServiceAccountCredentials and its JWT variant use local signing with the private key in the service account.

I've tested locally and in various google compute environments. an overview is below:

```
Flex:
    GCECredentials
    IAM

Standard 5.5
    AppIdentityCredentials
    AppIdentityService

Standard 7.2
    GCECredentials
    IAM

Compute Engine
    GCECredentials
    IAM
```

The reason for this change is to support work on Signed URLs in Google Cloud PHP, and to fill a gap in our library which is covered by other auth clients like Python and Node.


src/Credentials/GCECredentials.php
```diff
@@ -233,4 +265,87 @@ public function getLastReceivedToken()
 
         return null;
     }
+
+    /**
+     * Get the Client ID from GCE metadata.
```
dwsupplee: Why use the nomenclature `clientName` as opposed to `Client ID` as found here?

src/Credentials/GCECredentials.php
```diff
@@ -233,4 +265,87 @@ public function getLastReceivedToken()
...
+        $httpHandler = HttpHandlerFactory::build(HttpClientCache::getHttpClient());
+
+        // Providing a signer is useful for testing, but it's undocumented
+        // because it's not something a user would generally need to do.
+        $signer = $this->iam ?: new Iam($httpHandler);
+
+        $email = $this->getClientName($httpHandler);
+
+        $previousToken = $this->getLastReceivedToken();
+        if ($previousToken) {
```
dwsupplee: This would be a good opportunity for a ternary :).

src/HttpHandler/HttpHandlerFactory.php
```diff
@@ -24,24 +24,25 @@ class HttpHandlerFactory
     /**
      * Builds out a default http handler for the installed version of guzzle.
      *
-     * @param ClientInterface $client
+     * If a handler has been previously created, it will be returned instead
```
dwsupplee: This is outdated with the latest commits, could we please revert this as well as where we move `$version = ClientInterface::VERSION;`?

src/ServiceAccountSignerTrait.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Sign a string using a Service Account private key.
+ */
+trait ServiceAccountSignerTrait
+{
+    /**
+     * Sign a string using the service account private key.
+     *
+     * @param string $stringToSign
+     * @param array bool $forceOpenssl Whether to use OpenSSL regardless of
```
dwsupplee:
```suggestion
     * @param bool $forceOpenssl Whether to use OpenSSL regardless of
```

tests/Credentials/AppIdentityCredentialsTest.php
```diff
@@ -0,0 +1,229 @@
+<?php
```
dwsupplee: Nice catch :)

src/HttpHandler/HttpClientCache.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * Cache an HTTP Client for later calls.
+     *
+     * Passing null will unset the cached client.
+     *
+     * @param ClientInterface|null $client
+     * @return void
+     */
+    public static function setHttpClient(ClientInterface $client = null)
```
dwsupplee: WDYT about calling this somewhere which is a hot path in the code?

src/FetchAuthTokenCache.php
```diff
@@ -105,4 +105,38 @@ public function getLastReceivedToken()
...
+     * @return string The resulting signature.
+     * @throws \RuntimeException If the fetcher does not implement
+     *     `Google\Auth\SignBlobInterface`.
+     */
+    public function signBlob($stringToSign, $forceOpenSsl =  false)
+    {
+        if (!$this->fetcher instanceof SignBlobInterface) {
+            throw new \RuntimeException(
+                'Credentials fetcher does not implement ' .
+                'Google\Auth\SignBlobInterface'
```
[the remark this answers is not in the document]
jdpedrie: Not until we drop PHP 5.4 support. :(

src/Credentials/AppIdentityCredentials.php
```diff
@@ -100,25 +106,16 @@ public static function onAppEngine()
...
+     *         - access_token (string)
+     *         - expiration_time (string)
      */
     public function fetchAuthToken(callable $httpHandler = null)
     {
         if (!self::onAppEngine()) {
             return array();
         }
 
-        if (!class_exists('google\appengine\api\app_identity\AppIdentityService')) {
```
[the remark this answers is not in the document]
jdpedrie: No, you're correct; the check wasn't sufficient. I modified it, PTAL.

src/Credentials/InsecureCredentials.php
```diff
@@ -67,4 +67,15 @@ public function getLastReceivedToken()
     {
         return $this->token;
     }
+
+    /**
+     * Get the client ID. In this case, it returns an empty string.
```
dwsupplee: Small nit, let's stay consistent in our use of `client name` across these docs.

jdpedrie: done, PTAL/

PR #228, opened 2019-05-08
Load environment data as JSON

alexeydemin:
When service application json key located inside Kubernetes secrets it might be useful to load them directly from there.

### DIFF
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -66,6 +66,9 @@ abstract class CredentialsLoader implements FetchAuthTokenInterface
     public static function fromEnv()
     {
         $path = getenv(self::ENV_VAR);
+        if ($credentials = json_decode($path, true)) {
+            return $credentials;
+        }
         if (empty($path)) {
             return;
         }


alexeydemin:
I signed it!

jdpedrie:
Hi @alexeydemin,

If you already have the JSON key stored in an environment variable, you can simply call `CredentialsLoader::makeCredentials()` with that data. There's no need to call `fromEnv()`.

Additionally, why would you store it in `GOOGLE_APPLICATION_CREDENTIALS`? Is that a convention, or have you simply chosen it for simplicity's sake? If it's possible to use something else, I'd advise it, as many Google libraries expect that variable to contain a path to the keyfile, not the keyfile data itself, and the behavior on encountering JSON there would be unexpected. I [do not see any precedent](https://cloud.google.com/docs/authentication/production) for that variable containing the keyfile data itself.

Please let me know if I'm missing anything, or misunderstanding your intent.

alexeydemin:
That makes sense, @jdpedrie this PR may be closed. Thank you for your explanations.

Issue #225, opened 2019-04-22
Race condition in SysVCacheItemPool constructor


ericnorris:
I am using the [CacheSessionPool](https://googleapis.github.io/google-cloud-php/#/docs/google-cloud/master/spanner/session/cachesessionpool) in the Google Cloud Spanner PHP library for caching Spanner sessions.

In order to construct the `CacheSessionPool`, you are required to pass in a `SysVCacheItemPool`. Although the `CacheSessionPool` uses locking when interacting with the `SysVCacheItemPool`, there appears to be a race condition in `__construct` meaning you probably need to do locking yourself in order to create the object.

`loadItems` is called in `__construct`: https://github.com/googleapis/google-auth-library-php/blob/v1.4.0/src/Cache/SysVCacheItemPool.php#L135

`loadItems` then calls `shm_attach`: https://github.com/googleapis/google-auth-library-php/blob/v1.4.0/src/Cache/SysVCacheItemPool.php#L88-L92.

`shm_attach` attempts to 1) get the shared memory segment if it exists 2) create it if not: https://github.com/php/php-src/blob/PHP-7.2/ext/sysvshm/sysvshm.c#L171-L182.

If two processes attempt to construct a `SysVCacheItemPool` at the same time, see that there is not an existing shared memory segment, and then try to create one with `IPC_CREAT | IPC_EXCL` the `WARNING` on L178 (https://github.com/php/php-src/blob/PHP-7.2/ext/sysvshm/sysvshm.c#L177-L180) will occur.

Normally this may be okay, but I am running in an environment where I convert warnings to exceptions.

This could maybe be solved by not calling `loadItems` in the constructor, or I could file an issue in the https://github.com/googleapis/google-cloud-php repository to see what changes may need to be made there.

#### Environment details

  - OS: Container-Optimized OS v75
  - PHP version: PHP 7.2.11
  - Package name and version: 
    - google/auth@v1.4.0
    - google/cloud-spanner@1.11

#### Steps to reproduce

Not straightforward, since it is a race condition. You could try creating many `SysVCacheItemPool`s in a tight loop in multiple processes.

dwsupplee:
Thanks for the report @ericnorris, your proposal of moving `loadItems` out of the constructor seems reasonable to me. :)

It looks like we may just need to ensure we trigger `loadItems` in the `delete`/`save` methods as well if we go this route. Does this sound reasonable to you, @tmatsuo?

tmatsuo:
[...]

PR #229, opened 2019-05-16
Move loadItems out of constructor to avoid race condition

dwsupplee:
Closes: https://github.com/googleapis/google-auth-library-php/issues/225

The flag was added in order to closely simulate the same behavior we had before with `loadItems` being called in the constructor.


dwsupplee:
I added the conditional so the behavior would match closely with the call to `loadItems` in the constructor - so I don't believe we want to add it everywhere.

PR #230, opened 2019-05-24
feat: Add getProjectId to credentials types which support it

### LINKED ISSUE #135: Add method for detecting a default Project ID

jdpedrie:
Like a cross between #221 and #220. Provides for detection of a project ID on a Credentials instance.

Tested on GAE Standard 5.5 and 7.2 and Flex (which uses GCECredentials).

Closes #135.

### DIFF
--- a/src/Credentials/AppIdentityCredentials.php
+++ b/src/Credentials/AppIdentityCredentials.php
@@ -24,6 +24,7 @@ namespace Google\Auth\Credentials;
  */
 use google\appengine\api\app_identity\AppIdentityService;
 use Google\Auth\CredentialsLoader;
+use Google\Auth\ProjectIdProviderInterface;
 use Google\Auth\SignBlobInterface;
 
 /**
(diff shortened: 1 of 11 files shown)


jdpedrie:
@dwsupplee 	can you take a look at this this week? :)

dwsupplee:
Will do!

jdpedrie:
@dwsupplee I addressed your comments. PTAL? :)

jdpedrie:
@bshaffer done!

PR #239, opened 2019-08-09
feat: Support custom scopes on GCECredentials

### LINKED ISSUE #238: GCECredentials does not support custom scopes
#### Steps to reproduce

```php
    $this->googleClient = new Google_Client();
    $this->googleClient->useApplicationDefaultCredentials();
    $this->googleClient->setScopes(Google_Service_Sheets::SPREADSHEETS);
    $this->sheetsClient = new Google_Service_Sheets($this->googleClient);
    $body = new Google_Service_Sheets_ValueRange();
    $body->setRange("A1");
    $body->setValues([['spaghetti', 'potato']]);
    $this->sheetsClient->spreadsheets_values->update($sheetId, "A1", $body, ['valueInputOption' => 'RAW']);
```

Result:

```json
{
  "error": {
    "code": 403,
    "message": "Request had insufficient authentication scopes.",
    "errors": [
      {
        "message": "Request had insufficient authentication scopes.",
        "domain": "global",
        "reason": "forbidden"
      }
    ],
    "status": "PERMISSION_DENIED"
  }
}
```

The issue is that the `GCECredentials` class is not provided the desired scopes to pass along when asking for a token.

The fact you can even ask for custom scopes from the GCE metadata endpoint is poorly documented. It's not documented in the [main documentation page for GCE metadata](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#applications), but it *is* [documented in Cloud Run](https://cloud.google.com/run/docs/securing/service-identity#access_tokens). It also mentions that this magic-ish feature is only available for Cloud Functions, Cloud Run services, and App Engine.

jdpedrie:
This change adds support for custom scopes provided to GCECredentials, closes #238.

This change has been verified in Cloud Run. Note that as [documented](https://cloud.google.com/run/docs/securing/service-identity#access_tokens), the feature is available only in certain limited environments, including App Engine, Run and Functions. The App Engine context, in my testing, does not appear to be inclusive of the Flex environment.

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -121,11 +121,19 @@ class GCECredentials extends CredentialsLoader implements SignBlobInterface
     private $iam;
 
     /**
-     * @param Iam $iam [optional] An IAM instance.
+     * @var string|array|null
      */
-    public function __construct(Iam $iam = null)
+    private $scope;
+
+    /**
+     * @param Iam $iam [optional] An IAM instance.
+     * @param string|array $scope [optional] the scope of the access request,
+     *        expressed either as an array or as a space-delimited string.
+     */
+    public function __construct(Iam $iam = null, $scope = null)
     {
         $this->iam = $iam;
+        $this->scope = $scope;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Credentials/GCECredentials.php
```diff
@@ -235,7 +243,19 @@ public function fetchAuthToken(callable $httpHandler = null)
             return array();  // return an empty array with no access token
         }
 
-        $json = $this->getFromMetadata($httpHandler, self::getTokenUri());
+        $tokenUri = self::getTokenUri();
+        if ($this->scope) {
```
dwsupplee: WDYT about making sure we don't need to execute this computation every time `fetchAuthToken` is called?

jdpedrie: Sure, i could move it to a property created in the constructor.

dwsupplee: Sounds good to me

jdpedrie: Done, PTAL!

PR #244, opened 2019-09-23
feat: accept json string in `GOOGLE_APPLICATION_CREDENTIALS` var

StephaneBour:
Add support for a string in the environment variable `GOOGLE_APPLICATION_CREDENTIALS`, in addition to support for a disk file path.

```php
putenv('GOOGLE_APPLICATION_CREDENTIALS={"private_key_id": "key123","private_key": "privatekey","client_email": "hello@youarecool.com","client_id": "client123","type": "service_account"}');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ Once downloaded, store the path to this file in the
 putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
 ```
 
+or put the contents of the file directly into the environment variable
+
+```php
+putenv('GOOGLE_APPLICATION_CREDENTIALS={"private_key_id": "key123","private_key": "privatekey","client_email": "hello@youarecool.com","client_id": "client123","type": "service_account"}');
+```
+
 > PHP's `putenv` function is just one way to set an environment variable.
 > Consider using `.htaccess` or apache configuration files as well.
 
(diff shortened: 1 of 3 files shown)


StephaneBour:
@googlebot I signed it!

jdpedrie:
Hi @StephaneBour,

`GOOGLE_APPLICATION_CREDENTIALS` is a well-defined convention intended to be used to indicate a path to a service account file. Placing the contents of the key directly in the value is not correct behavior.

You are free to place your credentials in an environment variable of your own choosing, and may provide it as the second argument to [`CredentialsLoader::makeCredentials()`](https://github.com/googleapis/google-auth-library-php/blob/a84aa36d925d2d57f967ba3d52f02114bc2f0e70/src/CredentialsLoader.php#L116).

See #228 for previous discussion of a similar proposal. As this isn't the first time such a change has been proposed, I'm curious as to your use case. Is there a specific reason you considered this which may not be covered by my answer?

StephaneBour:
Hello @jdpedrie,

Thanks for your reply. Indeed, it is for the same use case as #228.

I'm going to look at the makeCredentials() side (but a pity that a method does not exist clearly to load the json since env).

Thank you !

jdpedrie:
Great! And thank you for the pull request, even if it's not one we are able to accept. We appreciate contributions, especially ones like yours with tests! :laughing:

Let us know if you have any more questions.

PR #248, opened 2019-10-23
feat: Adds ID token to auth token methods

bshaffer:
This uses existing `Auth Token` methods to also return ID tokens

Adding a fifth parameter is not ideal, but there isn't really a better way to do it without either

 1. Refactoring several methods to take an options array as the first parameter
 2. Breaking B.C. and bumping this library a major version

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,22 +133,23 @@ use GuzzleHttp\HandlerStack;
 // specify the path to your application credentials
 putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
 
-// Provide the target audience. This can be a Client ID associated with an IAP application,
+// Provide the ID token audience. This can be a Client ID associated with an IAP application,
 // Or the URL associated with a CloudRun App
-//    $targetAudience = 'IAP_CLIENT_ID.apps.googleusercontent.com';
-//    $targetAudience = 'https://service-1234-uc.a.run.app';
-$targetAudience = 'YOUR_TARGET_AUDIENCE';
+//    $idTokenAudience = 'IAP_CLIENT_ID.apps.googleusercontent.com';
+//    $idTokenAudience = 'https://service-1234-uc.a.run.app';
+$idTokenAudience = 'YOUR_ID_TOKEN_AUDIENCE';
 
 // create middleware
-$middleware = ApplicationDefaultCredentials::getMiddleware(null, null, null, null, $targetAudience);
+$middleware = ApplicationDefaultCredentials::getMiddleware(null, null, null, null, $idTokenAudience);
 $stack = HandlerStack::create();
 $stack->push($middleware);
 
 // create the HTTP client
 $client = new Client([
   'handler' => $stack,
-  'base_uri' => $targetAudience,
-  'auth' => 'google_auth'
+  'auth' => 'google_auth',
+  // Cloud Run, IAP, or custom resource URL
+  'base_uri' => 'https://YOUR_PROTECTED_RESOURCE',
 ]);
 
 // make the request
--- a/README.md
+++ b/README.md
@@ -133,9 +133,11 @@ use GuzzleHttp\HandlerStack;
 // specify the path to your application credentials
 putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
 
-// Provide the target audience, for example the Client ID associated with an IAP
-// application
-$targetAudience = 'IAP_CLIENT_ID.apps.googleusercontent.com';
+// Provide the target audience. This can be a Client ID associated with an IAP application,
+// Or the URL associated with a CloudRun App
+//    $targetAudience = 'IAP_CLIENT_ID.apps.googleusercontent.com';
+//    $targetAudience = 'https://service-1234-uc.a.run.app';
+$targetAudience = 'YOUR_TARGET_AUDIENCE';
 
 // create middleware
 $middleware = ApplicationDefaultCredentials::getMiddleware(null, null, null, null, $targetAudience);
@@ -145,8 +147,8 @@ $stack->push($middleware);
 // create the HTTP client
 $client = new Client([
   'handler' => $stack,
-  'base_uri' => 'https://IAP_PROJECT_ID.googleapis.com',
-  'auth' => 'google_id_token'
+  'base_uri' => $targetAudience,
+  'auth' => 'google_auth'
 ]);
 
 // make the request
--- a/README.md
+++ b/README.md
@@ -123,6 +123,38 @@ $subscriber = ApplicationDefaultCredentials::getSubscriber($scopes);
 $client->getEmitter()->attach($subscriber);
 
 ```
+#### Call using an ID Token
+
+```php
+use Google\Auth\ApplicationDefaultCredentials;
+use GuzzleHttp\Client;
+use GuzzleHttp\HandlerStack;
+
+// specify the path to your application credentials
+putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
+
+// Provide the target audience, for example the Client ID associated with an IAP
+// application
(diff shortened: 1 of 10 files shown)


bshaffer:
Other language implementations:

 ### Python

- [`IDTokenCredentials`](https://github.com/googleapis/google-auth-library-python/blob/master/docs/user-guide.rst#identity-tokens): The method `IDTokenCredentials` is called on the `compute_engine` or `service_account` modules, as opposed to calling `Credentials`, which is done when using access tokens:

```python
from google.oauth2 import service_account

target_audience = 'https://example.com'

creds = service_account.IDTokenCredentials.from_service_account_file(
        '/path/to/svc.json',
        target_audience=target_audience)

# make authenticated request 
authed_session = AuthorizedSession(creds)
resp = authed_session.get(url)
```

### Java

- [`IdTokenCredentials`](https://github.com/googleapis/google-auth-library-java/blob/master/oauth2_http/java/com/google/auth/oauth2/IdTokenCredentials.java#L51): The `IdTokenCredentials` class extends a base `OAuth2Credentials` class, and requires an `IdTokenProvider` interface object, which is implemented by the `ServiceAccountCredentials` and `ComputeEngineCredentials` classes.

```java
GoogleCredentials adcCreds = GoogleCredentials.getApplicationDefault();
IdTokenCredentials tokenCredential = IdTokenCredentials.newBuilder()
    .setIdTokenProvider(adcCreds)
    .setTargetAudience(targetAudience).build();

// Make the HTTP request
GenericUrl genericUrl = new GenericUrl("https://example.com");
HttpCredentialsAdapter adapter = new HttpCredentialsAdapter(tokenCredential);
HttpTransport transport = new NetHttpTransport();
HttpRequest request = transport.createRequestFactory(adapter).buildGetRequest(genericUrl);
HttpResponse response = request.execute();
```

src/ApplicationDefaultCredentials.php
```diff
@@ -165,7 +167,88 @@ public static function getCredentials(
...
+            throw new DomainException(self::notFound());
+        }
+        if (!is_null($cache)) {
+            $creds = new FetchAuthTokenCache($creds, $cacheConfig, $cache);
+        }
+        return $creds;
+    }
+
+    /**
+     * Obtains an AuthTokenMiddleware that uses the default FetchAuthTokenInterface
```
jdpedrie: This sentence doesn't make total sense. :)

edit: I see now it's copied from `getMiddleware` and `getCredentials`. Can we fix all instances?

Can this also highlight how the result differs from `getMiddleware`?

src/ApplicationDefaultCredentials.php
```diff
@@ -165,7 +167,88 @@ public static function getCredentials(
...
+     * implementation to use in this environment.
+     *
+     * If supplied, $scope is used to in creating the credentials instance if
+     * this does not fallback to the compute engine defaults.
```
jdpedrie: `$scope` is not an available argument to this method.

src/ApplicationDefaultCredentials.php
```diff
@@ -165,7 +167,88 @@ public static function getCredentials(
...
+        CacheItemPoolInterface $cache = null
+
+    ) {
+        $creds = self::getCredentials($targetAudience, $httpHandler, $cacheConfig, $cache);
+
+        return new AuthTokenMiddleware($creds, $httpHandler);
+    }
+
+    /**
+     * Obtains the default FetchAuthTokenInterface implementation to use
```
jdpedrie: Can we explain how this is different from `getCredentials`?

src/ApplicationDefaultCredentials.php
```diff
@@ -165,7 +167,88 @@ public static function getCredentials(
...
+            }
+
+            $httpHandler = HttpHandlerFactory::build($client);
+        }
+
+        if (!is_null($jsonKey)) {
+            if ($jsonKey['type'] == 'authorized_user') {
+                throw new InvalidArgumentException('ID tokens are not supported for end user credentials');
+            } elseif ($jsonKey['type'] != 'service_account') {
+                throw new InvalidArgumentException('invalid value in the type field');
```
jdpedrie: This error seems cryptic. Can we get rid of the previous throw and say "only service accounts are supported" or something to that effect?

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -75,11 +75,13 @@ class ServiceAccountCredentials extends CredentialsLoader implements SignBlobInt
...
      * @param string $sub an email address account to impersonate, in situations when
      *   the service account has been delegated domain wide access.
+     * @param string $targetAudience The audience for the ID token.
      */
     public function __construct(
         $scope,
         $jsonKey,
-        $sub = null
+        $sub = null,
+        $targetAudience = null
```
jdpedrie: In `GCECredentials::__construct()`, we throw if `$scope` and `$targetAudience` are both supplied, but here we don't. Is that because `$scope` isn't optional in the signature? If there aren't other reasons, should we just tell users to set `$scope` to `null` if they provide `$targetAudience`? Nasty but probably clearer and less prone to accidentally misusing the method and getting something unexpected.

README.md
```diff
@@ -121,7 +121,42 @@ $client = new Client([
 // create subscriber
 $subscriber = ApplicationDefaultCredentials::getSubscriber($scopes);
 $client->getEmitter()->attach($subscriber);
+```
+
+#### Call using an ID Token
```
jdpedrie: Can we add a bit of information about what this is used for?

src/Credentials/ServiceAccountCredentials.php
bshaffer: When both scope and target audience are set, the OAuth2 Token endpoint returns the following 400 error:

```json
{
  "error": "invalid_request",
  "error_description": "Cannot specify both scope and target audience."
}
```

So really, it's just an oversight and I can add the exception here as well!

src/ApplicationDefaultCredentials.php
bshaffer: That's essentially what the first exception means. This means something different, specifically that the the "type" attribute of the `credentials.json` file is set to an unknown value. 

I think we should keep it because it matches the exception we throw in `CredentialsLoader::makeCredentials`. I essentially copied that method into this one, but I didn't think it made a lot of sense to add a `CredentialsLoader::makeIdTokenCredentials`. I'm happy to add it for consistency:

```php
    /**
     * Create a new Credentials instance.
     *
     * @param string|array $targetAudience
     * @param array $jsonKey the JSON credentials.
     *
     * @return ServiceAccountCredentials
     */
    public static function makeIdTokenCredentials($targetAudience, array $jsonKey)
    {
        if (!array_key_exists('type', $jsonKey)) {
            throw new \InvalidArgumentException('json key is missing the type field');
        }

        if ($jsonKey['type'] == 'authorized_user') {
            throw new InvalidArgumentException('ID tokens are not supported for end user credentials');
        }

        if ($jsonKey['type'] == 'service_account') {
            return new ServiceAccountCredentials($scope, $jsonKey);
        }

        throw new \InvalidArgumentException('invalid value in the type field');
    }
```

jdpedrie: thanks for the explanation, I'm good with it as it exists now!

bshaffer:
@jdpedrie all your comments have been addressed, PTAL!

tests/ApplicationDefaultCredentialsTest.php
```diff
@@ -200,10 +200,122 @@ public function testSuccedsIfNoDefaultFilesButIsOnGCE()
...
+    {
+        $this->originalHome = getenv('HOME');
+    }
+
+    protected function tearDown()
+    {
+        if ($this->originalHome != getenv('HOME')) {
+            putenv('HOME=' . $this->originalHome);
+        }
+        putenv(ServiceAccountCredentials::ENV_VAR);  // removes it from
```
dwsupplee:
```suggestion
        putenv(ServiceAccountCredentials::ENV_VAR);  // removes environment variable
```

PR #251, opened 2019-11-05
fix: standardize behavior between JWT versions to conform to documentation

### LINKED ISSUE #250: verifyIdToken not giving payload for valid id_token
Hi,

I was using Google Auth library to get data.
I was trying to verify Id Token but not giving payload return

I was using like this 
$payload = $oauth2->verifyIdToken();

It is giving me error **Key may not be emptySome Error Occured!!!**

Can you please see why it is like this?

Thanks,

jdpedrie:
In the JWT dependency, payloads were returned regardless of whether a verification key was provided until v3.0.0. The documentation for `Google\Auth\OAuth2::verifyIdToken()` states that if no key is provided, the unverified payload is returned. Depending on the version of JWT installed (and almost certainly the version installed will be greater than 2.0), this is not the case.

This change skips signature verification and returns the payload if no key is provided by the user.

It may be that the alternative behavior, to always fail if no key is given, is preferable, but taking that decision would require more discussion of potential breaking changes.

Fixes #250.

References:
* [Skip verification if no key provided (JWT v2.0.0)](https://github.com/firebase/php-jwt/blob/v2.0.0/Authentication/JWT.php#L58)
* [Throw exception if no key provided (JWT >= v3.0.0)](https://github.com/firebase/php-jwt/blob/v3.0.0/src/JWT.php#L62-L64)

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -17,6 +17,9 @@
 
 namespace Google\Auth;
 
+use Firebase\JWT\BeforeValidException;
+use Firebase\JWT\ExpiredException;
+use Firebase\JWT\SignatureInvalidException;
 use Google\Auth\HttpHandler\HttpClientCache;
 use Google\Auth\HttpHandler\HttpHandlerFactory;
 use GuzzleHttp\Psr7;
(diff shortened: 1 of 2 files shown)


bshaffer:
I would prefer to not change this behavior. Although inconsistent between versions of `firebase/php-jwt`, and inconsistent with the documentation, changing the behavior would be a BC break and could potentially put applications at risk.

Also, Firebase 3.0 is so old practically speaking I doubt anyone uses it. I am more in favor of updating the documentation to match the current behavior, with a disclaimer for Firebase 3.0 and a note that we will fix this in the next major version.

jdpedrie:
See #278.

PR #254, opened 2019-11-20
feat: add support for x-goog-user-project

bshaffer:
Just typed up a possible implementation for the `x-goog-user-project` header implementation.

Other considerations

 - [ ] Implement a `GetQuotaProjectTrait`

### DIFF
--- a/src/Credentials/ServiceAccountCredentials.php
+++ b/src/Credentials/ServiceAccountCredentials.php
@@ -18,6 +18,7 @@
 namespace Google\Auth\Credentials;
 
 use Google\Auth\CredentialsLoader;
+use Google\Auth\GetQuotaProjectInterface;
 use Google\Auth\OAuth2;
 use Google\Auth\ServiceAccountSignerTrait;
 use Google\Auth\SignBlobInterface;
@@ -55,7 +56,7 @@ use Google\Auth\SignBlobInterface;
  *
  *   $res = $client->get('myproject/taskqueues/myqueue');
  */
-class ServiceAccountCredentials extends CredentialsLoader implements SignBlobInterface
+class ServiceAccountCredentials extends CredentialsLoader implements SignBlobInterface, GetQuotaProjectInterface
 {
     use ServiceAccountSignerTrait;
 
@@ -65,6 +66,11 @@ class ServiceAccountCredentials extends CredentialsLoader implements SignBlobInt
      * @var OAuth2
      */
     protected $auth;
+
+    /**
+     * The quota project associated with the JSON credentials
+     */
+    protected $quotaProject;
 
     /**
      * Create a new ServiceAccountCredentials.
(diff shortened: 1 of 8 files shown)


src/Credentials/UserRefreshCredentials.php
```diff
@@ -85,16 +91,19 @@ public function __construct(
...
             && getenv(self::SUPPRESS_CLOUD_SDK_CREDS_WARNING_ENV) !== 'true') {
             trigger_error(
                 'Your application has authenticated using end user credentials '
-                . 'from Google Cloud SDK. We recommend that most server '
-                . 'applications use service accounts instead. If your '
-                . 'application continues to use end user credentials '
-                . 'from Cloud SDK, you might receive a "quota exceeded" '
-                . 'or "API not enabled" error. For more information about '
-                . 'service accounts, see '
+                . 'from Google Cloud SDK. We recommend setting the '
```
tmatsuo: I think, gcloud sdk will automatically add `quota_project` field and user should not do anything about it.

This means that, probably the warning message is something like "use more recent version of gcloud sdk" or something.

@broady 

Do you have any advise on the warning message here?

bshaffer: Ahh yes, that makes sense. "Please upgrade to the latest version of Google Cloud SDK"

tmatsuo: Yeah, and depending on the gcloud sdk's implementation, you'll also have to re-create the auth file (I'm not sure how it's implemented).

broady: Great question, I think we should point people to a page on c.g.c that gives you resolution steps.

It's probably "upgrade gcloud and run `gcloud auth application-default login` again"

We will also have a command like `gcloud auth application-default set-quota-project XXX`.

src/Middleware/AuthTokenMiddleware.php
```diff
@@ -123,4 +131,11 @@ private function fetchToken()
             return $auth_tokens['access_token'];
         }
     }
+
+    private function getQuotaProject()
+    {
+        if ($this->fetcher instanceof GetQuotaProjectInterface) {
```
broady: Are there any cases where things get wrapped and this interface would get lost?

e.g. 
```
$urc instanceof UserRefreshCredentials // true
$cred instanceof GetQuotaProjectInterface // true
$cred = cacheWrapper($urc)
$cred instanceof GetQuotaProjectInterface // false
```

bshaffer: Good catch! Yes, I absolutely need to add this to the cache wrapper

src/Credentials/UserRefreshCredentials.php
broady: Hold off on changing the error message to something that references gcloud. There'll be a separate timeline for updating docs, samples, tutorials, error messages.

broady: (I'd just revert this block)

bshaffer:
Can I bother someone to review / approve this?

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -67,6 +68,11 @@ class ServiceAccountCredentials extends CredentialsLoader implements SignBlobInt
      */
     protected $auth;
 
+    /**
+     * The quota project associated with the JSON credentials
```
jdpedrie: nit: (applies to other credential types as well)

```suggestion
     * The quota project associated with the JSON credentials
     *
     * @var string
```

bshaffer: > nit: (applies to other credential types as well)

This is true, but in the context of `ServiceAccountCredentials`, it will always be from JSON credentials. Would it be better as `The quota project associated with the service account credentials`?

PR #255, opened 2019-12-21
feat: Adds support for ES256 in AccessToken::verify

bshaffer:
Verify an IAP token:

```php
use Google\Auth\AccessToken;

$token = new AccessToken();
$result = $token->verify($iapJwt, [
    'certsLocation' => AccessToken::IAP_CERT_URL
]);
```

Some points of interest:

 - The cache key, when `certsLocation` is set, is now a `sha1` of the location, to prevent collisions with multiple cert types
 - Added a `cacheKey` option to override this in the `$options` array argument to `verify`
 - The exceptions have changed slightly - when `IAP_CERT_URL` is provided as the `certLocation`, the message `federated sign-on certs expect "keys" to be set` now reads `certs expect "keys" to be set`. This is because the IAP certs are not federated sign-on certs.
 - `ES256` verification requires the use of the library `kelvinmo/simplejwt`, which is in a pre-1.0 state.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
     "friendsofphp/php-cs-fixer": "^1.11",
     "phpunit/phpunit": "^4.8.36|^5.7",
     "sebastian/comparator": ">=1.2.3",
-    "phpseclib/phpseclib": "^2"
+    "phpseclib/phpseclib": "^2",
+    "kelvinmo/simplejwt": "^0.2.5"
   },
   "suggest": {
     "phpseclib/phpseclib": "May be used in place of OpenSSL for signing strings or for token management. Please require version ^2."
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -106,7 +106,6 @@ class AccessToken
             : $this->getCacheKeyFromCertLocation($certsLocation);
 
         // Check signature against each available cert.
-        // allow the loop to complete unless a known bad result is encountered.
         $certs = $this->getCerts($certsLocation, $cacheKey, $options);
         $alg = $this->determineAlg($certs);
 
@@ -123,6 +122,14 @@ class AccessToken
         }
     }
 
+    /**
+     * Identifies the expected algorithm to verify by looking at the "alg" key
+     * of the provided certs.
+     *
+     * @param array $certs Certificate array according to the JWK spec (see
+     *                     https://tools.ietf.org/html/rfc7517).
+     * @return string The expected algorithm, such as "ES256" or "RS256".
+     */
     private function determineAlg($certs)
     {
         $keys = [];
@@ -144,6 +151,17 @@ class AccessToken
         return $alg;
     }
 
+    /**
+     * Verifies an ES256-signed JWT.
+     *
+     * @param string $token The JSON Web Token to be verified.
+     * @param array $certs Certificate array according to the JWK spec (see
+     *                     https://tools.ietf.org/html/rfc7517).
+     * @param string|null $audience If set, returns false if the provided
+     *                              audience does not match the "aud" claim on
+     *                              the JWT.
+     * @return array|bool the token payload, if successful, or false if not.
+     */
     private function verifyEs256($token, $certs, $audience = null)
     {
         $this->checkSimpleJwt();
@@ -169,6 +187,17 @@ class AccessToken
         return $jwt->getClaims();
     }
 
+    /**
+     * Verifies an RS256-signed JWT.
+     *
+     * @param string $token The JSON Web Token to be verified.
+     * @param array $certs Certificate array according to the JWK spec (see
+     *                     https://tools.ietf.org/html/rfc7517).
+     * @param string|null $audience If set, returns false if the provided
+     *                              audience does not match the "aud" claim on
+     *                              the JWT.
+     * @return array|bool the token payload, if successful, or false if not.
+     */
     private function verifyRs256($token, $certs, $audience = null)
     {
         $this->checkAndInitializePhpsec();
@@ -193,6 +222,8 @@ class AccessToken
                     $cert['e']
                 ]), 256),
             ]);
+
+            // create an array of key IDs to certs for the JWT library
             $keys[$cert['kid']] =  $rsa->getPublicKey();
         }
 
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -18,7 +18,8 @@
     "friendsofphp/php-cs-fixer": "^1.11",
     "phpunit/phpunit": "^4.8.36|^5.7",
     "sebastian/comparator": ">=1.2.3",
-    "phpseclib/phpseclib": "^2"
+    "phpseclib/phpseclib": "^2",
+    "kelvinmo/simplejwt": "^0.2.5"
```
jdpedrie: Does `firebase/php-jwt` not have the utilities needed to support the new verification? Perhaps we could think about moving everything over? The firebase library leaves much to be desired.

src/AccessToken.php
```diff
@@ -103,57 +101,129 @@ public function verify($token, array $options = [])
...
             ? $options['certsLocation']
             : self::FEDERATED_SIGNON_CERT_URL;
-
-        unset($options['audience'], $options['certsLocation']);
+        $cacheKey = isset($options['cacheKey'])
+            ? $options['cacheKey']
+            : $this->getCacheKeyFromCertLocation($certsLocation);
 
         // Check signature against each available cert.
         // allow the loop to complete unless a known bad result is encountered.
```
jdpedrie: nit: this comment should move to where the loop happens now.

src/AccessToken.php
```diff
@@ -103,57 +101,129 @@ public function verify($token, array $options = [])
...
+            case 'RS256':
+                return $this->verifyRs256($token, $certs, $audience);
+
+            default:
+                throw new \InvalidArgumentException(
+                    'unrecognized "alg" in certs, expected ES256 or RS256');
+        }
+    }
+
+    private function determineAlg($certs)
```
jdpedrie: would you mind documenting the private methods?

src/AccessToken.php
```diff
@@ -266,20 +340,24 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
 
-        if (property_exists($class, 'leeway') && $class::$leeway < 1) {
-            // Ensures JWT leeway is at least 1
-            // @see https://github.com/google/google-api-php-client/issues/827
-            $class::$leeway = 1;
+    private function checkSimpleJwt()
+    {
+        // @codeCoverageIgnoreStart
+        if (!class_exists(SimpleJWT::class)) {
+            throw new \RuntimeException('Please require kelvinmo/simplejwtlib ^0.2 to use this utility.');
```
jdpedrie: Why not require this? (If it stays in require-dev only, we need to add it to suggests)

src/AccessToken.php
```diff
@@ -317,4 +395,30 @@ protected function callJwtStatic($method, array $args = [])
...
+     * exception of using "federated_signon_certs_v3" to preserve BC.
+     *
+     * @param string $certsLocation
+     * @return string
+     */
+    private function getCacheKeyFromCertLocation($certsLocation)
+    {
+        return $certsLocation === self::FEDERATED_SIGNON_CERT_URL
+            ? 'federated_signon_certs_v3'
+            : sha1($certsLocation);
```
jdpedrie: perhaps we can prefix this with a readable string to give a hint as to what it means for anyone who may be inspecting the cache? might be good for the default ones too.

```php
        $key = $certsLocation === self::FEDERATED_SIGNON_CERT_URL
            ? 'federated_signon_certs_v3'
            : sha1($certsLocation);

		return 'google_auth_certs_cache:' . $key;
```

src/AccessToken.php
```diff
@@ -266,20 +340,24 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
-     * Set required defaults for JWT.
-     */
-    private function configureJwtService()
+    private function checkAndInitializePhpsec()
     {
-        $class = class_exists('Firebase\JWT\JWT')
-            ? 'Firebase\JWT\JWT'
-            : '\JWT';
+        // @codeCoverageIgnoreStart
+        if (!class_exists(RSA::class)) {
```
jdpedrie: We didn't fail if phpseclib was uninstalled before. Was that incorrect/broken behavior?

composer.json
bshaffer: The firebase library does not support ES256. I submitted a PR which adds support, but it's pretty limited compared to the support afforded by these other libraries (https://github.com/firebase/php-jwt/pull/256)

The problem with switching over entirely is, in the case of `kelvinmo/simplejwt`, the additional extenion requirement  of `opeenssl` and `hash`, and also that it comes in a pre-1.0 version.

bshaffer: Looking into it, `hash` is compiled into PHP by default, so it may be possible to submit a PR to the maintainer to remove the `openssl` extension requirement

src/AccessToken.php
bshaffer: Because this would require the `openssl` PHP extension for all of our client libraries

src/AccessToken.php
bshaffer: Actually, there is no loop, because the Firebase JWT library supports an array of certs. So I'll remove the comment.

src/AccessToken.php
bshaffer: Done.

src/AccessToken.php
bshaffer: Great idea, and done!

src/AccessToken.php
bshaffer: It would have thrown a PHP Fatal error "class not found" instead of an exception. I think this behavior is better, but I can revert it to uphold the almighty Backwards Compatibility!

bshaffer:
@jdpedrie One thing we _could_ do is support PHPSec if it's installed, otherwise use and expect `SimpleJWT` as the only optional dependency and document it as well. This way users only worry about ONE optional dependency unless they're _already using_ PHPSec. This would preserve BC and make a better experience overall.

src/AccessToken.php
```diff
@@ -317,4 +425,32 @@ protected function callJwtStatic($method, array $args = [])
             : 'JWT';
         return call_user_func_array([$class, $method], $args);
     }
+
+    /**
+     * Provide a hook to mock calls to the JWT static methods.
+     *
+     * @param string $method
```
dwsupplee: This is unused

bshaffer: What do you mean? It's used in the function:

```php
    protected function callJwtStatic($method, array $args = [])
    {
        $class = class_exists('Firebase\JWT\JWT')
            ? 'Firebase\JWT\JWT'
            : 'JWT';
        return call_user_func_array([$class, $method], $args);
    }
```

dwsupplee: Well, I just noticed it isn't used in the signature of this protected method here

PR #256, opened 2019-12-31
feat: Add support for Guzzle 7.x

jeromegamez:
Greetings 👋

**Update 2020-06-27: [Guzzle 7.0 has been released](https://github.com/guzzle/guzzle/releases)**

Guzzle 7 is only in beta at the moment, but due to its popularity/once it's released I assume that many developers will want to use it, perhaps already today. This PR aims to make google/auth ready for that and is a prerequisite to making similar changes to [`google/cloud(-core)`](https://github.com/googleapis/google-cloud-php) and [`google/api-client`](https://github.com/googleapis/google-api-php-client).

### Changes

- Updated composer.json
- ~Extracted a version helper into a dedicated method in `HttpHandlerFactory::getGuzzleVersion()` and replaced usages~ _(Reverted after review)_ 
- ~Updated test matrix to test all supported Guzzle Versions (5.x is covered by `--prefer-lowest`, 6.x by the new environment, 7.x by default)~ _(The tests have moved to GitHub Actions since)_
- Updated the check itself: ~According to https://github.com/guzzle/guzzle/blob/7.0.0-beta.1/UPGRADING.md#other-backwards-compatibility-breaking-changes , the `ClientInterface::VERSION` constant has been removed - that means that we can safely assume that a missing constant means that we're on 7.x~ Guzzle 7 has introduced `ClientInterface::MAJOR_VERSION` that can be used for version checks.
- Changed the HTTP client option `exceptions` to `http_errors` for tests with Guzzle >=6.0 (support for `exceptions` has been removed in Guzzle 7)

:octocat:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
   "require": {
     "php": ">=5.4",
     "firebase/php-jwt": "~2.0|~3.0|~4.0|~5.0",
-    "guzzlehttp/guzzle": "~5.3.1|~6.0",
+    "guzzlehttp/guzzle": "^5.3.1|^6.0|^7.0",
     "guzzlehttp/psr7": "^1.2",
     "psr/http-message": "^1.0",
     "psr/cache": "^1.0"
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
   "require": {
     "php": ">=5.4",
     "firebase/php-jwt": "~2.0|~3.0|~4.0|~5.0",
-    "guzzlehttp/guzzle": "^5.3.1|^6.0|^7.0",
+    "guzzlehttp/guzzle": "^5.3.1|^6.2.1|^7.0",
     "guzzlehttp/psr7": "^1.2",
     "psr/http-message": "^1.0",
     "psr/cache": "^1.0"
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -20,6 +20,7 @@ namespace Google\Auth;
 use Google\Auth\Credentials\InsecureCredentials;
 use Google\Auth\Credentials\ServiceAccountCredentials;
 use Google\Auth\Credentials\UserRefreshCredentials;
+use GuzzleHttp\ClientInterface;
 
 /**
  * CredentialsLoader contains the behaviour used to locate and find default
@@ -52,6 +53,24 @@ abstract class CredentialsLoader implements FetchAuthTokenInterface
     private static function isOnWindows()
     {
         return strtoupper(substr(PHP_OS, 0, 3)) === 'WIN';
+    }
+
+    /**
+     * Returns the currently available major Guzzle version.
+     *
+     * @return int
+     */
+    private static function getGuzzleMajorVersion()
+    {
+        if (defined('GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
+            return ClientInterface::MAJOR_VERSION;
+        }
+
+        if (defined('GuzzleHttp\ClientInterface::VERSION')) {
+            return (int) substr(ClientInterface::VERSION, 0, 1);
+        }
+
+        throw new \Exception('Version not supported');
     }
 
     /**
@@ -145,35 +164,30 @@ abstract class CredentialsLoader implements FetchAuthTokenInterface
         callable $httpHandler = null,
         callable $tokenCallback = null
     ) {
-        $version = \GuzzleHttp\ClientInterface::VERSION;
-
-        switch ($version[0]) {
-            case '5':
-                $client = new \GuzzleHttp\Client($httpClientOptions);
-                $client->setDefaultOption('auth', 'google_auth');
-                $subscriber = new Subscriber\AuthTokenSubscriber(
-                    $fetcher,
-                    $httpHandler,
-                    $tokenCallback
-                );
-                $client->getEmitter()->attach($subscriber);
-                return $client;
-            case '6':
-                $middleware = new Middleware\AuthTokenMiddleware(
-                    $fetcher,
-                    $httpHandler,
-                    $tokenCallback
-                );
-                $stack = \GuzzleHttp\HandlerStack::create();
-                $stack->push($middleware);
-
-                return new \GuzzleHttp\Client([
-                   'handler' => $stack,
-                   'auth' => 'google_auth',
-                ] + $httpClientOptions);
-            default:
-                throw new \Exception('Version not supported');
+        if (self::getGuzzleMajorVersion() === 5) {
+            $client = new \GuzzleHttp\Client($httpClientOptions);
+            $client->setDefaultOption('auth', 'google_auth');
+            $subscriber = new Subscriber\AuthTokenSubscriber(
+                $fetcher,
+                $httpHandler,
+                $tokenCallback
+            );
+            $client->getEmitter()->attach($subscriber);
+            return $client;
         }
+
+        $middleware = new Middleware\AuthTokenMiddleware(
+            $fetcher,
+            $httpHandler,
+            $tokenCallback
+        );
+        $stack = \GuzzleHttp\HandlerStack::create();
+        $stack->push($middleware);
+
+        return new \GuzzleHttp\Client([
+            'handler' => $stack,
+            'auth' => 'google_auth',
+        ] + $httpClientOptions);
     }
 
     /**
--- a/src/HttpHandler/Guzzle7HttpHandler.php
+++ b/src/HttpHandler/Guzzle7HttpHandler.php
@@ -1,12 +1,12 @@
 <?php
-/*
- * Copyright 2010 Google Inc.
+/**
+ * Copyright 2020 Google LLC
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *      http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -14,26 +14,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+namespace Google\Auth\HttpHandler;
 
-namespace Google\Auth\Tests\Middleware;
-
-use Google\Auth\Tests\BaseTest;
-
-class SimpleMiddlewareTest extends BaseTest
+class Guzzle7HttpHandler extends Guzzle6HttpHandler
 {
-    private $mockRequest;
-
-    /**
-     * @todo finish
-     */
-    protected function setUp()
-    {
-        $this->onlyGuzzle6And7();
-
-        $this->mockRequest = $this->prophesize('GuzzleHttp\Psr7\Request');
-    }
-
-    public function testTest()
-    {
-    }
 }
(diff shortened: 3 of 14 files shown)


tests/BaseTest.php
```diff
@@ -2,23 +2,21 @@
 
 namespace Google\Auth\Tests;
 
-use GuzzleHttp\ClientInterface;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
 use PHPUnit\Framework\TestCase;
 
 abstract class BaseTest extends TestCase
 {
     public function onlyGuzzle6()
```
bshaffer: This method needs to be renamed to `atLeastGuzzle6` or `onlyGuzzle6Plus`

jeromegamez: You're of course right! I didn't change the method at first to make the footprint of my changes as small as possible, but it totally makes sense - done!

jeromegamez:
The `Guzzle6HttpHandler` has a similar naming issue as the test helper [mentioned here](https://github.com/googleapis/google-auth-library-php/pull/256#discussion_r364475095), but this is not as easily resolvable I think.

Do you think we can/should keep it as it is, or search for an approach that wouldn't disrupt devs using it directly? My first thought was to create a `Guzzle7Httphandler` and either extend it (with an empty body) or extract a trait and use it in both classes, but I'm not sure - I personally would like the unambiguity, but since the behavior is the same, it wouldn't add a "real" benefit beside this 🤔

bshaffer:
@jeromegamez I like your idea of extending the `Guzzle6HttpHandler` with an empty body. This doesn't create a huge "footprint" and provides unambiguity.

jeromegamez:
@bshaffer Done! As I didn't think it to be a good idea to duplicate the tests for the new handler, I simplified the `Guzzle6HttpHandlerTest` and extended the `Guzzle7HttpHandlerTest` to perform the same tests - it's more changed code, but I hope it's worth it 😅

.travis.yml
```diff
@@ -45,11 +45,15 @@ matrix:
     - name: "PHP 7.4"
       php: "7.4"
 
+    - name: "PHP 7.4 (Use Guzzle 6)"
+      php: "7.4"
+      before_script: composer require "guzzlehttp/guzzle:^6.0"
```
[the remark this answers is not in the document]
jeromegamez: I haven't tested it extensively, but I believe if we use `--prefer-lowest` then a 5.x release will be pulled in - this is our lowest dependency, and Guzzle's 5.x branch has `"php": ">=5.4"` in its `composer.json` so that we couldn't even trick it by using PHP 7.4 (or 8.0 😅)

composer.json
```diff
@@ -8,7 +8,7 @@
   "require": {
     "php": ">=5.4",
     "firebase/php-jwt": "~2.0|~3.0|~4.0|~5.0",
-    "guzzlehttp/guzzle": "~5.3.1|~6.0",
+    "guzzlehttp/guzzle": "^5.3.1|^6.0|^7.0@beta",
```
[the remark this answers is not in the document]
jeromegamez: I was afraid you would say that, but I totally understand. I'll keep an eye then on Guzzle's releases and rebase and ping you once it's considered stable 🤞

src/HttpHandler/HttpHandlerFactory.php
```diff
@@ -30,16 +30,29 @@ class HttpHandlerFactory
...
                 throw new \Exception('Version not supported');
         }
     }
+
+    /**
+     * Returns the installed Guzzle version.
+     *
+     * @return string
+     */
+    public static function getGuzzleVersion()
```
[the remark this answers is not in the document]
jeromegamez: The `CredentialsLoader::makeHttpClient()` method uses this method as well 🙈

src/CredentialsLoader.php
```diff
@@ -160,6 +159,7 @@ public static function makeHttpClient(
                 $client->getEmitter()->attach($subscriber);
                 return $client;
             case '6':
+            case '7':
```
bshaffer: I would prefer this:

```php
if (
    defined('GuzzleHttp\ClientInterface::VERSION')
    && \GuzzleHttp\ClientInterface::VERSION[0] == '5'
) {
    // Use Subscriber
} else {
   // use Middleware 
}
```

src/HttpHandler/HttpHandlerFactory.php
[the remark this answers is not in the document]
bshaffer: See above comment, I'd rather be slightly less DRY than expose this method.

bshaffer:
After [chatting over on the guzzle repo](https://github.com/guzzle/guzzle/issues/2579#issuecomment-581641392), instead of a Guzzle7 handler we could do an `HttpMessageHandler` which handles anything from PSR 18. And Guzzle7Handler could extend this for BC.

jeromegamez:
Absolutely, I agree! The reason why I didn't pursue this in the first place is that depending on a PSR-18 client and leaving choosing a client to the consumer would be a BC break as we won't be able to remove the Guzzle dependency in the 1.x releases.

So my approach was to achieve Guzzle 7 compatibility with the least impact possible (thinking: the less code I change the likelier it is that it will get merged :).

If you'd like me to pursue this instead of the current approach, I will close the PR and create a new one that requires guzzle *and* `psr/http-client` as an intermediary step where we could deprecate the dependency on Guzzle 5/6.

It will be difficult though - everywhere the HTTP handler is currently used, it is expected to be a `callable($request, $options = [])` where the options are Guzzle specific and can be set by the consumer, which could potentially lead to breaks.

bshaffer:
It's definitely an issue that PSR-18 doesn't support request options. This is something we are not sure what do with moving forward. I am ok keeping the current implementation for now.

jeromegamez:
[Guzzle 7.0 has been released](https://github.com/guzzle/guzzle/releases) 🥳  - I updated the version constraint in the `composer.json` accordingly 🤞

src/HttpHandler/HttpHandlerFactory.php
```diff
@@ -30,16 +30,16 @@ class HttpHandlerFactory
      */
     public static function build(ClientInterface $client = null)
     {
-        $version = ClientInterface::VERSION;
         $client = $client ?: new Client();
 
-        switch ($version[0]) {
-            case '5':
+        if (defined('GuzzleHttp\ClientInterface::VERSION')) {
```
Nyholm, line 35: We can detect versions like: 

```php
    $guzzleVersion = null;
    if (defined('\GuzzleHttp\ClientInterface::MAJOR_VERSION')) {
      $guzzleVersion = ClientInterface::MAJOR_VERSION;
    } elseif (defined('\GuzzleHttp\ClientInterface::VERSION')) {
      $guzzleVersion = (int) substr(ClientInterface::VERSION, 0, 1);
    }

```

tests/BaseTest.php
```diff
@@ -7,22 +7,47 @@
...
+
+    protected function onlyGuzzle6()
+    {
+        if ($this->getGuzzleVersion() !== 6) {
             $this->markTestSkipped('Guzzle 6 only');
         }
     }
 
-    public function onlyGuzzle5()
+    protected function atLeastGuzzle6()
```
Nyholm, line 24: hm... Im not sure about naming. This will need to change when Guzzle 8 comes out. 

How about `onlyGuzzle6and7`? Or `onlyGuzzle6` and `onlyGuzzle7`?
Or maybe even `requireGuzzleVersion(... $versions)`?

tests/FetchAuthTokenTest.php
```diff
@@ -61,14 +61,21 @@ class_implements($fetcherClass)
             $this->assertEquals('xyz', $accessToken);
         };
 
-        $client = CredentialsLoader::makeHttpClient(
-            $mockFetcher->reveal(),
-            [
+        if (strpos($this->getGuzzleVersion(), '5') === 0) {
```
Nyholm, line 64:
```suggestion
        if ($this->getGuzzleVersion() === 5) {
```

tests/BaseTest.php
bshaffer: I am okay with `onlyGuzzle6and7`, might as well be explicit

src/CredentialsLoader.php
```diff
@@ -145,35 +145,30 @@ public static function makeHttpClient(
...
-                $stack = \GuzzleHttp\HandlerStack::create();
-                $stack->push($middleware);
-
-                return new \GuzzleHttp\Client([
-                   'handler' => $stack,
-                   'auth' => 'google_auth',
-                ] + $httpClientOptions);
-            default:
-                throw new \Exception('Version not supported');
+        if (defined('GuzzleHttp\ClientInterface::VERSION') && strpos(\GuzzleHttp\ClientInterface::VERSION, '5') === 0) {
```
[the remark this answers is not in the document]
jeromegamez: Just to be sure - we had a discussion in February where I had such a method and you preferred the code to be a little less dry, that's why I kept the version check here as  it was 😅.

I'll add the method right away (and make it static), I just wanted to mention it.

jeromegamez:
I think we're ready 😅 🤞

src/CredentialsLoader.php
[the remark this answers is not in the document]
bshaffer: My concern then (and still today) was in making that method public.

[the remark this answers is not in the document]
jeromegamez: Got it! Thanks for the clarification!

src/HttpHandler/Guzzle7HttpHandler.php
```diff
@@ -0,0 +1,7 @@
+<?php
```
jdpedrie, line 1: This file needs a copyright notice.

jeromegamez: Note to self: That's what happens if you just copy and paste 😅.

I added one to the Guzzle6Handler as well (with 2015, since it was introduced in that year)

jeromegamez:
Thanks @jdpedrie - I added the section you suggested, but had to remove the `--prefer-lowest` as it installed Guzzle 6.0 which [broke the tests](https://github.com/googleapis/google-auth-library-php/actions/runs/153670557).

I hope that's okay though - an alternative could be to raise the Guzzle 6 minimum version to `^6.2.1` - it's generally recommended to use at least that version anyway to address the HTTP_PROXY security vulnerability.

Nyholm:
Yes. Please do.

jdpedrie:
Thanks @jeromegamez,

I didn't mean to keep the `--prefer-lowest` flag. We can remove that from the added test matrix. I don't see a reason why we can't bump the 6.x requirement to `^6.2.1` as you suggest, so let's go ahead and do both!

src/HttpHandler/Guzzle7HttpHandler.php
jdpedrie: Thanks! And thank you for catching the missing copyright in the other file! I never noticed. 😆

src/HttpHandler/Guzzle7HttpHandler.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Copyright 2020 Google Inc. All Rights Reserved.
```
jdpedrie, line 3: One more nit: The new copyright should be

```suggestion
 * Copyright 2020 Google LLC
```

I'm sorry for not mentioning it before, since it's totally non-obvious!

Thanks again for all the help here, we all really appreciate it!

jeromegamez: I also could have checked in other files myself 🙈. Done! (And I enjoy the interaction as well!)

Nyholm:
Wohoo!

jeromegamez:
Yessssss!!! 🥳

PR #265, opened 2020-02-12
feat: add option to throw exception in AccessToken::verify

bshaffer:
Our wrapper in `AccessToken` for verifying JWTs is made significantly less valuable because we are suppressing the exceptions thrown by the underlying libraries which tell us what went wrong.  We need to expose a way for the user to see what error took place. I see the following options as viable:

### Option 1
Add a method `getVerifyException` which returns the exception which was thrown:

```php
$token = new AccessToken();
if (!$result = $token->verify($jwt)) {
    print($token->getVerifyException()->getMessage());
}
```

### Option 2
Add an optional third argument which will contain the error message when the call fails:

```php
$token = new AccessToken();
$result = $token->verify($token, [], $message);

if (!$result) {
    print($message);
}
```

### Option 3
Add an option to throw the error/exception on failure:

```php
$token = new AccessToken();
try {
    $result = $token->verify($token, [
        'throwException' => true
    ]);
} catch (Exception $e) {
    print($e->getMessage());
}
```

### Option 4
Add an option to return the error/exception on failure:

```php
$token = new AccessToken();
$result = $token->verify($token, [
    'returnException' => true
]);

if ($result instanceof Exception) {
    print($result->getMessage());
}
```

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -32,6 +32,7 @@ use SimpleJWT\Keys\KeyFactory;
 use SimpleJWT\Keys\KeySet;
 use SimpleJWT\InvalidTokenException;
 use Psr\Cache\CacheItemPoolInterface;
+use Exception;
 
 /**
  * Wrapper around Google Access Tokens which provides convenience functions.
@@ -55,6 +56,11 @@ class AccessToken
      * @var CacheItemPoolInterface
      */
     private $cache;
+
+    /**
+     * @var Exception
+     */
+    private $verifyException;
 
     /**
      * @param callable $httpHandler [optional] An HTTP Handler to deliver PSR-7 requests.
@@ -174,6 +180,7 @@ class AccessToken
         try {
             $jwt = $this->callSimpleJwtDecode([$token, $jwkset, 'ES256']);
         } catch (InvalidTokenException $e) {
+            $this->verifyException = $e;
             return false;
         }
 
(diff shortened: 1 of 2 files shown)


bshaffer:
Another potential way to solve this issue would be to change the signature of `verify`:

```php
public function verify($token, array $options = [], &$error = null)
{
    //...
    try {
        // ...
    } catch (InvalidTokenException $e) {
        $message = $e->getMessage();
    }
}
```

jdpedrie:
What about adding `$options.returnErrorOnFailure`? If set to true and token is invalid, the return value is an Exception. The various spots where we currently return `false` without an exception will need to return something more useful for debugging (i.e. if the token issuer doesn't match).

bshaffer:
@jdpedrie I'm not crazy about having a variable return type, especially with the multitude of exception classes that are thrown in those methods. I would be more in favor of an `$options.throwExceptionOnFailure` option.

bshaffer:
@jdpedrie @dwsupplee Can we get consensus on this? I would be happy with any of them, but I'd rank them as follows, with Option 3 being my favorite:

 - Option 3
 - Option 2
 - Option 4
 - Option 1

jdpedrie:
I'd mostly agree, except swapping 2 and 4. Option 3 seems the best by a decent margin.

Would we want to add exceptions to cases which simply return false currently? e.g. [here](https://github.com/jdpedrie/google-auth-library-php/blob/39e243a7d8320b1889ab8bb4cd6f98f7af83c582/src/AccessToken.php#L238) and [here](https://github.com/jdpedrie/google-auth-library-php/blob/39e243a7d8320b1889ab8bb4cd6f98f7af83c582/src/AccessToken.php#L246)

bshaffer:
@jdpedrie that seems like the way to go, I'll add that as well!

src/AccessToken.php
```diff
@@ -167,8 +168,12 @@ private function determineAlg(array $certs)
      *                              the JWT.
      * @return array|bool the token payload, if successful, or false if not.
      */
-    private function verifyEs256($token, array $certs, $audience = null)
-    {
+    private function verifyEs256(
+        $token,
+        array $certs,
+        $audience = null,
+        $throwException = false
```
jdpedrie: Wouldn't it be easier to just allow the private methods to throw exceptions instead of `catch { if ($throwException) {`, bubble up to verify() and be suppressed or thrown there?

bshaffer: This will look way cleaner, good call

bshaffer:
@jdpedrie @dwsupplee PTA(nother)L!

PR #267, opened 2020-03-11
feat: add option to specify issuer in AccessToken::verify

stephenmcd:
This allows issuer to be passed to AccessToken::verify, providing support for any JWTs

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -215,7 +215,7 @@ class AccessToken
         }
 
         // @see https://cloud.google.com/iap/docs/signed-headers-howto#verifying_the_jwt_payload
-        $issuer = isset($issuer) ? $issuer : self::IAP_ISSUER;
+        $issuer = $issuer ?: self::IAP_ISSUER;
         if (!isset($payload['iss']) || $payload['iss'] !== $issuer) {
             throw new UnexpectedValueException('Issuer does not match');
         }
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -280,7 +280,7 @@ class AccessToken
 
         // support HTTP and HTTPS issuers
         // @see https://developers.google.com/identity/sign-in/web/backend-auth
-        $issuers = isset($issuer) ? array($issuer) : [self::OAUTH2_ISSUER, self::OAUTH2_ISSUER_HTTPS];
+        $issuers = $issuer ? [$issuer] : [self::OAUTH2_ISSUER, self::OAUTH2_ISSUER_HTTPS];
         if (!isset($payload->iss) || !in_array($payload->iss, $issuers)) {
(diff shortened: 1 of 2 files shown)


src/AccessToken.php
```diff
@@ -129,7 +133,7 @@ public function verify($token, array $options = [])
         }
         try {
             if ($alg == 'RS256') {
-                return $this->verifyRs256($token, $certs, $audience);
+                return $this->verifyRs256($token, $certs, $audience, $issuer);
             }
             return $this->verifyEs256($token, $certs, $audience);
```
bshaffer: You need to pass it here also!

stephenmcd: Done in 94a110effd2a6ce370007169e17e7002febad965

src/AccessToken.php
```diff
@@ -208,7 +212,8 @@ private function verifyEs256($token, array $certs, $audience = null)
         }
 
         // @see https://cloud.google.com/iap/docs/signed-headers-howto#verifying_the_jwt_payload
-        if (!isset($payload['iss']) || $payload['iss'] !== self::IAP_ISSUER) {
+        $issuer = isset($issuer) ? $issuer : self::IAP_ISSUER;
```
jdpedrie: nit: this can be contracted:

```suggestion
        $issuer = $issuer ?: self::IAP_ISSUER;
```

src/AccessToken.php
```diff
@@ -269,7 +274,7 @@ private function verifyRs256($token, array $certs, $audience = null)
 
         // support HTTP and HTTPS issuers
         // @see https://developers.google.com/identity/sign-in/web/backend-auth
-        $issuers = [self::OAUTH2_ISSUER, self::OAUTH2_ISSUER_HTTPS];
+        $issuers = isset($issuer) ? array($issuer) : [self::OAUTH2_ISSUER, self::OAUTH2_ISSUER_HTTPS];
```
jdpedrie: Can't do the shorthand ternary here, but could we drop the `isset()` and use shorthand arrays?

```suggestion
        $issuers = $issuer ? [$issuer] : [self::OAUTH2_ISSUER, self::OAUTH2_ISSUER_HTTPS];
```

src/AccessToken.php
```diff
@@ -226,7 +231,7 @@ private function verifyEs256($token, array $certs, $audience = null)
      *                              the JWT.
      * @return array|bool the token payload, if successful, or false if not.
      */
-    private function verifyRs256($token, array $certs, $audience = null)
+    private function verifyRs256($token, array $certs, $audience = null, $issuer = null)
```
dwsupplee: It would be nice to add this to the docblock, otherwise LGTM! :)

stephenmcd: Done

PR #268, opened 2020-03-24
Update PHPDocs

jeromegamez:
Greetings 👋,

in one of my projects, PHPStan complained about a new condition 

```php
if (GCECredentials::onGce()) {

}
```

being always true. I then used PHPStorm's Code Inspection to find other PHPDoc recommendations and tried to make the changes as "non-invasive" as possible 🙈

:octocat:

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -325,6 +325,7 @@ class AccessToken
      * are PEM encoded certificates.
      *
      * @param string $location The location from which to retrieve certs.
+     * @param string $cacheKey The key under which to cache the retrieved certs.
      * @param array $options [optional] Configuration options.
      * @return array
      * @throws InvalidArgumentException If received certs are in an invalid format.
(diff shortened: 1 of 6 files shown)


src/ApplicationDefaultCredentials.php
```diff
@@ -136,7 +136,7 @@ public static function getMiddleware(
      * @param array $cacheConfig configuration for the cache when it's present
      * @param CacheItemPoolInterface $cache
      *
-     * @return CredentialsLoader
+     * @return CredentialsLoader|null
```
jdpedrie: So far as I can determine, this always returns a `CredentialsLoader` or throws an exception.

jeromegamez: You're right, good catch! I was blinded by `$creds` being initialized with  `null` and returned at the end, without checking in between 🙈. I reverted the change, thank you!

PR #270, opened 2020-03-25
Add support signing key id for jwt service account authenticate

arturgspb:
Hi! This PR add optional case for send param "kid" in JWT header for service account jwt auth


arturgspb:
Failed build on php 5.4 but not on my code - composer download error

arturgspb:
Hi! @dwsupplee , please can you review PR?

arturgspb:
Hey

arturgspb:
@dwsupplee , done.

But I think travis use old JWT version < 3.0.0 on build and exceptions on test is various:

https://github.com/firebase/php-jwt/blob/v2.2.0/Authentication/JWT.php#L80

https://github.com/firebase/php-jwt/blob/v3.0.0/src/JWT.php#L91

https://github.com/firebase/php-jwt/blob/v4.0.0/src/JWT.php#L103

Also in old version JWT has Undefined index problem: https://travis-ci.org/github/googleapis/google-auth-library-php/jobs/683565112

arturgspb:
@bshaffer ping. Review pls

bshaffer:
@arturgspb this looks good for me, but what is your use case? Can you provide an example of when you're using this and why?

arturgspb:
@bshaffer I coding lib to work with API Garpun ( https://en.garpun.com/ ), auth can be service accounts issued by Garpun.

I think it will be good, if my clients use fully tested library for auth with service accounts .
I chose libraries by Google: Java, Python and Php. Java and Python works fine, in PHP no support parameter **signingKeyId** now. Apparently this is not necessary for Google, but I would like it to be possible to use it. Because JWT lib it supports.

I don’t think I should change the behavior of _new \Google\Auth\Credentials\ServiceAccountJwtAccessCredentials($jsonKey)_ so I decided that I just need to give my clients an example authorization code with parameter substitution in the main class.

```php
<?php
require_once __DIR__ . '/vendor/autoload.php';

use Google\Auth\Middleware\AuthTokenMiddleware;
use GuzzleHttp\Client;
use GuzzleHttp\HandlerStack;

$credFileName = "/Users/arturgspb/IdeaProjects/garpun-api-java-lib/src/main/resources/creds.json";
$jsonKey = json_decode(file_get_contents($credFileName), true);


$creds = new \Google\Auth\OAuth2([
    'audience' => $jsonKey['token_uri'],
    'issuer' => $jsonKey['client_email'],
    'scope' => ['account-management'],
    'signingAlgorithm' => 'RS256',
    'signingKey' => $jsonKey['private_key'],
    'signingKeyId' => $jsonKey['private_key_id'],
    'sub' => null,
    'tokenCredentialUri' => $jsonKey['token_uri'],
]);

$middleware = new AuthTokenMiddleware($creds);
$stack = HandlerStack::create();
$stack->push($middleware);

$client = new Client([
    'handler' => $stack,
    'base_uri' => 'https://accountmanagement-api.garpun.com',
    'auth' => 'google_auth'
]);

$response = $client->post('v1beta2/report/client', [
    'json' => [
        "modification_time" => "2020-01-09T18:31:42",
        "paging" => [
            "offset" => 0,
            "limit" => 10
        ]
    ]
]);

print_r((string)$response->getBody());
```

bshaffer:
Thank you for the thorough reply @arturgspb. Thank you for the confirmation that this feature isn't being used inside the Google Auth ecosystem. I am glad you've found this library useful for other applications as well!

PR #273, opened 2020-03-30
chore: switch from Travis to Github Actions

jdpedrie:
This change removes Travis from the test matrix and replaces it with Github Actions. It also replaces the code style checker with a similar one to what Google Cloud PHP uses. It includes fixes to conform to that style checker.

It generates and pushes documentation to github pages on push to master and branch.

PTAL, LMK, IDK.

Sample run: https://github.com/jdpedrie/google-auth-library-php/actions/runs/67515476
Generated documentation: http://jdpedrie.github.io/google-auth-library-php/master/


.github/workflows/tests.yml
```diff
@@ -0,0 +1,108 @@
+name: Test Suite
+on: [push, pull_request]
```
bcoe, line 2: I suggest listening on these events:

```
on:
  push:
    branches:
      - master
  pull_request:
```

This way you don't kick off two sets of tests for a pull_request.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,108 @@
...
+jobs:
+    test:
+        runs-on: ${{matrix.operating-system}}
+        strategy:
+            matrix:
+                operating-system: [ ubuntu-latest ]
+                php: [ "5.6", "7.0", "7.1", "7.2", "7.3", "7.4" ]
+        name: PHP ${{matrix.php }} Unit Test
+        steps:
+            - uses: actions/checkout@master
```
bcoe, line 13: I might be tempted to use this at a specific tag, I think we use `v2` for our Node.js repos (just so you don't get any surprise regressions).

src/AccessToken.php
```diff
@@ -82,22 +82,19 @@ public function __construct(
...
+     *        "federated_signon_certs_v3".
+     * @param string $options.certsLocation The location (remote or local) from which
      *        to retrieve certificates, if not cached. This value should only be
      *        provided in limited circumstances in which you are sure of the
      *        behavior.
-     *     @type string $cacheKey The cache key of the cached certs. Defaults to
-     *        the sha1 of $certsLocation if provided, otherwise is set to
-     *        "federated_signon_certs_v3".
-     *     @type bool $throwException Whether the function should throw an
+     * @param bool $options.throwException Whether the function should throw an
```
bshaffer, line 95: This is different than the style we use for `google/cloud`. Are we moving to this format over there as well?

src/OAuth2.php
```diff
@@ -766,7 +770,7 @@ public function setRedirectUri($uri)
      */
     public function getScope()
     {
-        if (is_null($this->scope)) {
+        if (is_null($this->scope) || is_string($this->scope)) {
```
bshaffer, line 773: The typehint is wrong, the `$scope` property should always be an array. What prompted this change?

src/AccessToken.php
jdpedrie: The documentation generator used doesn't support the style we use on google cloud php. I wouldn't be opposed to moving to this style there as well, but it's a bigger job and would require changes to the generated code as well.

src/OAuth2.php
jdpedrie: fixed (and fixed the property type)

PR #276, opened 2020-04-15
fix: Excludes broken version of guzzle

bshaffer:


bshaffer:
Context: https://github.com/guzzle/guzzle/issues/2511#issuecomment-613649077

bshaffer:
Guzzle released a fix so this PR is no longer necessary

https://github.com/guzzle/guzzle/releases/tag/6.5.3

PR #277, opened 2020-05-08
feat: add quotaProject param for extensible client options support

bshaffer:
Request for Comment - Yet another parameter seems like the only sane way to do this.

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -192,6 +192,7 @@ class GCECredentials extends CredentialsLoader implements
         }
 
         $this->tokenUri = $tokenUri;
+        $this->quotaProject = $quotaProject;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Credentials/GCECredentials.php
```diff
@@ -149,13 +151,20 @@ class GCECredentials extends CredentialsLoader implements
      */
     private $targetAudience;
 
+    /**
+     * @var string
```
jdpedrie, line 155:
```suggestion
     * @var string|null
```

src/Credentials/GCECredentials.php
```diff
@@ -149,13 +151,20 @@ class GCECredentials extends CredentialsLoader implements
...
+     */
+    private $quotaProject;
+
     /**
      * @param Iam $iam [optional] An IAM instance.
      * @param string|array $scope [optional] the scope of the access request,
      *        expressed either as an array or as a space-delimited string.
      * @param string $targetAudience [optional] The audience for the ID token.
+     * @param string $quotaProject specifies a project to bill for access
```
jdpedrie, line 164:
```suggestion
     * @param string $quotaProject [optional] Specifies a project to bill for access
```

PR #282, opened 2020-05-18
chore: switch back to nanasess/setup-php in GH actions

bshaffer:


.github/workflows/tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
         steps:
             - uses: actions/checkout@v2
             - name: Setup PHP
-              uses: jdpedrie/setup-php@master
+              uses: nanasess/setup-php@master
```
jdpedrie, line 19:
```suggestion
              uses: nanasess/setup-php@v3
```

.github/workflows/tests.yml
```diff
@@ -37,7 +37,7 @@ jobs:
...
```
jdpedrie, line 40:
```suggestion
              uses: nanasess/setup-php@v3
```

.github/workflows/tests.yml
```diff
@@ -99,7 +99,7 @@ jobs:
...
```
jdpedrie, line 102:
```suggestion
              uses: nanasess/setup-php@v3
```

Issue #286, opened 2020-07-01
FetchAuthTokenCache does not work for id_token


jsoung:
There are two problems:

1. In [FetchAuthTokenCache::fetchAuthToken()](https://github.com/googleapis/google-auth-library-php/blob/master/src/FetchAuthTokenCache.php#L77-L98), the code checks whether the token is an `access_token`, if not, it bails.
2. In [OAuth2::getCacheKey()](https://github.com/googleapis/google-auth-library-php/blob/master/src/OAuth2.php#L540-L548), it returns ` $this->scope` or empty string. This will only work for `access_token`. The function should return `$this->audience` if it's available, which will work for `id_token`.

bshaffer:
This has been added in #291 and released in [`v1.11.0`](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.11.0)

PR #289, opened 2020-07-14
Refactor CacheTrait to enable expiration tracking

ericnorris:
This PR makes changes to the `CacheTrait` trait in order to add the ability to enable per-item expiration dates, and updates the `FetchAuthTokenCache` class in order to specify the auth token expiry when caching.

At the time of this pull request, there is a > 4 year old TODO in the `FetchAuthTokenCache` class:

https://github.com/googleapis/google-auth-library-php/blob/077d6ae98d550161d3b2a0ba283bdce785c74d85/src/FetchAuthTokenCache.php#L81-L84

Unfortunately this means that code using this library is subject to a potential error condition in which the auth token is cached for longer than it is valid. While the library does set a maximum lifetime of 1500 seconds (25 minutes) for tokens, it is possible for a Google metadata server to return an access token that expires within that lifetime; any API calls made using this token during that time frame will fail with a `403 Permission Denied` error.

From experience this bug has lead to more than one brief outage of a service that utilizes the cached tokens, and the goal of this PR is to eliminate the issue.

Note to maintainers:

* It may be helpful to go through the PR one commit at a time.
* I am under the impression that you strive to maintain a compatible public API, I consider the `CacheTrait` internal to the library and felt comfortable making breaking changes here, let me know if this is not the case.
* I'm fairly certain tests in `tests/Subscriber` will fail, I did not include updates to them as I wanted to get feedback on the general idea of the PR first.
* I've removed a decent number tests that I consider redudant - the `AuthTokenMiddlewareTest`, for example, includes tests that depend on the cache although the `AuthTokenMiddleware` class itself doesn't care whether or not it is using a caching `FetchAuthTokenInterface`. If there are any you disagree with, I'm happy to discuss and re-include them.

Thanks!

### DIFF
--- a/src/CacheTrait.php
+++ b/src/CacheTrait.php
@@ -17,49 +17,62 @@
 
 namespace Google\Auth;
 
+use Google\Auth\Cache\MemoryCacheItemPool;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
 trait CacheTrait
 {
+
     private $maxKeyLength = 64;
 
-    /**
-     * Gets the cached value if it is present in the cache when that is
-     * available.
-     */
-    private function getCachedValue($k)
-    {
-        if (is_null($this->cache)) {
-            return;
-        }
+    /** @var ?CacheItemPoolInterface */
+    private $cache;
 
-        $key = $this->getFullCacheKey($k);
-        if (is_null($key)) {
-            return;
-        }
+    /** @var array */
+    private $cacheConfig;
 
-        $cacheItem = $this->cache->getItem($key);
-        if ($cacheItem->isHit()) {
-            return $cacheItem->get();
-        }
+    private function initCacheTrait(
+        ?CacheItemPoolInterface $cache,
+        array $cacheConfig = []
+    ) {
+        $this->cache = $cache ?? new MemoryCacheItemPool;
+
+        $this->cacheConfig = $cacheConfig + [
+            'lifetime' => 1500,
+            'prefix'   => '',
+        ];
     }
 
     /**
-     * Saves the value in the cache when that is available.
+     * Looks up the key in the cache and, on cache miss, calls the $fetcher
+     * callback in order to fetch the real value.
+     *
+     * @param string $key the cache key to lookup
+     * @param callable $fetcher a callback of the form
+     *        function(CacheItemInterface $item): mixed, to be called on cache
+     *        miss
+     * @return mixed the cached value
      */
-    private function setCachedValue($k, $v)
+    private function getCachedValue(string $key, callable $fetcher)
     {
-        if (is_null($this->cache)) {
-            return;
+        $normalizedKey = $this->getFullCacheKey($key);
+
+        $cacheItem = $this->cache->getItem($normalizedKey);
+
+        if ($cacheItem->isHit())
+        {
+            return $cacheItem->get();
         }
 
-        $key = $this->getFullCacheKey($k);
-        if (is_null($key)) {
-            return;
-        }
-
-        $cacheItem = $this->cache->getItem($key);
-        $cacheItem->set($v);
+        // set the default expiry, the callable can override this
         $cacheItem->expiresAfter($this->cacheConfig['lifetime']);
-        return $this->cache->save($cacheItem);
+
+        $result = $fetcher($cacheItem);
+
+        $this->cache->save($cacheItem);
+
+        return $result;
     }
 
     private function getFullCacheKey($key)
--- a/src/FetchAuthTokenCache.php
+++ b/src/FetchAuthTokenCache.php
@@ -17,6 +17,7 @@
 
 namespace Google\Auth;
 
+use Psr\Cache\CacheItemInterface;
 use Psr\Cache\CacheItemPoolInterface;
 
 /**
@@ -57,11 +58,7 @@ class FetchAuthTokenCache implements
         CacheItemPoolInterface $cache
     ) {
         $this->fetcher = $fetcher;
-        $this->cache = $cache;
-        $this->cacheConfig = array_merge([
-            'lifetime' => 1500,
-            'prefix' => '',
-        ], (array) $cacheConfig);
+        $this->initCacheTrait($cache, $cacheConfig ?? []);
     }
 
     /**
@@ -76,25 +73,17 @@ class FetchAuthTokenCache implements
      */
     public function fetchAuthToken(callable $httpHandler = null)
     {
-        // Use the cached value if its available.
-        //
-        // TODO: correct caching; update the call to setCachedValue to set the expiry
-        // to the value returned with the auth token.
-        //
-        // TODO: correct caching; enable the cache to be cleared.
-        $cacheKey = $this->fetcher->getCacheKey();
-        $cached = $this->getCachedValue($cacheKey);
-        if (!empty($cached)) {
-            return ['access_token' => $cached];
-        }
+        return $this->getCachedValue(
+            $this->fetcher->getCacheKey(),
+            function(CacheItemInterface $item) use ($httpHandler) {
+                $authToken = $this->fetcher->fetchAuthToken($httpHandler);
 
-        $auth_token = $this->fetcher->fetchAuthToken($httpHandler);
+                $item->set($authToken);
+                $item->expiresAfter($authToken['expires_in']);
 
-        if (isset($auth_token['access_token'])) {
-            $this->setCachedValue($cacheKey, $auth_token['access_token']);
-        }
-
-        return $auth_token;
+                return $authToken;
+            }
+        );
     }
 
     /**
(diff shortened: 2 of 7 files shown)


ericnorris:
I noticed that some test failures are from syntax that is only available in newer PHP versions; I can fix these if you all feel it's worth moving forward with the PR.

Additionally, I think it is worth noting that this would solve the first problem in #286.

bshaffer:
Hi @ericnorris ! Thank you for submitting this!

I'd prefer to avoid large refactorings, as this library has been stable for some time and we will be releasing a new major version soon. I also see the following issues:

 - If we assume `expires_at` is set, some token types (JwtAccess and ID Tokens) which do not include this field will error or behave unexpectedly.
 - If we rely on the cache time to match with the token expiration time, we lose the ability to set a custom cache duration. It's possible that the CachePool being used requires a timeout different from the expiration time.

I'll write up a PR with potential changes and you can verify this will work for your implementation. I believe the main components are:

1. Update the cache to store the value of `expires_at`
2. Update `AuthTokenMiddleware` to check the stored expires, and fetch a new token if necessary

ericnorris:
@bshaffer thanks for taking a look!

> I'd prefer to avoid large refactorings, as this library has been stable for some time and we will be releasing a new major version soon. 

I'm not sure I agree that this is a _large_ refactoring, the actual code change is quite small and the majority of this PR is changing tests to be less specific to the underlying implementation.

> If we assume expires_at is set, some token types (JwtAccess and ID Tokens) which do not include this field will error or behave unexpectedly.

That's fair, I think this could be solved by adding in a conditional check for `expires_at`, and potentially checking for whatever field is in the JwtAccess and ID tokens.

> I'll write up a PR with potential changes and you can verify this will work for your implementation. I believe the main components are:
> 1. Update the cache to store the value of expires_at
> 2. Update AuthTokenMiddleware to check the stored expires, and fetch a new token if necessary

I'm not 100% sure it's simply the `AuthTokenMiddleware` class that has this issue - at the time of writing the [idiomatic PHP library does not use this class](https://github.com/googleapis/google-cloud-php/search?q=AuthTokenMiddleware&unscoped_q=AuthTokenMiddleware). It does, however, use the `FetchAuthTokenInterface` (as a `credentialsFetcher` option), which is how it is most often used at my employer.

Additionally, the `AuthTokenMiddleware` class _itself_ uses a `FetchAuthTokenInterface` ([here](https://github.com/googleapis/google-auth-library-php/blob/43146e564ec86685eded51a6cc1ee00f70e974b4/src/Middleware/AuthTokenMiddleware.php#L60)), which makes sense - it shouldn't care how the token is retrieved. I'd argue that it shouldn't need to check for expiration either, as the `FetchAuthTokenInterface` should always return a valid token.

This is why I targeted the `FetchAuthTokenCache` in my PR. It should only return tokens that are considered valid, which removes the need for consumers of the interface to double check.

I suspect any change to the `FetchAuthTokenCache` would largely end up similar to my PR, so I'd ask that you reconsider writing your own, or at least provide me with a path to make this PR more tenable. That being said, I'd be happy to take a look at an alternative PR to see if it addresses the production outages I've experienced.

bshaffer:
> I'm not sure I agree that this is a large refactoring

The only changes in this PR should be changes that are critical in order to deliver the required fix. I see a lot of changes here which do not appear critical. 

> I'm not 100% sure it's simply the AuthTokenMiddleware class that has this issue

I believe you're right, and I agree we should be able to assume the `FetchAuthTokenInterface::fetchAuthToken` returns a valid token.

> I'd ask that you reconsider writing your own

I am happy to work with you on this, but I do think we'll be able to merge the fix faster if you reviewed a new PR

ericnorris:
Understood! It would be helpful to have examples of what you consider extraneous to the fix - I did take some liberties as I consider `CacheTrait` internal to the API and as such doesn't count as a breaking change, but without specifics, it's hard to know. I'll list everything that I could think of below:

* Moving `$cache` and `$cacheOptions` to `CacheTrait` (e.g. https://github.com/googleapis/google-auth-library-php/pull/289/files#diff-20e3df603b4a9802ff034c7b9e121033R29-R30) and calling `initCacheTrait`

I'm not a fan of traits relying on class properties implicitly, and prefer to be explicit that they are owned by the trait. Can revert this back to each class having its own properties.

* Using `MemoryCacheItemPool` as a default if no cache is passed in (https://github.com/googleapis/google-auth-library-php/pull/289/files#diff-20e3df603b4a9802ff034c7b9e121033R39)

If you're using the `CacheTrait` it seems odd to have a null cache, but can revert this back and add null checks.

* Changing the `getCachedValue` method to take a cache miss callable (https://github.com/googleapis/google-auth-library-php/pull/289/files#diff-20e3df603b4a9802ff034c7b9e121033R57)

This is a common pattern for caching, [it's how Symfony does it](https://symfony.com/doc/current/components/cache.html#cache-contracts). It also allows consumers of the caching trait to have ultimate control over the cache item details (e.g. expiration), rather than needing it to be exposed in a separate `setCacheValue` method that ends up wrapping the `CacheItem` calls.

* Test changes

I feel pretty confident in the test changes, they either a) use an actual cache implementation to avoid needing mocks b) delete tests that are testing caching for things that don't care whether or not they are using a cache.

* Conditional `expires_at` check

Per your originally comment, `expires_at` may be missing. Should add a conditional check in `FetchAuthTokenCache`.

In general I'd like to think that the original changes I've suggested improve the code quality of the library; while fixing the bug is important to me, I'd also like to leave things nicer than how I've found it (e.g. fixing #286). For what it's worth, I would not mind if this requires waiting until the major version update.

If you'd like me to make any of the above changes or if there is something I've missed, let me know. If you really would rather just make a PR yourself, go for it :)

bshaffer:
I appreciate all these changes! It would be better to submit the features in separate pull requests, so each of them can be discussed and reviewed individually. Each of the suggestions above sound valid and would indeed improve the library, but in the pursuit of getting this issue resolved quickly, I'd prefer to focus on the one issue for now. Thanks again for your contributions.

ericnorris:
@bshaffer I understand the desire to keep the change to a minimum, but I am somewhat confused by what exactly constitutes a "feature" here - I wouldn't classify any of the changes I made as something that, on their own, are worthy of submitting as a feature.

The absolute minimum unit of code I could pull out of this particular implementation would be:

* the change to `getCachedValue` https://github.com/googleapis/google-auth-library-php/pull/289/files#diff-20e3df603b4a9802ff034c7b9e121033R57-R76, although it does need a conditional `expires_at` check and should cache for no longer than `lifetime`
* update callers of `getCachedValue`
* fix tests without making any changes to how they function

What's left in this PR, taking out the above, is the change to how the trait is initialized and the test updates. I don't think the trait initialization change is worth making a PR for, and since I had to fix the tests anyway, making them less finicky seemed like the right choice. I certainly wouldn't normally file a separate PR for changing tests.

Since your PR already implements an immediate fix, I'm not sure if I should either submit the absolute minimum PR, which would blow away your implementation, or leave this PR as a "v2" that could be reviewed and updated with requested changes after the immediate fix is merged.

The core of this PR is to use a `callable` in `getCachedValue` - while I think this is the right approach, if that change is objectionable and you'd rather check expiration manually or pass it in separately via a `setCachedValue` call, understood, let me know and I'll close the PR. If it is not, let me know if I should either split this into two PRs (one immediate and one for test updates, I'll drop the trait changes), or wait for further review.

bshaffer:
@ericnorris My thoughts are that this PR represents several different features (in addition to the original bug fix). I use the term "feature" but some of them are closer to "chore", since they are more along the lines of refactoring:

 1. chore: using `initCacheTrait` instead of setting the properties. 
 2. feature: default to memory cache
 3. feature: make `setCacheValue` take a callable
 4. chore: remove unneeded tests 

Each of these could be their own PR. If they were submitted individually, it would be easier for me as a maintainer to review them. The larger PR is also acceptable, but this can stall if one part of it requires further discussion. In the case of this PR, I would push back on the changing of the signature of `getCachedValue`, as this would be considered a BC-break. 

The best path forward is to rebase this PR after #291 is merged, and remove the `setCachedValue` changes, or close it and submit individual PRs containing the features above.

ericnorris:
Gotcha, it wasn't clear to me that changing `CacheTrait` would count as a backwards compatibility break - as I mentioned in the PR description, it feels internal to the workings of the library and I'd be very surprised if someone was using the `CacheTrait` in their code. For posterity, It'd have been helpful to hear that up front.

While I don't agree, I understand the desire to avoid BC breaks, and so if that is the case I will close this PR. Thanks for your time :)

bshaffer:
@ericnorris for the other changes you made, we would love to merge your contributions in a new pull request!

ericnorris:
Thanks @bshaffer! I actually took a look at your `v2` branch and it looks like you're making things better over there :). You are more than welcome to incorporate any of my ideas in that branch, or ping me for a review.

I'd also be happy to provide feedback on what I personally would be excited to see in a `v2` implementation (e.g. a credentials fetcher class for impersonation, something to enable auth caching for the `Google_Client` class, opinionated defaults, etc.).

In any case, glad to see you're working on it, looking forward to see how it turns out.

bshaffer:
@ericnorris I'd be very interested to hear your opinion on what you'd like to see in a `v2` implementation! Let's set something up.

PR #291, opened 2020-07-20
feat: checks cache expiration and fixes oauth2 cache key

bshaffer:
addresses #289 and #286 

cc @ericnorris

### DIFF
--- a/src/FetchAuthTokenCache.php
+++ b/src/FetchAuthTokenCache.php
@@ -84,14 +84,25 @@ class FetchAuthTokenCache implements
         // TODO: correct caching; enable the cache to be cleared.
         $cacheKey = $this->fetcher->getCacheKey();
         $cached = $this->getCachedValue($cacheKey);
-        if (!empty($cached)) {
-            return ['access_token' => $cached];
+        if (!empty($cached) && $token = json_decode($cached, true)) {
+            if (empty($token['expires_at'])) {
+                // If there is no expiration data, assume token is not expired.
+                // (for JwtAccess and ID tokens)
+                return $token;
+            }
+            if (time() < $token['expires_at']) {
+                // access token is not expired
+                return $token;
+            }
         }
 
         $auth_token = $this->fetcher->fetchAuthToken($httpHandler);
 
-        if (isset($auth_token['access_token'])) {
-            $this->setCachedValue($cacheKey, $auth_token['access_token']);
+        if (
+            isset($auth_token['access_token'])
+            || isset($auth_token['id_token'])
+        ) {
+            $this->setCachedValue($cacheKey, json_encode($auth_token));
         }
 
         return $auth_token;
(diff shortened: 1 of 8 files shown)


bshaffer:
We can parse the expiry from ID tokens and JWT self-signed tokens with something like this:

```php
use Firebase\JWT\JWT;
$headerString = substr($access_token, 0, strpos($access_token, '.'));
$header = JWT::jsonDecode(JWT::urlsafeB64Decode($headerString));
$expires = isset($header->exp) ? $header->exp : null;
```

I am not sure we want to parse this on every ID token / JWT self-signed token. For the self-signed tokens, we can return the expiry with the token somehow, but we're not doing that currently. We can expect ID tokens to expire after 60 minutes, so I think the tradeoff is not worth it.

bshaffer:
As for `ScopedAccessTokenMiddleware`, this is not a super useful class, and used if a user is pulling a raw access token from somewhere and wants to set it on a request. It would make sense to add caching here as well, although I agree it's not something that gets used as far as I know.

ericnorris:
> We can expect ID tokens to expire after 60 minutes, so I think the tradeoff is not worth it.

Understood, but doing an extra `json_decode` is likely negligible compared to the cost of actually generating the token (e.g. the network hop). When combined with caching, you'd be calling that extra `json_decode` once every 60 minutes.

For what it's worth, I expected access tokens to always expire after 60 minutes, but that is not the case in practice :). It doesn't feel right to assume the metadata server will do the right thing here and generate a fresh identity token every time.

All that said, my original PR didn't even address this, so addressing it in a different PR seems fine.

> It would make sense to add caching here as well, although I agree it's not something that gets used as far as I know.

As far as I can tell the `ScopedAccessTokenMiddleware` _already has_ caching, and is susceptible to the same bug that `FetchAuthTokenCache` has.

bshaffer:
> Understood, but doing an extra `json_decode` is likely negligible compared to the cost of actually generating the token

This is true. My concern is in implementations which use only in-memory caching, and so generate a new ID token on every request. This is the vast majority of cases, and will result in poorer performance for those cases. 

> It doesn't feel right to assume the metadata server will do the right thing here and generate a fresh identity token every time.

I understand not wanting to trust the metadata server. However, the behavior of the metadata server is to generate a new ID token on each request, with the current time for `iat` and a `3600` expiry. 

> As far as I can tell the `ScopedAccessTokenMiddleware` already has caching

Yes, that's true, I meant specifically the "expiry caching". The behavior of that class only deals with `string` tokens, so there's no way to handle expiry without breaking backwards compatibility (or adding new behavior, which isn't necessary).

>  It feels weird to have an extra expiry check in the `FetchAuthTokenCache` class... If lifetime is a big deal, you could pick whichever was shorter.

"lifetime" is a big deal because it's the current behavior of the library, which we don't want to change. Taking whichever is shorter is an interesting idea, and would probably work as well.

I'm on the fence about both setting expiry as the lifetime and decoding ID tokens. I'm interested to get feedback from @dwsupplee and @jdpedrie

PR #298, opened 2020-08-11
feat: add updateMetadata func to token cache

bshaffer:
This function, while not officially implemented as part of `FetchAuthTokenInterface`, is defined on all Credentials instances, and _should_ be a part of `FetchAuthTokenCache` as well.

This is required so the cache can take advantage of the logic in `updateMetadata` for defaulting to self-signed JWTs (see googleapis/gax-php#272)


src/FetchAuthTokenCache.php
```diff
@@ -188,4 +188,27 @@ public function getProjectId(callable $httpHandler = null)
...
+
+    /**
+     * Updates metadata with the authorization token.
+     *
+     * @param array $metadata metadata hashmap
+     * @param string $authUri optional auth uri
+     * @param callable $httpHandler callback which delivers psr7 request
+     * @return array updated metadata hashmap
+     */
+    public function updateMetadata(
```
dwsupplee, line 200: Should this end up calling the underlying implementation from the class it wraps? It looks like the implementations vary between credentials type.

bshaffer: This is a great point. The reason I didn't do it this way is because I thought this would bypass the cache, as calling `parent::updateMetadata` would result in `parent::fetchAuthToken` method being called instead of `FetchAuthTokenCache::fetchAuthToken`. However, I think now that `parent::updateMetadata` would execute`FetchAuthTokenCache::fetchAccessToken` instead, and so making this a wrapper call is the way to go.
 We just need to ensure a test confirms this.

bshaffer: After further consideration, this doesn't work because the Cache is a wrapper and not a subclass.

So we can:

1. Bypass the logic in `updateMetadata` - this is bad for obvious reasons
2. Bypass the cache in `fetchAuthToken` - this renders the class useless if `updateMetadata` is used
3. Cache the result of `fetchAuthToken` separately - this is somewhat useful, but still not optimal
4. Use the passed-in `$metadata` array to include the cached value, and update the existing classes so they only fill in the header if it isn't set - I don't think we want this behavior.
5. Leave `updateMetadata` out of the picture until v2 - this is the best option I see for now.

bshaffer:
After *even further* consideration I have done the following

 - Added logic to bypass fetching a token if the `authorization` key is set in the metadata. This allows the cached access token to propagate w/o the fetcher fetching a new one
 - Added `UpdateMetadataInterface` for better implementation checking
 - Added `UpdateMetadataTrait` for dryer/cleaner code

Also important:
 - moved the `AUTH_METADATA_KEY` constant from `CredentialsLoader` into new `UpdateMetadataInterface`
 - Implemented `UpdateMetadataInterface` in `GCECredentials`, `FetchAuthTokenCache` and `CredentialsLoader`
 - used `UpdateMetadataTrait` in `GCECredentials` and `CredentialsLoader`
 - Added tests for the new functionality!

src/UpdateMetadataInterface.php
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+/**
+ * Describes a Credentials object which supports updating request metadata
+ * (request headers).
+ */
+interface UpdateMetadataInterface
```
dwsupplee, line 24: Should the `IAMCredentials` class implement this too? The auth keys seem to differ in that implementation a bit though, which makes it seem like the constant being on this class might not be ideal.

src/Credentials/GCECredentials.php
```diff
@@ -58,7 +59,8 @@
 class GCECredentials extends CredentialsLoader implements
     SignBlobInterface,
     ProjectIdProviderInterface,
-    GetQuotaProjectInterface
+    GetQuotaProjectInterface,
+    UpdateMetadataInterface
```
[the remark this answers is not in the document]
bshaffer: For some reason I thought this didn't extend credentials loader... So you're right, this can be removed

src/CredentialsLoader.php
```diff
@@ -26,13 +26,16 @@
...
+    FetchAuthTokenInterface,
+    UpdateMetadataInterface
 {
     const TOKEN_CREDENTIAL_URI = 'https://oauth2.googleapis.com/token';
     const ENV_VAR = 'GOOGLE_APPLICATION_CREDENTIALS';
     const WELL_KNOWN_PATH = 'gcloud/application_default_credentials.json';
     const NON_WINDOWS_WELL_KNOWN_PATH_BASE = '.config';
-    const AUTH_METADATA_KEY = 'authorization';
+
+    use UpdateMetadataTrait;
```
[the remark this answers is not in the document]
bshaffer: See above... I thought we needed the logic here in two places, but I was mistaken. I can move it back into credentials loader...

src/Credentials/GCECredentials.php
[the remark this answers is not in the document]
bshaffer: Removed!

src/UpdateMetadataInterface.php
bshaffer: I don't really know how the `IAMCredentials` class is being used, so I left it out for now. It doesn't even extend `FetchAuthTokenInterface` so it makes more sense for me to leave it out.

src/CredentialsLoader.php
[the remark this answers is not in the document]
bshaffer: Removed `UpdateMetadataTrait`!

src/UpdateMetadataInterface.php
dwsupplee: Do we need the interface then? Perhaps I missed it but it looks like anything that extends `CredentialsLoader` will implement the `updateMetadata` method. If the goal is to provide better implementation checking, we can do that by checking if the credentials are an instance of `CredentialsLoader` I believe.

dwsupplee: After thinking on this a bit, the interface is better, let's keep it.

bshaffer: The interface is used by `FetchAuthTokenCache`, which does not extend `CredentialsLoader`. It's only real function is to ensure that the `updateMetadata` function exists on the fetcher, since we can't add that method to `FetchAuthTokenInterface` without breaking backwards compatibility.

It isn't strictly necessary but I think it's best practice to have it. Not only is the interface cleaner, but it ensures that the function signature is compatible.

bshaffer:
Any thoughts on this @dwsupplee or @jdpedrie ? I'm hoping to get this sucker into `v1.12.0`

jdpedrie (approved):
LGTM

Issue #297, opened 2020-08-03
DDoS-ing the Metadata Server.


xvzf:
Hello,

With high volume traffic, a check (on each interaction with Google APIs!!!) if the application is running on GCP by calling the metadata server is becoming quite a bottleneck response-time wise.

The specific code is located here:
https://github.com/googleapis/google-auth-library-php/blob/c438cdaadd2b1fa334c4da7560cad238625d8c10/src/Credentials/GCECredentials.php#L258

#### Environment details
  - OS: Linux
  - PHP version: 7.x
  - Package name and version: irrelevant

xvzf:
@dwsupplee since you tagged it as investigating, it's not just PHP 7.4 but all 7.x releases

ericnorris:
At Etsy we've encountered the same issue; during periods of high enough traffic (e.g. backfills) we've seen metadata server outages that result in the following backtrace:

```
Could not load the default credentials. Browse to https://developers.google.com/accounts/docs/application-default-credentials for more information
DomainException trace:
[Dev_Backtrace]:
#12 getCredentials() /GoogleCloudCredentials.php:70
...
```

...where the `getCredentials` is this call:

 https://github.com/googleapis/google-auth-library-php/blob/c05c0a5c24759949f16c8d868858f05ab3899413/src/ApplicationDefaultCredentials.php#L148

...which ends up calling the `onGce` method in the original comment.

We use the `FetchAuthTokenCache` to avoid hammering the metadata server with auth calls, but this `onGce` call went unnoticed until we experience outages. Internally we are trying to resolve this by passing in a special `$httpHandler` to `ApplicationDefaultCredentials::getCredentials`, but ideally the library would offer a PSR-6 way of caching this, much like how the `FetchAuthTokenCache` works.

bshaffer:
If we added a new class, like @ericnorris suggested, we could have something like this:

```php
// GCE cache implementing CacheItemPoolInterface
$cache = new FileCache();
// optional GCE cache config
$cacheConfig = [
    'lifetime' => 1500,
    'prefix' => '',
];
$gceCache = new GCECache($cache, $cacheConfig);

// fetch default credentials with the cache
$credentials = ApplicationDefaultCredentials::getCredentials($scope, null, null, null, null, $gceCache);
// check directly with the cache
$onGce = GCECredentials::getCredentials(null, $gceCache);
```

[...]

PR #301, opened 2020-09-01
feat: adds caching for call to ::onGce

bshaffer:
Addresses #297 

 - New class `GCECache`
 - `ApplicationDefaultCredentials` automatically caches `onGce` response using the same `$cacheOptions` and `$cache` passed into `getCredentials` and `getIdTokenCredentials`
 - `$cacheOptions` are the same for `GCECache` and `FetchAuthTokenCache`. Is this ok?
 - `key` option allows flexibility for cache key, which defaults to `google_auth_on_gce_cache` (do we need this?)

The biggest issue I see is not being able to set a different lifetime/prefix between GCE and Access Token caches. To address this, `ApplicationDefaultCredentials` could create a new `$cacheOptions` for `GCECache` and allow for keys like `gce_lifetime` and `gce_prefix`

```php
$cacheOptions = [
    'lifetime' => 1500,
    'prefix' => 'access_token_',
    'gce_lifetime' => 3000,
    'gce_prefix' => 'on_gce_',
];
$creds = ApplicationDefaultCredentials::getCredentials(null, null, $cacheOptions, $cache);
```

### DIFF
--- a/src/GCECache.php
+++ b/src/GCECache.php
@@ -52,7 +52,6 @@ class GCECache
         $this->cacheConfig = array_merge([
             'lifetime' => 1500,
             'prefix' => '',
-            'key' => self::GCE_CACHE_KEY,
         ], (array) $cacheConfig);
     }
 
@@ -69,7 +68,7 @@ class GCECache
             return GCECredentials::onGce($httpHandler);
         }
 
-        $cacheKey = $this->cacheConfig['key'];
+        $cacheKey = self::GCE_CACHE_KEY;
         $onGce = $this->getCachedValue($cacheKey);
 
         if (is_null($onGce)) {
--- a/src/GCECache.php
+++ b/src/GCECache.php
@@ -21,8 +21,19 @@ use Google\Auth\Credentials\GCECredentials;
 use Psr\Cache\CacheItemPoolInterface;
 
 /**
- * A class to implement caching for any object implementing
- * FetchAuthTokenInterface
+ * A class to implement caching for calls to GCECredentials::onGce. This class
+ * is used automatically when you pass a `Psr\Cache\CacheItemPoolInterface`
+ * cache object to `ApplicationDefaultCredentials::getCredentials`.
+ *
+ * ```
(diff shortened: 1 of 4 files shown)


dwsupplee:
> The biggest issue I see is not being able to set a different lifetime/prefix between GCE and Access Token caches. To address this, ApplicationDefaultCredentials could create a new $cacheOptions for GCECache and allow for keys like gce_lifetime and gce_prefix

My vote goes with this, if it isn't much trouble

bshaffer:
This change uses Memory caching by default, and SysV caching via the following:

```php
use Google\Cloud\Bigtable\BigtableClient;

$credentialsConfig = ['authCache' => new Google\Auth\SysvCacheItemPool()];
$bigtable = new BigTableClient(['credentialsConfig' => $credentialsConfig]);
```

bshaffer:
Note: this will require an update in [`Google\Cloud\Core\ClientTrait::onGce`](https://github.com/googleapis/google-cloud-php/blob/master/Core/src/ClientTrait.php#L247) to pass the caching options through. Fortunately, the caching options are available in [ClientTrait::configureAuthentication](https://github.com/googleapis/google-cloud-php/blob/master/Core/src/ClientTrait.php#L97), so this shouldn't be difficult.

bshaffer:
@dwsupplee done! PTAL!

ericnorris:
LGTM, thanks for digging into this.

Unfortunately we haven't had time recently to attempt upstreaming the other fix I mentioned, but we'll definitely try to do that at some point.

src/GCECache.php
```diff
@@ -0,0 +1,81 @@
...
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+use Google\Auth\Credentials\GCECredentials;
+use Psr\Cache\CacheItemPoolInterface;
+
+/**
+ * A class to implement caching for any object implementing
```
dwsupplee, line 24: The description needs to be updated

bshaffer: Good catch!! done.

Issue #303, opened 2020-09-10
Add support for Workload Identity on the GKE


mruoss:
**Is your feature request related to a problem? Please describe.**
When running an application on the GKE I would like to be able to get credentials for a workload identity. The class `GCECredentials` already implements a great deal, however it only gets the credentials for the default service account. For workload identities, I would like to specify the google service account (GSA) the credentials are fetched for.

 **Describe the solution you'd like**
By refactoring the `GCECredentials` it is fairly easy to pass the service account name as parameter in the constructor. Or maybe adding a separate class like GKEWorkloadIdentityCredentials that inherits from GCECredentials would make sense.

bshaffer:
@mruoss can you explain how you were able to get everything working?

mruoss:
Thanks @bshaffer, I just figured out, this is actually really needed. Reopening therefore.

mruoss:
I thought it was working but in fact it wasn't. I got a token, but for the wrong GSA.

bshaffer:
@mruoss so essentially, the ask is for `GCECredentials` to have an option to request more than just the default service account from the metadata server?

```php
$gce = new GCECredentials(null, null, null, null, $serviceAccountIdentity);
$accessToken = $gce->fetchAuthToken();
```

Behind the scenes this would just change the metadata URL paths from `*/default/*` to `*/$serviceAccountIdentity/*`:

I don't see how we could add this to `ApplicationDefaultCredentials`, as no other credential type has the flexibility of only needing the service account name.

I've added #304 for this feature. Please take a look and see if this solution will work for you!

mruoss:
Hi @bshaffer 

Yes, that is exactly the request. #304 seems to do the job.

mruoss:
@bshaffer Can you roughly guess how long it's gonna take for this feature to be released?

bshaffer:
@mruoss we just merged the PR, so it's available now via the `dev-master` packagist release. I'll push a new version by the end of week. Does that work for you?

mruoss:
That sounds perfect, thank you very much!

bshaffer:
@mruoss this is now released in [`v1.13.0`](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.13.0)!

PR #304, opened 2020-09-10
feat: add service account identity support to GCECredentials

bshaffer:
Adds the ability to change from using the `default` service account in `GCECredentials`

See #303

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -160,15 +160,27 @@ class GCECredentials extends CredentialsLoader implements
     private $quotaProject;
 
     /**
+     * @var string|null
+     */
+    private $serviceAccountIdentity;
+
+    /**
      * @param Iam $iam [optional] An IAM instance.
      * @param string|array $scope [optional] the scope of the access request,
      *        expressed either as an array or as a space-delimited string.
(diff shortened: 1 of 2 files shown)


src/Credentials/GCECredentials.php
```diff
@@ -187,41 +199,82 @@ public function __construct(Iam $iam = null, $scope = null, $targetAudience = nu
...
+     *   account identity name to use instead of "default".
+     * @return string
+     */
+    public static function getClientNameUri($serviceAccountIdentity = null)
+    {
+        $base = 'http://' . self::METADATA_IP . '/computeMetadata/';
+        $base .= self::CLIENT_ID_URI_PATH;
+
+        if ($serviceAccountIdentity) {
+            return str_replace(
```
mruoss, line 247: Instead of replacing 'default' with $serviceAccountIdentity, 'default' could be the default value of the argument. It would simplify the code...

bshaffer: This would then perform an extra string replace opereration on every call, which I would like to avoid.

If not for BC, I'd change the value of the `TOKEN_URI_PATH` to contain a `%s`.

mruoss: True, I see now.

jdpedrie (approved):

PR #306, opened 2020-09-15
feat: add difference between default and user-defined scope

bshaffer:
This PR allows the calling libraries (GAPIC, veneer, Apiary) to pass `$defaultScope` in separately from `$scope`. This allows `ServiceAccountCredentials` to use self-signed JWTs more frequently. Before, default scopes were preferred to self-signed JWTs, but if we can differentiate between user-defined and default scopes, we can prefer user-defined scopes to self-signed JWTs, but self-signed JWTs to default scopes.

So in effect, our clients can use self-signed JWTs when:

 - A service account is being used
 - A scope has not been set by the user
 - A target_audience has not been set by the user
 - An api_endpoint has not been set by the user

Related PRs:

 - downstream changes to GAX: https://github.com/googleapis/gax-php/pull/272/files
 - See also the [NodeJS implementation](https://github.com/googleapis/google-auth-library-nodejs/pull/1054)

**Note**: In this implementation, `ServiceAccountCredentials` no longer calls `fetchAuthToken` with default scopes. If they are invoked without an `$authUri`, no credentials will be provided. This is not a BC break, as it only will take place if `$defaultScopes` is utilized by the calling party. But we need to verify this is the correct behavior.


src/CredentialsLoader.php
```diff
@@ -133,18 +133,23 @@ public static function fromWellKnownFile()
      * @param array $jsonKey the JSON credentials.
      * @return ServiceAccountCredentials|UserRefreshCredentials
      */
-    public static function makeCredentials($scope, array $jsonKey)
-    {
+    public static function makeCredentials(
+        $scope,
+        array $jsonKey,
+        $defaultScope = null
```
jdpedrie, line 139: Need this added to the docblock

jdpedrie (approved):
One nit.

PR #311, opened 2020-10-03
fix: better FetchAuthTokenCache and getLastReceivedToken

bshaffer:
- Improves `FetchAuthTokenCache::updateMetadata` so that **self-signed JWTs can be used** with `FetchAuthTokenCache`
 - Improves `getLastReceivedToken` in OAuth2 so the results can be cached by `FetchAuthTokenCache`
 - Improves `getLastReceivedToken` in `ServiceAccountCredentials` and `ServiceAccountJwtAccessCredentials` so they will return the self-signed token when self-signed JWTs are used


src/FetchAuthTokenCache.php
```diff
@@ -212,20 +193,62 @@ public function updateMetadata(
...
-        // again by the fetcher
-        $result = $this->fetchAuthToken($httpHandler);
-
-        if (isset($result['access_token'])) {
-            $metadata[self::AUTH_METADATA_KEY] = [
-                'Bearer ' . $result['access_token']
-            ];
+        $cached = $this->fetchAuthTokenFromCache();
+        if ($cached) {
+            // Set the access token in the `Authentication` metadata header so
```
dwsupplee, line 198:
```suggestion
            // Set the access token in the `Authorization` metadata header so
```

tests/FetchAuthTokenCacheTest.php
```diff
@@ -177,6 +181,54 @@ public function testUpdateMetadataWithoutCache()
...
+
+        $lastReceivedToken = $cachedFetcher->getLastReceivedToken();
+        $this->assertArrayHasKey('access_token', $lastReceivedToken);
+        $this->assertEquals($token, $lastReceivedToken['access_token']);
+
+        // Ensure token is cached
+        $metadata2 = $cachedFetcher->updateMetadata([], 'http://test-auth-uri');
+        $this->assertEquals($metadata, $metadata2);
+
+        // TODO: Make this work
```
dwsupplee, line 226: Just a reminder about this todo

dwsupplee (approved):

tests/FetchAuthTokenCacheTest.php
bshaffer: done!

PR #320, opened 2021-02-05
chore: prepare v1.15.0

bshaffer:


jdpedrie (approved):

bshaffer:
Note: We need to fix docs generation before merging

bshaffer:
This should be fixed in https://github.com/googleapis/google-auth-library-php/pull/321

PR #324, opened 2021-02-17
Show how to use a specific JSON key.

imdhemy:
**What?**
Allow creation of middleware using a JSON key supplied as an array.

**Why?**
To allow projects to handle multiple service accounts.

**Extra info**
I'm working on a project that handles in-app purchases for multiple mobile applications. Verifying the in-app purchase receipt requires Google auth, and hence there are more than one service accounts used, I could not rely on a single environment variable `GOOGLE_APPLICATION_CREDENTIALS`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ used when you set up your protected resource as the target audience. See how to
 [secure your IAP app with signed headers](https://cloud.google.com/iap/docs/signed-headers-howto).
 
 #### Call using a specific JSON key
-If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, You can
+If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, you can
  do this:
  
 ```php
--- a/README.md
+++ b/README.md
@@ -173,7 +173,7 @@ used when you set up your protected resource as the target audience. See how to
 [secure your IAP app with signed headers](https://cloud.google.com/iap/docs/signed-headers-howto).
 
 #### Call using a specific JSON key
-If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, Yo can
+If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, You can
  do this:
  
 ```php
--- a/README.md
+++ b/README.md
@@ -172,6 +172,47 @@ For invoking Cloud Identity-Aware Proxy, you will need to pass the Client ID
 used when you set up your protected resource as the target audience. See how to
 [secure your IAP app with signed headers](https://cloud.google.com/iap/docs/signed-headers-howto).
 
+#### Call using a specific JSON key
+If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, Yo can
+ do this:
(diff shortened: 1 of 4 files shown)


bshaffer:
I do not think this is necessary. If you'd like to create a request with a specific JSON key, you can do this:

```php
$creds = CredentialsLoader::makeCredentials($scope, $jsonKey);

// optional caching
// $creds = new FetchAuthTokenCache($creds, $cacheConfig, $cache);

$middleware = new AuthTokenMiddleware($creds);
```

imdhemy:
Hi @bshaffer,

Awesome, So I:
- Reverted back.
- Updated the README to show how to make a call using a specific JSON key.

Regards,

kumamidori:
@imdhemy  @bshaffer If we take the above approach, does it mean that we cannot use the useful features of the `Google_Client class` and have to use the `GuzzleHttp\Client class`?

@imdhemy 
I thought it would be easier to understand if there was a use statement for "Client" in the README.

bshaffer:
> @imdhemy  @bshaffer If we take the above approach, does it mean that we cannot use the useful features of the `Google_Client class` and have to use the `GuzzleHttp\Client class`?

Just use the `credentials` option or call `setAuthConfig`:

```php
// Using credentials option
$client = new Google\Client(['credentials' => $jsonKey]);
// Using setAuthConfig
$client = new Google\Client();
$client->setAuthConfig($jsonKey);
```

Then pass this client to a new service class to use that key instead.

README.md
```diff
@@ -172,6 +172,47 @@ For invoking Cloud Identity-Aware Proxy, you will need to pass the Client ID
 used when you set up your protected resource as the target audience. See how to
 [secure your IAP app with signed headers](https://cloud.google.com/iap/docs/signed-headers-howto).
 
+#### Call using a specific JSON key
+If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CREDENTIALS` environment variable, Yo can
```
bshaffer, line 176: "Yo can" should be "you can"

imdhemy: @bshaffer Fixed

bshaffer:
@imdhemy it looks like you haven't signed the CLA. I cannot merge this until you do!

bshaffer (approved):

imdhemy:
@googlebot I signed it!

README.md
```diff
@@ -172,6 +172,47 @@ For invoking Cloud Identity-Aware Proxy, you will need to pass the Client ID
...
+ do this:
+ 
+```php
+use Google\Auth\CredentialsLoader;
+use Google\Auth\Middleware\AuthTokenMiddleware;
+use GuzzleHttp\HandlerStack;
+
```
kumamidori, line 183: @imdhemy  @bshaffer  I suggest adding a use statement for `Client`.

```
$client = new Client([
  'handler' => $stack,
  'base_uri' => 'https://www.googleapis.com',
  'auth' => 'google_auth'  // authorize all requests
]);
```

Or is it more common to use the following?
https://github.com/googleapis/google-auth-library-php/pull/324#issuecomment-793060136

```
// Using credentials option
$client = new Google\Client(['credentials' => $jsonKey]);
// Using setAuthConfig
$client = new Google\Client();
$client->setAuthConfig($jsonKey);
```

bshaffer: `Client` here is `GuzzleHttp\Client`

PR #326, opened 2021-03-08
Update README (how to use a specific JSON key)

kumamidori:
I suggest the following PR change to a common code.
https://github.com/googleapis/google-auth-library-php/pull/324


kumamidori:
@googlebot I signed it!

bshaffer:
This example would be appropriate in the README of https://github.com/googleapis/google-api-php-client, but not here. This library does not contain `Google\Client`, it is a generally useful auth dependency.

kumamidori:
Thanks!

PR #330, opened 2021-04-16
feat: Allow caching the result of ADC::onGce separately

MrMage:
Motivation: Lets us cache the result of ADC::onGce without returning a `FetchAuthTokenCache`, as that might lead to nested caches.

Also see https://github.com/googleapis/google-cloud-php/pull/3959 and https://github.com/googleapis/gax-php/pull/318.

### DIFF
--- a/src/ApplicationDefaultCredentials.php
+++ b/src/ApplicationDefaultCredentials.php
@@ -144,6 +144,10 @@ class ApplicationDefaultCredentials
      * @param string|array $defaultScope The default scope to use if no
      *   user-defined scopes exist, expressed either as an Array or as a
      *   space-delimited string.
+     * @param CacheItemPoolInterface $onGceCache A cache implementation to store
+     *        the results of internal helper methods, without causing a
+     *        `FetchAuthTokenCache` to be returned. Will only be used if `$cache` is
+     *        null.
      *
      * @return CredentialsLoader
      * @throws DomainException if no implementation can be obtained.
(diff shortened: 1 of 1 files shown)


bshaffer:
Thank you for submitting this PR!

This change seems okay, but I don't quite understand why it's useful. Can you provide an instance where the result would be nested caches, or provide the use-case for why someone would need to keep those caches separate?

MrMage:
Sorry, I should have provided more details! Currently, e.g. `RequestWrapperTrait` in the Google Cloud SDK passes a `null` auth cache to `ADC::getCredentials`, in order to avoid nesting two `FetchAuthTokenCache` (see https://github.com/googleapis/google-cloud-php/pull/3959). However, this means that `onGce` will also be called with a `null` auth cache, i.e. that request needs to be sent at least once on every request to my web app. This adds ~1ms of avoidable latency to my response times.

To allow caching the result of `onGce` _without_ nesting `FetchAuthTokenCache`, this PR adds a separate cache argument that only gets forwarded to `onGce`.

bshaffer:
I could be missing something, but it seems like we could simply avoid nested `FetchAuthTokenCache` by checking to make sure the `$cache` object is not `FetchAuthTokenCache` before wrapping it.

If this is sufficient, it seems like a better solution to me.

MrMage:
> I could be missing something, but it seems like we could simply avoid nested `FetchAuthTokenCache` by checking to make sure the `$cache` object is not `FetchAuthTokenCache` before wrapping it.
> 
> If this is sufficient, it seems like a better solution to me.

That indeed sounds like a better solution and updated https://github.com/googleapis/google-cloud-php/pull/3959 accordingly. Please have a look and let me know what you think :-)

MrMage:
Made obsolete by https://github.com/googleapis/google-cloud-php/pull/3959.

Issue #328, opened 2021-03-10
Issue with signBlob not using cached token in the FetchAuthTokenCache class


Babacooll:
Hello,

As we wanted to reduce the calls to the Google metadata endpoint to retrieve fresh tokens we decided to use the `\Google\Auth\FetchAuthTokenCache `class with a `\Psr\Cache\CacheItemPoolInterface` implementation using Redis. This works very well for most services such as BigTable, PubSub, ...

But when using the `FetchAuthTokenCache->signBlob() `method (to generate secure URLs) it actually delegates the signing to `\Google\Auth\Credentials\GCECredentials->signBlob()` method which doesn't support cached token. This ends up with 1 call to the Google metadata service per user request.

There is no clean solution out of the box but this is the fix we applied on our fork https://github.com/oat-sa/google-auth-library-php/commit/c1f3752613a6dad94236d5731bb15504d20a10c4. This allows passing the cached token from the `FetchAuthTokenCache`to the `GCECredentials`.

Do you have any other ideas on how to fix this?

Thank you very much!

bshaffer:
Wow, super good catch and very interesting problem. Thanks for reporting this and demonstrating a solution. I see a few ways to do this:

1. Check for `GCECredentials` and set the last received token manually (your solution)
2. Add a `SetLastReceivedTokenInterface` and set the last received token manually (like your solution, but using something generic instead of checking for a specific credentials class)
3. Add `$accessToken` as an optional third argument to `SignBlobInterface::signBlob`
4. Add `$accessToken` as an optional third argument just to `GCECredentials::signBlob`

I'm not sure i see much of an advantage to any of these options over the other. None of them are particularly great, and none of them are particularly terrible either.

PR #340, opened 2021-06-10
fix: ensure cached tokens are used for GCECredentials::signBlob

bshaffer:
fixes https://github.com/googleapis/google-auth-library-php/issues/328

@Babacooll PTAL

### DIFF
--- a/src/FetchAuthTokenCache.php
+++ b/src/FetchAuthTokenCache.php
@@ -142,6 +142,13 @@ class FetchAuthTokenCache implements
             );
         }
 
+        // Fetch access token from cache for signing a blob
+        if ($this->fetcher instanceof Credentials\GCECredentials) {
+            $cached = $this->fetchAuthTokenFromCache();
+            $accessToken = isset($cached['access_token']) ? $cached['access_token'] : null;
+            return $this->fetcher->signBlob($stringToSign, $forceOpenSsl, $accessToken);
+        }
+
         return $this->fetcher->signBlob($stringToSign, $forceOpenSsl);
     }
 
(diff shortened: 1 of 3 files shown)


src/FetchAuthTokenCache.php
```diff
@@ -142,6 +142,14 @@ public function signBlob($stringToSign, $forceOpenSsl = false)
...
 
+        // Pass the access token from cache to GCECredentials for signing a blob.
+        // This saves a call to the metadata server when a cached token exists.
+        if ($this->fetcher instanceof Credentials\GCECredentials) {
+            $cached = $this->fetchAuthTokenFromCache();
+            $accessToken = isset($cached['access_token']) ? $cached['access_token'] : null;
+            return $this->fetcher->signBlob($stringToSign, $forceOpenSsl, $accessToken);
+        }
+
         return $this->fetcher->signBlob($stringToSign, $forceOpenSsl);
```
jdpedrie, line 153:
```suggestion
        $accessToken = null;
        if ($this->fetcher instanceof Credentials\GCECredentials) {
            $cached = $this->fetchAuthTokenFromCache();
            $accessToken = isset($cached['access_token']) ? $cached['access_token'] : null;
        }

        return $this->fetcher->signBlob($stringToSign, $forceOpenSsl, $accessToken);
```

Just the tiniest of nits, feel free to totally ignore this.

jdpedrie (approved):

src/FetchAuthTokenCache.php
bshaffer: I am going to ignore this, but only because the _only_ fetcher which accepts the third argument is `GCECredentials`, and so I think it's misleading to pass the third argument in, even if null, to the other fetchers.

PR #341, opened 2021-06-10
feat: allow ServiceAccountJwtAccessCredentials to sign scopes

bshaffer:
See https://google.aip.dev/auth/4111#scope-vs-audience

cc @shinfan

### DIFF
--- a/src/Credentials/ServiceAccountJwtAccessCredentials.php
+++ b/src/Credentials/ServiceAccountJwtAccessCredentials.php
@@ -57,8 +57,10 @@ class ServiceAccountJwtAccessCredentials extends CredentialsLoader implements
      *
      * @param string|array $jsonKey JSON credential file path or JSON credentials
      *   as an associative array
+     * @param string|array $scope the scope of the access request, expressed
+     *   either as an Array or as a space-delimited String.
      */
-    public function __construct($jsonKey)
+    public function __construct($jsonKey, $scope = null)
     {
         if (is_string($jsonKey)) {
             if (!file_exists($jsonKey)) {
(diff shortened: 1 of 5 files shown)


shinfan (approved):

src/Credentials/ServiceAccountJwtAccessCredentials.php
```diff
@@ -128,10 +132,17 @@ public function updateMetadata(
...
     {
         $audience = $this->auth->getAudience();
-        if (empty($audience)) {
+        $scope = $this->auth->getScope();
+        if (empty($audience) && empty($scope)) {
             return null;
         }
 
+        if (!empty($audience) && !empty($scope)) {
+            throw new UnexpectedValueException(
```
dwsupplee, line 141:
```suggestion
            throw new \UnexpectedValueException(
```

Would we be able to document this being thrown? Wondering if we need a test for this as well.

src/OAuth2.php
```diff
@@ -437,9 +436,18 @@ public function toJwt(array $config = [])
...
+        if (!(is_null($this->getAudience()))) {
+            $assertion['aud'] = $this->getAudience();
+        }
+
         if (!(is_null($this->getScope()))) {
             $assertion['scope'] = $this->getScope();
         }
+
+        if (empty($assertion['scope']) && empty($assertion['aud'])) {
+            throw new \DomainException('one of scope or aud should not be null');
```
dwsupplee, line 448: Do we need a test to exercise this?

dwsupplee (changes requested):

src/OAuth2.php
bshaffer: good point, added!

src/Credentials/ServiceAccountJwtAccessCredentials.php
bshaffer: Good catch and yes, I added a test!

I'm happy to add `@throws`. In this library in general (not that it's the right thing to do), we haven't been documenting thrown exceptions.

dwsupplee: I trust your judgement 👍

dwsupplee (approved):

PR #345, opened 2021-06-21
chore: drop support for php < 5.6

bshaffer:
- drop support for PHP 5.4 and 5.5
 - drops support for Guzzle 5
 - drops support for Firebase/php-jwt < 5.0

Let's create a minor version release for this quickly, and then drop support for 5.6 as well.


.github/workflows/tests.yml
```diff
@@ -35,14 +33,17 @@ jobs:
...
+              run: sh .github/apply-phpunit-patches.sh
             - name: Run Script
               run: vendor/bin/phpunit
     test_lowest:
-        runs-on: ${{matrix.operating-system}}
+        runs-on: ubuntu-latest
         strategy:
             matrix:
                 operating-system: [ ubuntu-latest ]
-                php: [ "5.6", "7.0", "7.1", "7.2" ]
```
vishwarajanand, line 45: Is removing `"7.0", "7.1"` intentional?

.github/workflows/tests.yml
```diff
@@ -35,14 +33,17 @@ jobs:
...
                   composer remove --dev --ignore-platform-reqs phpunit/phpunit
                   composer require --dev --ignore-platform-reqs --update-with-all-dependencies phpunit/phpunit:^7
+            - if: ${{ matrix.php == '5.6' || matrix.php == '7.0' || matrix.php == '7.1' }}
+              name: Run PHPUnit Patches
+              run: sh .github/apply-phpunit-patches.sh
             - name: Run Script
               run: vendor/bin/phpunit
     test_lowest:
-        runs-on: ${{matrix.operating-system}}
+        runs-on: ubuntu-latest
```
vishwarajanand, line 42: This is better, avoids repetition from `operating-system: [ ubuntu-latest ]` ?
`runs-on: ${{matrix.operating-system}}`

.github/workflows/tests.yml
bshaffer: Yes, I wanted to simplify the tests, we didn't really need all three.

.github/workflows/tests.yml
bshaffer: As the matrix only applies to one test (and not to all of them), it doesn't save us any repetition, it only adds an extra line of code.

.github/apply-phpunit-patches.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+# Script used from php-webdriver/php-webdriver
+
+# All commands below must no fail
```
vishwarajanand, line 5:
```suggestion
# All commands below must not fail
```

.github/workflows/tests.yml
```diff
@@ -35,14 +33,16 @@ jobs:
                 command: |
                   composer remove --dev --ignore-platform-reqs phpunit/phpunit
                   composer require --dev --ignore-platform-reqs --update-with-all-dependencies phpunit/phpunit:^7
+            - if: ${{ matrix.php == '5.6' || matrix.php == '7.0' || matrix.php == '7.1' }}
```
vishwarajanand, line 36: If there are too many versions, you may use:

```suggestion
            - if: ${{ contains(fromJson('["5.6", "7.0", "7.1"]'), matrix.php)}}
```

composer.json
```diff
@@ -9,16 +9,18 @@
     "docs": "https://googleapis.github.io/google-auth-library-php/main/"
   },
   "require": {
-    "php": ">=5.4",
+    "php": ">=5.6",
     "firebase/php-jwt": "~2.0|~3.0|~4.0|~5.0",
-    "guzzlehttp/guzzle": "^5.3.1|^6.2.1|^7.0",
-    "guzzlehttp/psr7": "^1.7|^2.0",
+    "guzzlehttp/guzzle": "^6.2.1|^7.0",
+    "guzzlehttp/psr7": "^1.7",
```
vishwarajanand, line 15: Dropping `guzzlehttp/psr7:^2.0` ?

bshaffer: I'm not sure how that happened - it may have been on accident. I've added it back.

vishwarajanand:
I checked the tests log and found side issues which can be taken separately:

1. `Middleware\SimpleMiddlewareTest::testTest` and `OAuth2JwtTest::testFailDecodeWithoutSigningKeyId` do not do any assertion. 
2. Some tests warnings due to deprecation in phpunit~v9 like `assertContains`, `assertInternalType` and `@expectedException`.

vishwarajanand (approved):
I posted 1 nit and a clarification question but LGTM hence approving.

bshaffer:
> 1. `Middleware\SimpleMiddlewareTest::testTest` and `OAuth2JwtTest::testFailDecodeWithoutSigningKeyId` do not do any assertion.

I added assertions for both of these!

> 2. Some tests warnings due to deprecation in phpunit~v9 like `assertContains`, `assertInternalType` and `@expectedException`.

I cannot fix these at the moment until we drop support for more PHP versions, but as they're only warnings I believe it's okay for now.

PR #350, opened 2021-06-25
chore: switch cs fixers

bshaffer:


jdpedrie (approved):

saranshdhingra (approved):
Although tests don't run in PHP 8.x as I get `Call to undefined function each() ...` as this func removed in PHP 8.0.
Perhaps we should add an issue to track this.
But the git build checks pass :)

bshaffer:
@saranshdhingra we do something hacky to get around this (for now) which is conditionally install `phpunit:^7` for our tests on PHP 8.0. See [.github/workflows/tests.yml](https://github.com/googleapis/google-auth-library-php/blob/master/.github/workflows/tests.yml#L37)

PR #353, opened 2021-07-01
feat: add loading and executing of default client cert source

bshaffer:
Adds logic for detecting and loading the default client certificates for mTLS

Implementation doc: https://google.aip.dev/auth/4114
Related: https://github.com/googleapis/gax-php/pull/325


src/CredentialsLoader.php
```diff
@@ -247,4 +250,52 @@ public function updateMetadata(
...
+            exec(implode(' ', $cmd), $output, $returnVar);
+
+            if (0 === $returnVar) {
+                return implode(PHP_EOL, $output);
+            }
+            return null;
+        };
+    }
+
+    public static function shouldLoadClientCertSource()
```
dwsupplee, line 282:
```suggestion
    /**
     * Determines whether or not the default device certificate should be loaded.
     *
     * @return bool
     */
    public static function shouldLoadClientCertSource()
```

src/CredentialsLoader.php
```diff
@@ -247,4 +250,52 @@ public function updateMetadata(
...
+    {
+        return 'true' === getenv(self::MTLS_CERT_ENV_VAR);
```
dwsupplee, line 284:
```suggestion
        return filter_var(getenv(self::MTLS_CERT_ENV_VAR), FILTER_VALIDATE_BOOLEAN);
```

src/CredentialsLoader.php
```diff
@@ -247,4 +250,52 @@ public function updateMetadata(
 
         return $metadata_copy;
     }
+
+    public static function getDefaultClientCertSource()
+    {
+        if (!$clientCertSourceJson = self::loadDefaultClientCertSourceFile()) {
+            return null;
+        }
+        if (!isset($clientCertSourceJson['cert_provider_command'])) {
```
dwsupplee, line 259: WDYT about putting this validation in the body of `loadDefaultClientCertSourceFile`?

tests/CredentialsLoaderTest.php
```diff
@@ -29,6 +29,123 @@ public function testUpdateMetadataSkipsWhenAuthenticationisSet()
...
+    }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testActualDefaultClientCertSource()
+    {
+        $clientCertSource = CredentialsLoader::getDefaultClientCertSource();
+        if (is_null($clientCertSource)) {
+            $this->markTestSkipped('No client cert source found');
```
dwsupplee, line 97: What are the cases where this wouldn't be found? Should we stub it out?

tests/CredentialsLoaderTest.php
```diff
@@ -29,6 +29,123 @@ public function testUpdateMetadataSkipsWhenAuthenticationisSet()
...
+        $creds = $clientCertSource();
+        $this->assertTrue(is_string($creds));
+        $this->assertContains('-----BEGIN CERTIFICATE-----', $creds);
+        $this->assertContains('-----BEGIN PRIVATE KEY-----', $creds);
+    }
+
+    /**
+     * @runInSeparateProcess
+     */
+    public function testDefaultClientCertSourceInvalidCmdThrowsException()
```
dwsupplee, line 108: Not sure if at some point the original intent was to throw an exception, but I do like that approach. If we keep as is, we should update the method here.

```suggestion
    public function testDefaultClientCertSourceInvalidCmdReturnsNull()
```

dwsupplee (changes requested):

tests/CredentialsLoaderTest.php
bshaffer: The  client cert source is only found when the corresponding file exists (`~/.secureConnect/context_aware_metadata.json`). This is tested in `testNonExistantDefaultClientCertSource`!

tests/CredentialsLoaderTest.php
bshaffer: I agree throwing an exception is better. I can't remember if this was to ensure we don't break customers, or to be consistent with other languages, but I will update it to throw an exception as that's a clearer UX.

src/CredentialsLoader.php
bshaffer: done

tests/CredentialsLoaderTest.php
bshaffer: Updated to throw an exception if the command errors

dwsupplee (approved):

PR #357, opened 2021-08-24
feat: Adding support for guzzlehttp/psr7:^2.0

saranshdhingra:
Replaced the use of deprecated functions in **guzzlehttp/psr7** for googleapis/google-cloud-php#4247

This is done so that we can use 1.x and 2.x branches of the **guzzlehttp/psr7** package.

The version in composer.json has been bumped to a **minimum of 1.7** because the alternatives to the deprecated versions are available starting **1.7**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,3 @@
-## 1.17.1 (08/25/2021)
-
-* [chore]: Replaced the use of deprecated functions in Guzzlehttp/Psr7 and bumbed the version from **^1.2** to **^1.7|^2.0**
-
 ## 1.17.0 (08/17/2021)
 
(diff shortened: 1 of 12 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.17.1 (08/25/2021)
+
+* [chore]: Replaced the use of deprecated functions in Guzzlehttp/Psr7 and bumbed the version from **^1.2** to **^1.7|^2.0**
+
```
bshaffer, line 4: Please remove this change, as this will be added in a separate release PR. Also this is more of a `feat` than a bug.

bshaffer (commented):
Just one request, otherwise this looks great

bshaffer (approved):

dwsupplee:
Thanks @saranshdhingra! Again, great find on 1.7+ supporting the Utils class as well.

Issue #363, opened 2021-09-23
Library incompatible with psr/cache ^3.0


Geolim4:
Hello,

I'm running a project ([Phpfastcache#v9](https://github.com/PHPSocialNetwork/phpfastcache/tree/v9)) that requires `psr/cache ^3.0`, which conflict with your constraints: `^1.0|^2.0`

So actually I cannot make use of any Google product at all because of this very "central" constraint.

Do you plan to make it compatible with `psr/cache ^3.0` ?

It's basically the same interface as previous but with PHP-8 type hint.

This should not break compatibility at all to support it since `psr/cache ^3.0`  already constraints on `php ^8.0` and the interface remains the same.

Thanks.

#### Environment details

  - OS: Win 10 x64
  - PHP version: 8.0
  - Package name and version: latest

Geolim4:
@bshaffer this issue will become more and more blocking over the months coming with the PHP-8 popularity growing up :)

bshaffer:
Hi @Geolim4 ! yes you're right, and it's a concern. But correct me if I'm wrong, every project supporting `3.0` _should_ be supporting `2.0`, and so as long as all the projects which support `1.0` also support `2.0`, the problem should be resolved.

Upgrading to `3.0` would break BC and require a major version bump, which we may do if this issue becomes an issue for other auth library consumers.

Geolim4:
> Upgrading to `3.0` would break BC and require a major version bump, which we may do if this issue becomes an issue for other auth library consumers.

If you add the `|| ^3.0` the library _should not_ have a BC break (or I don't see it :P):

If you project make use of `google/auth` as a dependecy:

- Without `psr/cache ^3.0` support (Current behavior)
  - You use php < 8.0:  no problem, composer will stay on a PHP-compatible version of `psr/cache`: `1.0.1`
  - You use php >= 8.0: no problem, composer will select a PHP-compatible version of `psr/cache`: `1.0.1` or `2.0.0`
- With `psr/cache ^3.0` support (new behavior)
  - You use php < 8.0: , no problem, composer will stay on a PHP-compatible version of `psr/cache`: `1.0.1` or `2.0.0` (`3.0.0`  is indeed incompatible, but not blocking for Composer)
[...]

PR #364, opened 2021-09-24
feat: add TypedItem to allow for psr/cache:3

bshaffer:
Supports `psr/cache:3.0` and addresses #363 

This PR raises the minimum version supported to PHP 7.1 and adds return types to `SysvCacheItemPool` and `MemoryCacheItemPool`

This change will break BC for anyone extending `SysvCacheItemPool`, but I imagine this is a very unlikely scenario.


tests/BaseTest.php
```diff
@@ -55,4 +55,13 @@ public function getValidKeyName($key)
     {
         return preg_replace('|[^a-zA-Z0-9_\.! ]|', '', $key);
     }
+
+    protected function getCacheItemClass()
+    {
+        if (\PHP_VERSION_ID >= 80000) {
+            return 'Google\Auth\Cache\TypedItem';
```
Geolim4, line 62: You can use this syntax (even if the class does not exists)
```suggestion
            return \Google\Auth\Cache\TypedItem::class;
```

(You can do it with a class import too).

src/Cache/TypedItem.php
```diff
@@ -0,0 +1,187 @@
...
+        // We test for two types here due to the fact the DateTimeInterface
+        // was not introduced until PHP 5.5. Checking for the DateTime type as
+        // well allows us to support 5.4.
+        if ($expiration instanceof \DateTimeInterface) {
+            return true;
+        }
+
+        if ($expiration instanceof \DateTime) {
+            return true;
+        }
```
Geolim4, line 178:
```suggestion
        if ($expiration instanceof \DateTimeInterface || $expiration instanceof \DateTime) {
            return true;
        }
```

src/Cache/TypedItem.php
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Handles an error.
+     *
+     * @param string $error
+     * @throws \TypeError
+     */
+    private function handleError($error)
+    {
+        if (class_exists('TypeError')) {
```
Geolim4, line 150:
```suggestion
        if (class_exists(\TypeError::class)) {
```

src/Cache/TypedItem.php
```diff
@@ -0,0 +1,187 @@
...
+            throw new \TypeError($error);
+        }
+
+        trigger_error($error, E_USER_ERROR);
```
Geolim4, line 154:
```suggestion
        trigger_error($error, \E_USER_ERROR);
```

tests/BaseTest.php
bshaffer: I can't because this library currently supports PHP 5.4, which does not support this syntax.

Geolim4: Ohhh my bad, I didn't paid attention to such level backward compatibility.

My apologies.

src/Cache/MemoryCacheItemPool.php
```diff
@@ -55,12 +55,18 @@ public function getItem($key)
...
-            $items[$key] = $this->hasItem($key) ? clone $this->items[$key] : new Item($key);
+        if (\PHP_VERSION_ID >= 80000) {
+            foreach ($keys as $key) {
+                $items[$key] = $this->hasItem($key) ? clone $this->items[$key] : new TypedItem($key);
+            }
+        } else {
+            foreach ($keys as $key) {
+                $items[$key] = $this->hasItem($key) ? clone $this->items[$key] : new Item($key);
+            }
         }
```
noahdietz, line 64: I like how explicit it is with two distinct branches, but could it be collapsed into one loop with a variable "false" condition type?

```
$default = \PHP_VERSION_ID >= 80000 ? TypedItem:class : Item:class;
for ($keys as $key) {
  $items[$key] = $this->hasItem($key) ? clone $this->items[$key] : new $default($key);
}
```

src/Cache/SysVCacheItemPool.php
```diff
@@ -99,22 +99,30 @@ public function getItem($key)
...
+                $items[$key] = $this->hasItem($key) ?
+                    clone $this->items[$key] :
+                    new TypedItem($key);
+            }
+        } else {
+            foreach ($keys as $key) {
+                $items[$key] = $this->hasItem($key) ?
+                    clone $this->items[$key] :
+                    new Item($key);
+            }
```
noahdietz, line 117: Same comment here as above.

noahdietz (commented):
_Reviewed from my phone lol_

src/Cache/SysVCacheItemPool.php
bshaffer: Done!

src/Cache/MemoryCacheItemPool.php
bshaffer: Done!

src/Cache/MemoryCacheItemPool.php
```diff
@@ -55,12 +55,12 @@ public function getItem($key)
      *   key is not found. However, if no keys are specified then an empty
      *   traversable MUST be returned instead.
      */
-    public function getItems(array $keys = [])
+    public function getItems(array $keys = []): iterable
```
noahdietz, line 58: Does the PHP Doc `@return` need to be updated? I'm not sure if there is a difference between `array` and `iterable` here.

noahdietz (approved):
I will Approve, but perhaps double check with @dwsupplee that the BC is OK? Feel like I'm still learning a bit in these repos.

src/Cache/TypedItem.php
```diff
@@ -0,0 +1,187 @@
+<?php
+/*
+ * Copyright 2016 Google Inc.
```
dwsupplee, line 3:
```suggestion
 * Copyright 2022 Google Inc.
```

src/Cache/TypedItem.php
```diff
@@ -0,0 +1,187 @@
...
+ */
+
+namespace Google\Auth\Cache;
+
+use Psr\Cache\CacheItemInterface;
+
+/**
+ * A cache item.
+ */
+final class TypedItem implements CacheItemInterface
```
dwsupplee, line 25: It looks like the property declaration for `key` is missing.

dwsupplee (changes requested):
I agree the surface for the BC break is small enough to warrant the support here 👍 , LGTM with a few minor notes.

src/Cache/TypedItem.php
Lustmored: I see it is defined in the constructor (using property promotion available since PHP 8.0) 👍

dwsupplee: Good call 👍

.github/workflows/tests.yml
```diff
@@ -23,16 +23,13 @@ jobs:
...
-              name: Run PHPUnit Patches
-              run: sh .github/apply-phpunit-patches.sh
             - name: Run Script
               run: vendor/bin/phpunit
     test_lowest:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php: [ "5.6", "7.2" ]
+                php: [ "7.1", "7.2" ]
```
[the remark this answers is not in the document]
bshaffer: done!

src/Cache/TypedItem.php
bshaffer: done!

bshaffer:
@dwsupplee all done here!

dwsupplee (approved):

Geolim4:
Alleluia 🥳 

Thank you to all participants.

bshaffer:
This is now released in [v1.20.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.20.0)

Issue #371, opened 2022-01-30
Use FQCN for param/return types


kkmuffme:
Use FQCN for param/return types to avoid tons of false positive errors in static analysis when using this library

bshaffer:
A better approach would be to add `phpstan` static analyzer.

PR #372, opened 2022-01-30
use FQCN for oAuth

kkmuffme:
Partial fix for https://github.com/googleapis/google-auth-library-php/issues/371 for at least oAuth


bshaffer:
I'm pretty sure the syntax here is correct. Since we import the `UriInterface` at the top of the class, static analyzers know which `UriInterface` is being referred here (otherwise, there would be a conflict in the code).

What you've provided would need to be prefixed with a `\` in order to be correct, as what you have here would be in the namespace of the file, and so would actually be `Google\Auth\Psr\Http\Message\UriInterface`.

Thanks for your contribution, but what we have here is correct.

PR #375, opened 2022-02-12
chore: add explicit return types to address Symfony deprecations

rlcurrall:
This is meant to address a similar issue to what was done in this merge request:
https://github.com/googleapis/google-api-php-client/pull/2200

The deprecation checking functionality explicitly ignores the use of `inheritDoc`
so we need to add the return types explicitly here. Reference:
https://github.com/symfony/symfony/issues/44802#issuecomment-1001776200


rlcurrall:
Hey @bshaffer,

Could you take a look at this PR? Thanks!

bshaffer (approved):

PR #383, opened 2022-03-24
chore: raise minimum PHP version to 7.0

bshaffer:


bshaffer:
Closing because we can just do this in https://github.com/googleapis/google-auth-library-php/pull/364

PR #388, opened 2022-03-29
Upgrade firebase/php-jwt to ^6.0

prufrock:
The firebase/php-jwt package recently had to make breaking changes to resolve a security flaw in their library. This change upgrades that library and fixes the code that broke with the upgrade.

Most notably it changes OAuth2::verifyIdToken to only support a single algorithm to bring it inline with the way Jwt::decode works.

You can find more information about the fixes to the firebase/php-jwt here:
https://github.com/firebase/php-jwt/releases/tag/v6.0.0
https://github.com/firebase/php-jwt/issues/351
https://security.snyk.io/vuln/SNYK-PHP-FIREBASEPHPJWT-2434829

I also saw there was a [renovote bot PR](https://github.com/googleapis/google-auth-library-php/pull/386) for this but it didn't fix the code so I thought this might be an opportunity to lend a hand.

Hope this helps! Let me know if there's something else I can do to get this merged. Thanks! 😄

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -18,6 +18,7 @@
 namespace Google\Auth;
 
 use Firebase\JWT\JWT;
+use Firebase\JWT\Key;
 use Google\Auth\HttpHandler\HttpClientCache;
 use Google\Auth\HttpHandler\HttpHandlerFactory;
 use GuzzleHttp\Psr7\Query;
@@ -380,7 +381,7 @@ class OAuth2 implements FetchAuthTokenInterface
      * `\InvalidArgumentException`.
      *
      * @param string $publicKey The public key to use to authenticate the token
-     * @param array $allowed_algs List of supported verification algorithms
+     * @param string $allowed_alg The supported verification algorithm
      * @throws \DomainException if the token is missing an audience.
      * @throws \DomainException if the audience does not match the one set in
      *         the OAuth2 class instance.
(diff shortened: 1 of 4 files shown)


prufrock:
Working on getting the CLA figured out.

prufrock:
I think I got the CLA thing figured out 😄

src/OAuth2.php
```diff
@@ -1377,12 +1378,12 @@ private function coerceUri($uri)
     /**
      * @param string $idToken
      * @param string|array|null $publicKey
-     * @param array $allowedAlgs
+     * @param string $allowedAlg
      * @return object
      */
-    private function jwtDecode($idToken, $publicKey, $allowedAlgs)
+    private function jwtDecode($idToken, $publicKey, $allowedAlg)
```
bshaffer, line 1384: unfortunately this is a backwards breaking change.

bshaffer (commented):
I worked on a way to implement this in https://github.com/googleapis/google-auth-library-php/pull/391, and I can't find a way to implement it without breaking backwards compatibility.

I think it would be better to break these edge case implementations for security reasons, however.

bshaffer:
Closing in favor of https://github.com/googleapis/google-auth-library-php/pull/391, which is a little more robust in handling the BC breaking issues with this change

src/OAuth2.php
prufrock: I thought that might be the case. Thanks for the feedback and for the work done in #391. Ultimately, I just need the issue fixed 😄

PR #391, opened 2022-04-04
feat!: allow firebase/php-jwt v6

bshaffer:
see CVE-2021-46743 and https://github.com/advisories/GHSA-8xf4-w7qw-pjjw


ahthomsen:
Bump - would be great to get merged in to enable meeting security SLAs!

bshaffer:
@ahthomsen for the record, firebase PHP-JWT offers a fix for this in [v5.5](https://github.com/firebase/php-jwt/releases/tag/v5.5.0), which this library supports.

src/OAuth2.php
```diff
@@ -1376,23 +1378,60 @@ private function coerceUri($uri)
 
     /**
      * @param string $idToken
-     * @param string|array|null $publicKey
-     * @param array $allowedAlgs
+     * @param Key|Key[]|string|array|null $publicKey
+     * @param string|array $allowedAlg
```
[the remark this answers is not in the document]
vishwarajanand:
```suggestion
     * @param string|array $allowedAlgos
```

src/OAuth2.php
```diff
@@ -1376,23 +1378,60 @@ private function coerceUri($uri)
...
      * @return object
      */
     private function jwtDecode($idToken, $publicKey, $allowedAlgs)
     {
-        return JWT::decode($idToken, $publicKey, $allowedAlgs);
+        $keys = $this->getFirebaseJwtKeys($publicKey, $allowedAlgs);
+
+        // Default exception if none are caught
```
vishwarajanand, line 1389:
```suggestion
        // Default exception if all are caught
```

src/OAuth2.php
```diff
@@ -1376,23 +1378,60 @@ private function coerceUri($uri)
...
-            $signingKey,
-            $signingAlgorithm,
-            $signingKeyId
-        );
+        // If $allowedAlgs is empty, assume $publicKey is Key or Key[].
+        if (empty($allowedAlgs)) {
+            return (array) $publicKey;
+        }
+
+        if (is_string($allowedAlgs)) {
```
vishwarajanand, line 1411: I think `$allowedAlgs` check could instead check against all the supported algos mentioned [here](https://github.com/firebase/php-jwt/blob/main/src/JWT.php#L56-L66).

```
    public static $supported_algs = [
        'ES384' => ['openssl', 'SHA384'],
        'ES256' => ['openssl', 'SHA256'],
        'HS256' => ['hash_hmac', 'SHA256'],
        'HS384' => ['hash_hmac', 'SHA384'],
        'HS512' => ['hash_hmac', 'SHA512'],
        'RS256' => ['openssl', 'SHA256'],
        'RS384' => ['openssl', 'SHA384'],
        'RS512' => ['openssl', 'SHA512'],
        'EdDSA' => ['sodium_crypto', 'EdDSA'],
    ];
```

vishwarajanand (changes requested):

src/OAuth2.php
bshaffer: Actually I think what i have here is better. The reason is that the API exception being thrown is the default to preserve backwards compatibility. The validation doesn't happen here, but we pretend like it does for BC.

src/OAuth2.php
```diff
@@ -1376,23 +1378,60 @@ private function coerceUri($uri)
...
+     * @param string|array $allowedAlg
      * @return object
      */
     private function jwtDecode($idToken, $publicKey, $allowedAlgs)
     {
-        return JWT::decode($idToken, $publicKey, $allowedAlgs);
+        $keys = $this->getFirebaseJwtKeys($publicKey, $allowedAlgs);
+
+        // Default exception if none are caught
+        $e = new \InvalidArgumentException('Key may not be empty');
```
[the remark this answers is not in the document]
bshaffer: It's the same here .. this exception class and message were taken from firebase/php-jwt. I agree the exception and message don't make sense here, but they are here for backwards compatibility. I will add a comment to make this more clear.

src/OAuth2.php
bshaffer: This check is already done in `firebase/php-jwt`, so there is no need to do that check here... we only want to verify type-safety and make sure we handle the types correctly before calling `Firebase\JWT\JWT::decode`.

src/OAuth2.php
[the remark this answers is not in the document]
bshaffer: Comment added!

bshaffer:
@dwsupplee @vishwarajanand comments addressed! PTAL!

vishwarajanand (approved):
LGTM but please have @dwsupplee approve as well.

dwsupplee (approved):
Looks great! Thanks for the tests 👍

bshaffer:
This has been released in [v1.21.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.21.0)

Issue #399, opened 2022-04-30
AccessToken class incompatible with firebase/php-jwt 6.x


marickvantuil:
#### Environment details

  - OS:
  - PHP version: 8.1
  - Package name and version: google/auth v1.21.0

#### Steps to reproduce

1. Verify a legitimate token with google/auth v1.21.0 and firebase/php-jwt 6.x (v6.1.2)

#### Code example

```php
(new AccessToken())->verify(
    $token,
    [
        'audience' => $audience,
        'throwException' => true,
    ]
);
```

The AccessToken class decodes the token the '5.x way' instead of the '6.x way' (which now uses a Key class). I noticed [in the PR that adds firebase/php-jwt v6](https://github.com/googleapis/google-auth-library-php/pull/391) the OAuth2 class was changed, but the AccessToken class was skipped. Or is that because the AccessToken class is experimental?

bshaffer:
Thanks for bringing this up. I'll try to get a fix in quickly!

bshaffer:
fixed in #402

PR #402, opened 2022-05-13
fix: firebase php-jwt decode

bshaffer:
addresses https://github.com/googleapis/google-auth-library-php/issues/399

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -21,6 +21,7 @@ use DateTime;
 use Exception;
 use Firebase\JWT\ExpiredException;
 use Firebase\JWT\JWT;
+use Firebase\JWT\Key;
 use Firebase\JWT\SignatureInvalidException;
 use Google\Auth\Cache\MemoryCacheItemPool;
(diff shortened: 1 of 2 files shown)


src/AccessToken.php
```diff
@@ -257,13 +258,12 @@ private function verifyRs256($token, array $certs, $audience = null, $issuer = n
             ]);
 
             // create an array of key IDs to certs for the JWT library
-            $keys[$cert['kid']] =  $rsa->getPublicKey();
+            $keys[$cert['kid']] = new Key($rsa->getPublicKey(), 'RS256');
         }
 
         $payload = $this->callJwtStatic('decode', [
             $token,
             $keys,
```
dwsupplee, line 266:
```suggestion
            $keys
```

Just a minor thing, don't feel pressured to accept as I understand this can go either way.

dwsupplee (approved):

src/AccessToken.php
bshaffer: definitely a preference thing. I think I lean more towards leaving trailing commas on lists of array items which span multiple lines

PR #405, opened 2022-06-09
fix: `CredentialsLoader` should support `id_token`

Jeanno:
Currently, the following code doesn't work because the `public function updateMetadata` implementation in `CredentialsLoader` only checks for `access_token` but not `id_token`.
```
  $target_audience = 'https://some-endpoint.a.run.app/';  // A service deployed on Google Cloud Run
  $auth_credentials = ApplicationDefaultCredentials::getIdTokenCredentials($target_audience);
  $updated = $auth_credentials->updateMetadata([]); // `$updated` is still empty
```

The proposing CL adds handling for `id_token` which support the above case.

### DIFF
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -212,13 +212,17 @@ abstract class CredentialsLoader implements
             return $metadata;
         }
         $result = $this->fetchAuthToken($httpHandler);
-        if (!isset($result['access_token'])) {
+        if (isset($result['access_token'])) {
+            $metadata_copy = $metadata;
+            $metadata_copy[self::AUTH_METADATA_KEY] = ['Bearer ' . $result['access_token']];
+            return $metadata_copy;
+        } else if (isset($result['id_token'])) {
+            $metadata_copy = $metadata;
+            $metadata_copy[self::AUTH_METADATA_KEY] = ['Bearer ' . $result['id_token']];
+            return $metadata_copy;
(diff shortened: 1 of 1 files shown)


src/CredentialsLoader.php
```diff
@@ -212,13 +212,17 @@ public function updateMetadata(
             return $metadata;
         }
         $result = $this->fetchAuthToken($httpHandler);
-        if (!isset($result['access_token'])) {
+        if (isset($result['access_token'])) {
+            $metadata_copy = $metadata;
```
vishwarajanand, line 216: move this out of if block

src/CredentialsLoader.php
```diff
@@ -212,13 +212,17 @@ public function updateMetadata(
...
+            $metadata_copy[self::AUTH_METADATA_KEY] = ['Bearer ' . $result['access_token']];
+            return $metadata_copy;
+        } elseif (isset($result['id_token'])) {
+            $metadata_copy = $metadata;
+            $metadata_copy[self::AUTH_METADATA_KEY] = ['Bearer ' . $result['id_token']];
+            return $metadata_copy;
+        } else {
             return $metadata;
         }
```
vishwarajanand, line 225: No need of having the else block

Jeanno:
@vishwarajanand Updated according to your comments

bshaffer (approved):

PR #411, opened 2022-08-24
fix: add eager refresh

bshaffer:


dwsupplee (approved):
LGTM! Is there anything we need to worry about given GAX has this logic embedded as well, or are we good?

bshaffer:
As @dwsupplee pointed out, the lifetime is 30 minutes while an access token expires after 60. I still think it's worthwhile to have this check, as the lifetime can be configured differently and the access token could potentially be stored into cache 30 minutes after being fetched. There are both cases on the very edge, but nevertheless I think it's a best practice to have an eager threshold !

Issue #412, opened 2022-08-31
Wrong version in composer


wotekPL:
https://github.com/googleapis/google-auth-library-php/blob/deab93524279124917707189eef0038aa2bfd2a5/src/Cache/TypedItem.php#L50 this code is OK only in php8 but in composer you allow use PHP 7.1

vishwarajanand:
Hi @wotekPL , thanks for reporting this. I have raised a potential fix, will take it up internally.

bshaffer:
This class is specifically loaded only for PHP 8.0. [See here](https://github.com/googleapis/google-auth-library-php/blob/main/src/Cache/MemoryCacheItemPool.php#L60). This is in order to satisfy [`psr/cache:3`](https://github.com/php-fig/cache/blob/3.0.0/composer.json#L13), which is only available on PHP8+. So this is WAD.

PR #413, opened 2022-09-03
fix: remove Constructor Property Promotion syntax

vishwarajanand:
Addresses: https://github.com/googleapis/google-auth-library-php/issues/412


bshaffer:
This class is specifically loaded only for PHP 8.0. [See here](https://github.com/googleapis/google-auth-library-php/blob/main/src/Cache/MemoryCacheItemPool.php#L60). This is in order to satisfy [`psr/cache:3`](https://github.com/php-fig/cache/blob/3.0.0/composer.json#L13), which is only available on PHP8+. So this is WAD.

PR #415, opened 2022-09-14
feat: double default truncateAt for guzzle error output

bshaffer:
**_Guzzle 7 Only_**

This will help the user receive a more helpful error message when encountering Google API exceptions. For example, with the "reAuth" error:

**Before**
```
Fatal error: Uncaught GuzzleHttp\Exception\ClientException: Client error: `POST https://oauth2.googleapis.com/token` resulted in a `400 Bad Request` response:
{
  "error": "invalid_grant",
  "error_description": "reauth related error (rapt_required)",
  "error_uri": "https://sup (truncated...)
 in /path/to/project/vendor/guzzlehttp/guzzle/src/Exception/RequestException.php:113
```

**After**
```
Fatal error: Uncaught GuzzleHttp\Exception\ClientException: Client error: `POST https://oauth2.googleapis.com/token` resulted in a `400 Bad Request` response:
{
  "error": "invalid_grant",
  "error_description": "reauth related error (rapt_required)",
  "error_uri": "https://support.google.com/a/answer/9368756",
  "error_subtype": "rapt_required"
}
 in /path/to/project/vendor/guzzlehttp/guzzle/src/Exception/RequestException.php:113
```

### DIFF
--- a/src/HttpHandler/HttpHandlerFactory.php
+++ b/src/HttpHandler/HttpHandlerFactory.php
@@ -39,7 +39,8 @@ class HttpHandlerFactory
                 // double the # of characters before truncation by default
                 $bodySummarizer = new BodySummarizer(240);
                 $stack = HandlerStack::create();
-                $stack->push(Middleware::httpErrors($bodySummarizer), 'http_errors');
+                $stack->remove('http_errors');
+                $stack->unshift(Middleware::httpErrors($bodySummarizer), 'http_errors');
             }
             $client = new Client(['handler' => $stack]);
         }
--- a/src/HttpHandler/HttpHandlerFactory.php
+++ b/src/HttpHandler/HttpHandlerFactory.php
@@ -18,6 +18,9 @@ namespace Google\Auth\HttpHandler;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\ClientInterface;
+use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Middleware;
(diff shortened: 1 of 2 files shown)


src/HttpHandler/HttpHandlerFactory.php
```diff
@@ -30,7 +33,17 @@ class HttpHandlerFactory
      */
     public static function build(ClientInterface $client = null)
     {
-        $client = $client ?: new Client();
+        if (is_null($client)) {
+            $stack = null;
+            if (class_exists(BodySummarizer::class)) {
+                // double the # of characters before truncation by default
+                $bodySummarizer = new BodySummarizer(240);
```
noahdietz, line 40: Should this be made configurable somehow? Is there already a way to configure things in this package?

noahdietz (approved):

src/HttpHandler/HttpHandlerFactory.php
bshaffer: It's possible for the end user to pass in custom guzzle clients like this:
```php
$http = new GuzzleHttp\Client($someCustomConfig);
$assetClient = new AssetServiceGapicClient([
    'credentialsConfig' => [
        'authHttpHandler' => $http, // for calls made to the token endpoint
    ],
    'rest' => [
        'httpHandler' => $http,     // for calls made to the API using REST
    ]
]);
```

However, this change is to modify the default value, so we're just hardcoding it.

PR #422, opened 2022-10-21
fix: do not call GCECredentials::onGCE if ADC has already checked

bshaffer:
Adds a method `GCECredentials::setIsOnGce`, so that if `ApplicationDefaultCredentials` already checks that the library is running on GCE, we do not need to check again in `GCECredentials`.

```php
if (ApplicationDefaultCredentials::onGce()) {
    $creds = new GCECredentials();
    $creds->setIsOnGce(true); // save the credentials a trip to the metadata server
}
```

### DIFF
--- a/tests/Credentials/GCECredentialsTest.php
+++ b/tests/Credentials/GCECredentialsTest.php
@@ -20,9 +20,12 @@ namespace Google\Auth\Tests\Credentials;
 use Exception;
 use Google\Auth\Credentials\GCECredentials;
 use Google\Auth\HttpHandler\HttpClientCache;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
 use Google\Auth\Tests\BaseTest;
 use GuzzleHttp\Psr7;
 use GuzzleHttp\Psr7\Utils;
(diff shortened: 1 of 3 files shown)


tests/Credentials/GCECredentialsTest.php
```diff
@@ -375,6 +378,26 @@ public function testGetTokenUriWithServiceAccountIdentity()
...
+    }
+
+    public function testSetIsOnGceToTrueWhenNotOnGceThrowsException()
+    {
+        if (GCECredentials::onGCE()) {
+            $this->markTestSkipped('Must not be on GCE for this test');
+        }
+        $this->expectException(ConnectException::class);
+
+        $httpHandler = HttpHandlerFactory::build(new Client(['timeout' => 0.1]));
```
dwsupplee, line 395: Would you be open to mocking out the network call instead of using a real client?

dwsupplee (changes requested):

dwsupplee (approved):
LGTM, just also raising up the PHPStan failure if you'd like to address that.

PR #429, opened 2023-01-16
Replaced getenv/putenv with `$_ENV` and unset

DracoBlue:
The current version of google auth library php uses getenv/putenv to read and set environment variables in php. But the  function putenv is not threadsafe, which made libraries (Dotenv https://github.com/symfony/dotenv/blob/6.2/Dotenv.php#L61 and phpdotenv https://github.com/vlucas/phpdotenv#putenv-and-getenv) to disable putenv support by default. Frameworks like symfony (deprecated in 4.3 https://github.com/symfony/symfony/pull/31062 and removed in 5.0) and also in laravel there was a similiar solution to not use it anymore directly (https://github.com/vlucas/phpdotenv/issues/76).

Thus here is a pull request to change all the ocurences of:

- `getenv($variable)` to `array_key_exists($variable, $_ENV) ? $_ENV[$variable] : false`
- `putenv($variable)` to `unset($_ENV[$variable])`
- `putenv($variable . "=")` to `$_ENV[$variable]=''`
- `putenv($variable . "=" . $value)` to `$_ENV[$variable]=$value`

Threads about this:

- https://bugs.php.net/bug.php?id=74143 (putenv does not work if fast cgi sets the variable!)
- https://github.com/symfony/symfony/pull/31062
- https://github.com/vlucas/phpdotenv/issues/76
- https://github.com/laravel/framework/issues/7354 
- https://github.com/laravel/framework/commit/866a8cfc58430830069bd24feacf2d81e99385a2
- https://github.com/silverstripe/silverstripe-framework/pull/7484
- https://github.com/laravel/framework/issues/27949

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,10 +59,10 @@ Once downloaded, store the path to this file in the
 `GOOGLE_APPLICATION_CREDENTIALS` environment variable.
 
 ```php
-putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
+$_ENV['GOOGLE_APPLICATION_CREDENTIALS'] = '/path/to/my/credentials.json';
 ```
 
-> PHP's `putenv` function is just one way to set an environment variable.
+> PHP's `$_ENV` variable is just one way to set an environment variable.
 > Consider using `.htaccess` or apache configuration files as well.
 
 #### Enable the API you want to use
@@ -83,7 +83,7 @@ use GuzzleHttp\Client;
 use GuzzleHttp\HandlerStack;
 
 // specify the path to your application credentials
-putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
+$_ENV['GOOGLE_APPLICATION_CREDENTIALS'] = '/path/to/my/credentials.json';
 
 // define the scopes for your API call
 $scopes = ['https://www.googleapis.com/auth/drive.readonly'];
@@ -136,7 +136,7 @@ use GuzzleHttp\Client;
 use GuzzleHttp\HandlerStack;
 
 // specify the path to your application credentials
-putenv('GOOGLE_APPLICATION_CREDENTIALS=/path/to/my/credentials.json');
+$_ENV['GOOGLE_APPLICATION_CREDENTIALS'] = '/path/to/my/credentials.json';
 
 // Provide the ID token audience. This can be a Client ID associated with an IAP application,
 // Or the URL associated with a CloudRun App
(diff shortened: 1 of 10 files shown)


vishwarajanand:
@DracoBlue thanks for making this PR. I want to check once whether there was any reason for using `getenv/putenv` across all libraries.
If you can, pls fix the commit message according to conventional commit guidelines.

DracoBlue:
Done!

bshaffer (changes requested):
@DracoBlue Sorry for the confusing bot message, your previous PR title was correct, and it was just giving a warning to us (the administrators) in case we merged with the _commit message_ instead of the PR title, which was in the incorrect format. I fixed this in #430, so this message will no longer happen.

As for this PR, unfortunately, the changes here break existing implementations. If someone is setting their credentials using `putenv` (as we've instructed them to do), this library would no longer work as expected. This is unacceptable. 
I can see three options:

 1. Wait to make these changes in a v2 version
 2. Check `getenv` first and `_ENV` second. (This seems like it would cause even more issues)
 3. Provide a way (similar to symphony's `usePutenv` function, perhaps even use an environment variable as a flag 😛) to use the threadsafe version, e.g.:

```php
$_ENV['GOOGLE_AUTH_USE_PUTENV']=false
$_ENV['GOOGLE_APPLICATION_CREDENTIALS'] = '/path/to/credentials.json'';
```

There are some other implications of switching from `getenv`/`putenv` that aren't discussed in this PR, along with the implications of these functions not being threadsafe. Do you have any links to discussions of the implications?

DracoBlue:
> There are some other implications of switching from getenv/putenv that aren't discussed in this PR, along with the implications of these functions not being threadsafe. Do you have any links to discussions of the implications?

I added some more threads related to this to the initial post.

> Check getenv first and _ENV second. (This seems like it would cause even more issues)

I think so, too.

> Provide a way (similar to symphony's usePutenv function, perhaps even use an environment variable as a flag 😛) to use the threadsafe version, e.g.:

I guess for this to be interesting enough to implement, I would suggest I am supplying a test case (I don't have a proper one, yet!) which suffers from the getenv/putenv threadsafety issue - and then we can think about a way to make it possible for those using the frameworks possible to supply the credentials variable in a safe way.

bshaffer:
I've moved this into a feature request here: https://github.com/googleapis/google-auth-library-php/issues/432
> supplying a test case... which suffers from the getenv/putenv threadsafety issue

Yes! If you have one, or find one, please update the above issue!

I'm closing this for now since we will not be using this implementation.

PR #434, opened 2023-01-26
chore: fix docs generation for tag trigger

bshaffer:
Docs generation fails on tag triggers with the error that the `main` branch doesn't exist (see the [Generate Documentation actions runs](https://github.com/googleapis/google-auth-library-php/actions/workflows/docs.yml))

I cannot find a way to test this without tagging a release (which we don't want to do), so I may try to fork this repo to my personal account and try tagging there.


.github/actions/docs/entrypoint.sh
```diff
@@ -3,6 +3,7 @@
 apt-get update
 apt-get install -y git
 git reset --hard HEAD
+git fetch origin
```
vishwarajanand, line 6: Maybe its better to fetch origin before resetting to HEAD:
```suggestion
git fetch origin
git reset --hard HEAD
```

vishwarajanand (approved):

PR #441, opened 2023-03-16
feat: accessing granted scopes 

bshaffer:
b/253077796

Adds support for accessing granted scopes on the OAuth2 service class

**Q: Does this need to be accessible by the end user?**
 - Yes

**Q: If so, how will this be accessible by the end user?**
 - Exposing a method such as `getGrantedScopes` on the `Credentials` object is sufficient for this library.

**Q: How is this different from the existing `getScopes` method?**
 - It wouldn't be any different for the other credentials types, only for `OAuth2` and `UserRefreshCredentials`, because they get information back from the API

### DIFF
--- a/src/Credentials/UserRefreshCredentials.php
+++ b/src/Credentials/UserRefreshCredentials.php
@@ -139,4 +139,14 @@ class UserRefreshCredentials extends CredentialsLoader implements GetQuotaProjec
     {
         return $this->quotaProject;
     }
+
+    /**
+     * Get the granted scopes (if they exist) for the last fetched token.
+     *
+     * @return string|null
+     */
+    public function getGrantedScope()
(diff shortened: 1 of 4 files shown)


shivgautam (approved):

sai-sunder-s (approved):

TimurSadykov:
Sorry for the late comment, sorry if any confusion. Any reason why scopes were shared via credential, not the access token?

bshaffer:
Hi @TimurSadykov ! I'm sorry but i don't understand the question. If you mean why it isn't part of the class AccessToken, that's because that class is only a utility class used for validation and not used by the downstream client libraries during authentication. The credentials are the objects accessed by the downstream client libraries

PR #443, opened 2023-04-01
Proposing code to add partial consent flag

hsuharso:
Proposing code edit to add partial consent flag as an option in authorization URL to select the unbundled scope screen in OAuth2.

### DIFF
--- a/src/OAuth2.php
+++ b/src/OAuth2.php
@@ -269,6 +269,13 @@ class OAuth2 implements FetchAuthTokenInterface
     private $additionalClaims;
 
     /**
+     * Added flag to enable/disable partial consent screen
+     *
+     * @var ?string
+     */
+    private $partConsent;
+
+    /**
      * Create a new OAuthCredentials.
      *
      * The configuration array accepts various options
@@ -357,6 +364,7 @@ class OAuth2 implements FetchAuthTokenInterface
             'signingAlgorithm' => null,
             'scope' => null,
             'additionalClaims' => [],
+	    'partConsent' => '',
         ], $config);
 
         $this->setAuthorizationUri($opts['authorizationUri']);
@@ -377,6 +385,7 @@ class OAuth2 implements FetchAuthTokenInterface
         $this->setScope($opts['scope']);
         $this->setExtensionParams($opts['extensionParams']);
         $this->setAdditionalClaims($opts['additionalClaims']);
+        $this->setPartConsent($opts['partConsent']);
         $this->updateToken($opts);
     }
 
@@ -694,6 +703,7 @@ class OAuth2 implements FetchAuthTokenInterface
             'redirect_uri' => $this->redirectUri,
             'state' => $this->state,
             'scope' => $this->getScope(),
+            'enable_serial_consent' => $this->partConsent,
         ], $config);
 
         // Validate the auth_params
@@ -709,6 +719,11 @@ class OAuth2 implements FetchAuthTokenInterface
             throw new InvalidArgumentException(
                 'prompt and approval_prompt are mutually exclusive'
             );
+        }
+
+        // set default consent as old consent if not defined or has other values than 'true'
+        if (empty($params['enable_serial_consent'])||$params['enable_serial_consent'] != 'true') {
+            $params['enable_serial_consent'] = 'false';
         }
 
         // Construct the uri object; return it if it is valid.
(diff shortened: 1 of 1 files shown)


src/OAuth2.php
```diff
@@ -357,6 +364,7 @@ public function __construct(array $config)
             'signingAlgorithm' => null,
             'scope' => null,
             'additionalClaims' => [],
+	    'partConsent' => '',
```
bshaffer, line 367: whitespace is off here
Also, let's not use abbreviations. "partialConsent" is a perfectly fine variable name
```suggestion
	        'partialConsent' => '',
```

src/OAuth2.php
```diff
@@ -377,6 +385,7 @@ public function __construct(array $config)
         $this->setScope($opts['scope']);
         $this->setExtensionParams($opts['extensionParams']);
         $this->setAdditionalClaims($opts['additionalClaims']);
+        $this->setPartConsent($opts['partConsent']);
```
bshaffer, line 388:
```suggestion
        $this->setPartialConsent($opts['partialConsent']);
```

src/OAuth2.php
```diff
@@ -694,6 +703,7 @@ public function buildFullAuthorizationUri(array $config = [])
             'redirect_uri' => $this->redirectUri,
             'state' => $this->state,
             'scope' => $this->getScope(),
+            'enable_serial_consent' => $this->partConsent,
```
bshaffer, line 706: Just do the logic here that you have below. If this is a boolean (as it should be), it becomes very easy:
```php
'enable_serial_consent' => $this->partialConsent ? 'true' : 'false',
```

src/OAuth2.php
```diff
@@ -268,6 +268,13 @@ class OAuth2 implements FetchAuthTokenInterface
      */
     private $additionalClaims;
 
+    /**
+     * Added flag to enable/disable partial consent screen
+     *
+     * @var ?string
+     */
+    private $partConsent;
```
bshaffer, line 276: from the code, it seems like this is a boolean and not a string

src/OAuth2.php
```diff
@@ -1410,6 +1425,11 @@ public function getAdditionalClaims()
         return $this->additionalClaims;
     }
 
+    public function setPartConsent($partConsent)
+    {
+        $this->partConsent = $partConsent;
+    }
```
bshaffer, line 1431:
```suggestion
    /**
     * Sets whether to implement partial consent in the authorization request.
     *
     * return void
     */
    public function setPartialConsent(bool $partialConsent)
    {
        $this->partialConsent = $partialConsent;
    }
```

src/OAuth2.php
```diff
@@ -711,6 +721,11 @@ public function buildFullAuthorizationUri(array $config = [])
             );
         }
 
+        // set default consent as old consent if not defined or has other values than 'true'
+        if (empty($params['enable_serial_consent'])||$params['enable_serial_consent'] != 'true') {
+            $params['enable_serial_consent'] = 'false';
+        }
+
```
bshaffer, line 728: see suggestion above. This can be removed.

bshaffer (changes requested):
A bigger question I have on here is - are we sure we want to expose a parameter named "partial consent" to the end user? I want to confirm this is what the other languages are doing. It seems like it would make more sense to stick with the name of the _querystring parameter_ and just use `enable_serial_consent` as the configuration parameter. There's no other logic happening other than turning that parameter true or false, so why have two names for it?
cc @TimurSadykov

src/OAuth2.php
hsuharso: I will change it to "enable_serial_consent" as suggested

src/OAuth2.php
sai-sunder-s, line 276: The design is not to expose a parameter specifically for partial consent and add it to the auth url. Instead, we should be exposing a property bag with which the user can provide any custom key-value pairs to be added to the auth url. Checkout "Introduce a free-form query parameter-value option" section of go/custom-auth-params. It has a code sample for java as well.

sai-sunder-s (changes requested):

src/OAuth2.php
bshaffer: Hmm, is this the required implementation? It wouldn't be consistent with how we currently do things if we implemented it this way

bshaffer: Well, taking a look at our `buildFullAuthorizationUrl` method, it _already supports_ [additional custom querystring parameters](https://github.com/googleapis/google-auth-library-php/blob/main/src/OAuth2.php#L682) via the `$config` variable. So it's possible that this requirement is already satisfied.

bshaffer:
Closing this because we will be implementing it in a different way

src/OAuth2.php
sai-sunder-s: Thanks for pointing out the existing implementation for providing custom query parameters.

PR #445, opened 2023-04-05
chore: drop support for PHP 7.3 

bshaffer:


composer.json
```diff
@@ -9,20 +9,20 @@
     "docs": "https://googleapis.github.io/google-auth-library-php/main/"
   },
   "require": {
-    "php": "^7.1||^8.0",
-    "firebase/php-jwt": "^5.5||^6.0",
+    "php": "^7.4||^8.0",
+    "firebase/php-jwt": "^6.0",
     "guzzlehttp/guzzle": "^6.2.1|^7.0",
-    "guzzlehttp/psr7": "^1.7|^2.0",
+    "guzzlehttp/psr7": "^2.0",
```
vishwarajanand, line 15: Seems like older guzzle versions suffer from security vulnerabilities: https://packagist.org/packages/guzzlehttp/psr7/advisories?version=7023692
```suggestion
    "guzzlehttp/psr7": "^2.4.5",
```

composer.json
```diff
@@ -9,20 +9,20 @@
...
     "psr/http-message": "^1.0||^2.0",
```
vishwarajanand, line 16: [Optional] Seems [v1.0](https://packagist.org/packages/psr/http-message#1.0) supports php<7.4, so maybe we are better with 1.1
```suggestion
    "psr/http-message": "^1.1||^2.0",
```

tests/Credentials/ServiceAccountCredentialsTest.php
```diff
@@ -26,30 +26,31 @@
 use Google\Auth\CredentialsLoader;
 use Google\Auth\OAuth2;
 use GuzzleHttp\Psr7;
+use GuzzleHttp\Psr7\Response;
```
vishwarajanand, line 29: all imports here need to be sanitized. Some of them are not used at all hence leading to static check failures.

composer.json
bshaffer: good catch! updated.

composer.json
bshaffer: SGTM

vishwarajanand (approved):

PR #452, opened 2023-04-25
feat: implement quota project from env var in google/auth

bshaffer:


bshaffer:
Closing as this is better implemented in the GAX layer

tests/ApplicationDefaultCredentialsTest.php
```diff
@@ -590,6 +590,58 @@ public function testGetCredentialsUtilizesQuotaProjectInKeyFile()
         );
     }
 
+    /** @runInSeparateProcess */
+    public function testGetCredentialsUtilizesQuotaProjectEnvVar()
+    {
+        $quotaProject = 'quota-project-from-env-var';
+        putenv(ServiceAccountCredentials::QUOTA_PROJECT_ENV_VAR . '=' . $quotaProject);
```
sai-sunder-s, line 597: I am new to PHP, so could be a dumb qn. Why is the QUOTA_PROJECT_ENV_VAR constant used from ServiceAccountCredentials and not from CredentialsLoader?

sai-sunder-s (approved):

tests/ApplicationDefaultCredentialsTest.php
bshaffer: No reason, only because it's used like that elsewhere in the tests. I agree that from CredentialsLoader makes more sense

tests/ApplicationDefaultCredentialsTest.php
```diff
@@ -590,6 +591,58 @@ public function testGetCredentialsUtilizesQuotaProjectInKeyFile()
...
+            null  // $defaultScope
+        );
+
+        $this->assertEquals(
+            $quotaProject,
+            $credentials->getQuotaProject()
+        );
+    }
+
+    public function testGetCredentialsUtilizesQuotaProjectEnvVarOverKeyFile()
```
bshaffer, line 631:
```suggestion
    /** @runInSeparateProcess */
    public function testGetCredentialsUtilizesQuotaProjectEnvVarOverKeyFile()
```

tests/ApplicationDefaultCredentialsTest.php
bshaffer: changed!

sai-sunder-s (approved):

src/ApplicationDefaultCredentials.php
```diff
@@ -170,6 +170,11 @@ public static function getCredentials(
             $httpHandler = HttpHandlerFactory::build($client);
         }
 
+        if (is_null($quotaProject)) {
+            // if a quota project isn't specified, try to get one from the env var
+            $quotaProject = CredentialsLoader::quotaProjectFromEnv();
```
vishwarajanand, line 175: `quotaProject` isn't passed into `AppIdentityCredentials` on line # 188. Is that intentional?

vishwarajanand:
It appears that if `GOOGLE_CLOUD_QUOTA_PROJECT` is set then setting `$quotaProject` shouldn't be required for service clients. Questions:
1. We can't remove them from clients since they might be breaking? So, which one will be preferred?
2. In case both are passed with different values, should we add a doc block in clients to clarify this?

src/ApplicationDefaultCredentials.php
bshaffer: There are two reasons why we don't pass it into `AppIdentityCredentials`

 1. `AppIdentityCredentials` is deprecated, so we don't want to change it
 2. `AppIdentityCredentials` do not accept a quota project (because we no longer update it)

bshaffer:
> It appears that if `GOOGLE_CLOUD_QUOTA_PROJECT` is set then setting `$quotaProject` shouldn't be required for service clients. 

I do not think setting quotaProject is required anywhere.

> 1. We can't remove them from clients since they might be breaking? So, which one will be preferred?

Removing it would definitely be breaking. Also we don't want to remove it. There is no "preferred", it's just that in some situations a user might want to use an env var, and in some situations a parameter. 

> 2. In case both are passed with different values, should we add a doc block in clients to clarify this?

In the case that both are set, the parameter is used. I'm all for adding clarity on this, and will add documentation to reflect this.

ajupazhamayil (approved):

vishwarajanand (approved):
Lgtm

PR #456, opened 2023-05-16
FIX: Framework Symfony 5+ Compatibility

DamienLinux:
For security reasons, the use of putenv when loading environment variables declared in .env* is disabled on the Symfony framework.
A workaround method existed on Symfony 4 but:
- this did not solve the problem
- this is not applicable on Symfony 5.4 and above.

Thus for the recovery of the environment variable GOOGLE_APPLICATION_CREDENTIALS.
As this is associated with a security issue, to ensure clean compatibility with Symfony, I suggest:
- To always call getenv, needed for local-user/system environment variables (will not be usable for symfony)
- If getenv returns false, look directly at $_ENV to make sure it is not empty (this is where compatibility with Symfony is ensured)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *~
 vendor
-composer.lock
 .cache
 .docs
 .gitmodules
@@ -9,3 +8,9 @@ composer.lock
 .idea
 *.iml
 .php-cs-fixer.cache
+
+.env.local
+.env.local.php
+.env.*.local
+phpunit.xml
+.phpunit.result.cache
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -73,7 +73,7 @@ abstract class CredentialsLoader implements
     public static function fromEnv()
     {
         $path = $_ENV[self::ENV_VAR];
-        if (empty($path)) {
+        if (!isset($path) || empty($path)) {
             return null;
         }
         if (!file_exists($path)) {
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -72,10 +72,15 @@ abstract class CredentialsLoader implements
      */
     public static function fromEnv()
     {
-        $path = $_ENV[self::ENV_VAR];
+        $path = getenv(self::ENV_VAR);
+        if (!$path) {
+            $path = $_ENV[self::ENV_VAR];
+        }
+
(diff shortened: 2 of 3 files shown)


DamienLinux:
This pull request was made as part of the use of the Google pub/sub service

2 other git projects received a pull request of the same type for the same type of fix
- https://github.com/googleapis/google-cloud-php-pubsub/pull/8
- https://github.com/googleapis/google-cloud-php-core/pull/6

.gitignore
```diff
@@ -1,6 +1,5 @@
 *~
 vendor
-composer.lock
```
bshaffer, line 3: why did you remove `composer.lock` from the gitignore? We intentionally are ignoring this file.

src/CredentialsLoader.php
```diff
@@ -73,9 +73,14 @@ private static function isOnWindows()
     public static function fromEnv()
     {
         $path = getenv(self::ENV_VAR);
-        if (empty($path)) {
+        if (!$path) {
+            $path = $_ENV[self::ENV_VAR];
```
bshaffer, line 77: this will throw an Out Of Index error if the env var doesn't exist. This would need to be something like
```suggestion
            $path = $_ENV[self::ENV_VAR] ?? null;
```

src/CredentialsLoader.php
```diff
@@ -73,9 +73,14 @@ private static function isOnWindows()
     public static function fromEnv()
     {
         $path = getenv(self::ENV_VAR);
-        if (empty($path)) {
+        if (!$path) {
```
bshaffer, line 76: why did this change from `empty`? Let's keep it the way it was unless there's a specific reason to change it

bshaffer (changes requested):
Thank you for the PR! Please remove the `composer.lock` file and address my other suggestions.

src/CredentialsLoader.php
```diff
@@ -73,9 +73,14 @@ private static function isOnWindows()
...
+            $path = $_ENV[self::ENV_VAR];
+        }
+
+        if (!isset($path) || empty($path)) {
```
lanzelotik, line 80: Why can't we just check if it's empty? 

```
if (empty($path)) {
    return null;
}    
```

bshaffer:
Closing as it's been several months with no response from the author

PR #461, opened 2023-06-26
chore: refactor oauth2 tests

bshaffer:
Standard PHPUnit is to have only one test class in each test file. Substitute multiple test classes with proper use of the phpunit `@group` tag


noahdietz (approved):
Stamp

PR #462, opened 2023-06-27
feat: add STS

bshaffer:
Implement support for STS ([rfc8693](https://datatracker.ietf.org/doc/html/rfc8693)) in the auth library

 - [ ] Consider adding Workload Identity Federation as well (see https://github.com/googleapis/google-auth-library-ruby/pull/433 for example in the Ruby auth library)


src/CredentialSource/AwsNativeSource.php
```diff
@@ -0,0 +1,251 @@
...
+        if (is_null($signingVars)) {
+            throw new \LogicException('Unable to get credentials from ENV, and no security credentials URL provided');
+        }
+
+        // From here we use the signing vars to create the signed request to receive a token
+        [$accessKeyId, $secretAccessKey, $securityToken] = $signingVars;
+        $headers = $this->getSignedRequestHeaders($region, $accessKeyId, $secretAccessKey, $securityToken);
+
+        $url = new Uri($this->regionalCredVerificationUrl);
+        $url = $url->withQuery(self::CRED_VERIFICATION_QUERY);
```
bshaffer, line 67: this might be included already as part of the `regionalCredVerificationUrl`

bshaffer:
This PR has been split into two:
 - STS: https://github.com/googleapis/google-auth-library-php/pull/473
 - AWS credential source: https://github.com/googleapis/google-auth-library-php/pull/474

Issue #457, opened 2023-06-01
openssl_verify() error on OIDC authentication


wkopala:
I'm trying to authenticate a Cloud Scheduler job in PHP Laravel with this basic code:
    	$token = request()->bearerToken();

        if(!$token) {
            throw new Exception('Missing Auth header');
        }
        
        $auth = new AccessToken();
		$auth->verify($token, [
                'audience' => $audienceUrl,
                'throwException' => true,
            ]);
When I try to run the job, I keep getting this vague error:
openssl_verify(): Supplied key param cannot be coerced into a public key

Commenting out the auth code, the job runs just fine. Not sure what I'm doing wrong?

TheFetchZen:
We are also experiencing a similar issue

```
ErrorException: openssl_verify(): supplied key param cannot be coerced into a public key in /var/www/html/vendor/firebase/php-jwt/src/JWT.php:301 
It seems that the Google/Auth library might be causing the issue.
```

See the below stacktrace
```#1 /var/www/html/vendor/firebase/php-jwt/src/JWT.php(301): openssl_verify('JSUzI...', '\xAF2\xEF\xB0\xF6[\xE1\xBD\xE3\xCE\xA5\f\xC0\xE8\x05...', '-----BEGIN RSA ...', 'SHA256') #2 /var/www/html/vendor/firebase/php-jwt/src/JWT.php(144): Firebase\JWT\JWT::verify('JSUzI...', '\xAF2\xEF\xB0\xF6[\xE1\xBD\xE3\xCE\xA5\f\xC0\xE8\x05...', '-----BEGIN RSA ...', 'RS256') #3 [internal function]: Firebase\JWT\JWT::decode('eyJhbGciOiJSUzI...', Array) #4 /var/www/html/vendor/google/auth/src/AccessToken.php(474): call_user_func_array(Array, Array) #5 /var/www/html/vendor/google/auth/src/AccessToken.php(249): Google\Auth\AccessToken->callJwtStatic('decode', Array)```

krzyc:
Same problem here - switching to PKCS#8 resolves issue.
It is related to using obsolete version of OpenSSL (1.1.1).
But problem lies in phpseclib3 compatibility update (https://github.com/googleapis/google-auth-library-php/pull/425)
phpseclib2 used PKCS#8 by default, so upgrading to 1.26 and phpseclib3 can be breaking change.
[...]

PR #466, opened 2023-08-03
fix: use PKCS8 by default for ID token verify

marickvantuil:
This is (hopefully) a fix for #457.

It happens specifically with phpseclib 3.x.

---

When using the `AccessToken` class to decode an ID token, the "Federated Sign On" certificates are stored in the PKCS8 format after which the ID token is decoded with `Firebase/JWT`:

```php
$payload = $this->callJwtStatic('decode', [
    $token,
    $keys,
]);
```
Source: [AccessToken.php#L246](https://github.com/googleapis/google-auth-library-php/blob/main/src/AccessToken.php#L246)

Part of decoding the JWT is verifying the token which happens with `openssl_verify`:

```php
$success = \openssl_verify($msg, $signature, $keyMaterial, $algorithm);
```

Source: [JWT.php#L306](https://github.com/firebase/php-jwt/blob/main/src/JWT.php#L306)

I have to be honest that I am completely lacking any knowledge about RSA keys and their encoding, but according to [this StackOverflow issue](https://stackoverflow.com/a/64899836) PHP OpenSSL only accepts ["public key in X.509 style"](https://www.php.net/manual/en/function.openssl-pkey-get-public.php#101513). I can confirm this as I am getting the error "Supplied key param cannot be coerced into a public key" (same as in the linked issue at the start of this comment). After returning the key in PKCS8 format, the issue was fixed.

I understand my assessment of the issue might be completely wrong but I really like to help resolve this issue! 👍

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -408,7 +408,7 @@ class AccessToken
                 $exponent
             ]), 256),
         ]);
-        return $key->toString('PKCS1');
+        return $key->toString('PKCS8');
     }
 
     /**


bshaffer:
I have been looking around, and I can't find anywhere it's documented that the keys are `PKCS8`. The key URL is https://www.googleapis.com/oauth2/v3/certs

If we can confirm these changed from PCKS1 to PKCS8, I'm happy to make this change. Otherwise, I am confused why this would happen / would be necessary.

bshaffer (changes requested):
So, this has to do with the format in which the key is passed to openssl_verify. I suppose it's possible that your implementation doesn't support PKCS1? I find that strange, but I am not sure what the implications are across all our users and platforms.

the "Federated Sign On" certificates are stored in the PKCS8 format

This is a bit misleading - the certs aren't stored in this format, but they are converted to PKCS1 format from the [oauth2 certs URL](https://www.googleapis.com/oauth2/v3/certs) for openssl.

I don't want to just change the format being returned, because it could have a negative affect on some platforms. A better solution would be to:

detect the version of OpenSSL and/or whether or not PKCS1 and PKCS8 are supported, and use the best one
add a $format option to verify to allow modification (not preferred)
confirm with 100% confidence that this update would not break (I doubt this is possible)

marickvantuil:
Thanks for your the @bshaffer!

> I suppose it's possible that your implementation doesn't support PKCS1

After some more investigation, I found it throws this error when OpenSSL `1.1.1` is used. When I bump the PHP version (in my case from `8.1.12` to `8.1.13` using `php:8.1.12-cli-alpine` Docker image), OpenSSL is updated to 3.x and then it seems to work fine. 

So I'm not sure what the implications are! As far as I understand it then, something has changed in OpenSSL 3.x which fixes this issue but some other PHP 8.1 builds (depending on how it was built) ship with 1.1.1 in which this error happens. Hope I'm right here!

Not sure if it helps, but I set up a [reproducable example here](https://github.com/marickvantuil/google-auth-openssl-verify-example).

Output:

Screenshot 2023-08-05 at 18 11 07

bshaffer (approved):

PR #468, opened 2023-08-15
feat: add ID tokens for user refresh credentials

bshaffer:
Add ID token support to `UserRefreshCredentials`

 - [X] add tests
 - [x] manual E2E test to verify it works

Manually tested by running `gcloud auth application-default login`, and then executing the following:

```php
use Google\Auth\ApplicationDefaultCredentials;
use GuzzleHttp\Client;
use GuzzleHttp\Psr7\Request;
use Firebase\JWT\JWK;
use Firebase\JWT\JWT;

// Get the UserRefreshCredentials from "~/.config/gcloud/application_default_credentials.json",
// and supply the target audience
$credentials = ApplicationDefaultCredentials::getIdTokenCredentials('test-target-audience');
$idToken = $credentials->fetchAuthToken();

// Verify the ID token using the certs from Google
$certUrl = 'https://www.googleapis.com/oauth2/v3/certs';
$client = new GuzzleHttp\Client();
$certs = json_decode($client->send(new Request('GET', $certUrl))->getBody(), true);

$keys = JWK::parseKeySet($certs);
$decoded = JWT::decode($idToken['id_token'], $keys);
var_dump($decoded);
```

see also https://github.com/googleapis/google-auth-library-nodejs/pull/1811


src/Credentials/UserRefreshCredentials.php
```diff
@@ -55,48 +57,67 @@ class UserRefreshCredentials extends CredentialsLoader implements GetQuotaProjec
...
-        $jsonKey
+        $jsonKey,
+        string $targetAudience = null
     ) {
         if (is_string($jsonKey)) {
             if (!file_exists($jsonKey)) {
-                throw new \InvalidArgumentException('file does not exist');
+                throw new InvalidArgumentException('file does not exist');
             }
             $json = file_get_contents($jsonKey);
```
Hectorhammett, line 84: `file_get_contents` can return a `false` value on error. Should we take that into consideration?

src/Credentials/UserRefreshCredentials.php
```diff
@@ -55,48 +57,67 @@ class UserRefreshCredentials extends CredentialsLoader implements GetQuotaProjec
...
         }
         if (!array_key_exists('refresh_token', $jsonKey)) {
-            throw new \InvalidArgumentException(
+            throw new InvalidArgumentException(
                 'json key is missing the refresh_token field'
             );
         }
+        if ($scope && $targetAudience) {
+            throw new InvalidArgumentException(
+                'Scope and targetAudience cannot both be supplied'
```
Hectorhammett, line 106: It might be my non native brain but this reads weird to me.
`Scope and targetAudience cannot be supplied at the same time`

But might be just me though.
english

Hectorhammett (approved):

src/Credentials/UserRefreshCredentials.php
bshaffer: yes, that's a good point... maybe it should be `file does not exist or is unreadable` instead

src/Credentials/UserRefreshCredentials.php
```diff
@@ -55,48 +57,67 @@ class UserRefreshCredentials extends CredentialsLoader implements GetQuotaProjec
...
     public function __construct(
         $scope,
-        $jsonKey
+        $jsonKey,
+        string $targetAudience = null
     ) {
         if (is_string($jsonKey)) {
             if (!file_exists($jsonKey)) {
-                throw new \InvalidArgumentException('file does not exist');
+                throw new InvalidArgumentException('file does not exist');
```
bshaffer, line 82:
```suggestion
                throw new InvalidArgumentException('file does not exist or is unreadable');
```

PR #469, opened 2023-08-15
feat: check product name file on linux for gce

bshaffer:
cc @sai-sunder-s

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -101,6 +101,11 @@ class GCECredentials extends CredentialsLoader implements
     const FLAVOR_HEADER = 'Metadata-Flavor';
 
     /**
+     * The Linux file which contains the product name.
+     */
+    private const GKE_PRODUCT_NAME_FILE = '/sys/class/dmi/id/product_name';
+
+    /**
      * Note: the explicit `timeout` and `tries` below is a workaround. The underlying
      * issue is that resolving an unknown host on some networks will take
      * 20-30 seconds; making this timeout short fixes the issue, but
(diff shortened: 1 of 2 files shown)


TimurSadykov (approved):
LGTM, thanks!

src/Credentials/GCECredentials.php
```diff
@@ -340,6 +345,22 @@ public static function onGce(callable $httpHandler = null)
...
+
+        // Detect GCE residency on Linux
+        return self::detectResidencyLinux(self::GKE_PRODUCT_NAME_FILE);
+    }
+
+    private static function detectResidencyLinux(string $productNameFile): bool
+    {
+        if (file_exists($productNameFile)) {
+            $productName = trim((string) file_get_contents($productNameFile));
+            return 0 === strpos($productName, 'Google');
```
vishwarajanand, line 362: Maybe lowercase is better?
```suggestion
            return 0 === strpos($productName, 'google');
```

src/Credentials/GCECredentials.php
```diff
@@ -340,6 +345,22 @@ public static function onGce(callable $httpHandler = null)
             } catch (ConnectException $e) {
             }
         }
+
+        if (PHP_OS === 'Windows') {
+            // @TODO: implement GCE residency detection on Windows
```
noahdietz, line 350: No need to include it here, just want to make sure: Is there a bug filed internally or in this repo for this todo?

src/Credentials/GCECredentials.php
noahdietz: My only question is: Should we normalize the contents (e.g. force lowercase) in order to avoid any sort of subtle issues with string matching/discrepancies in this file's contents?

src/Credentials/GCECredentials.php
bshaffer: No, but that's a great idea! Filed https://github.com/googleapis/google-auth-library-php/issues/470

src/Credentials/GCECredentials.php
bshaffer: The Product Name is either `Google` or `Google Compute Engine`. The lowercase name `google` is not valid (or at least, there are no known Operating Systems which use that). So there's no reason to add a check for the lowercase value.

To my knowledge, other language implementations have not included such a check

noahdietz (approved):

PR #473, opened 2023-08-25
feat: add support for BYOID / STS

bshaffer:
@TimurSadykov @sai-sunder-s

PR #474, opened 2023-08-25
feat: add AWS credential source

bshaffer:
Depends on https://github.com/googleapis/google-auth-library-php/pull/473

 - [x] Successfully used to retrieve a token on AWS!
 - [ ] Verify implementation using test cases used by the other languages

### DIFF
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -1,0 +1,314 @@
+<?php
+/*
+ * Copyright 2023 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth\CredentialSource;
+
+use Google\Auth\CredentialSourceInterface;
+use Google\Auth\HttpHandler\HttpClientCache;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
+use GuzzleHttp\Psr7\Request;
+
+/**
+ * Authenticates requests using IAM credentials.
+ */
+class AwsNativeSource implements CredentialSourceInterface
+{
+    private const CRED_VERIFICATION_QUERY = 'Action=GetCallerIdentity&Version=2011-06-15';
+
+    private string $audience;
+    private string $regionUrl;
+    private string $regionalCredVerificationUrl;
+    private ?string $securityCredentialsUrl;
+
+    public function __construct(
+        string $audience,
+        string $regionUrl,
+        string $regionalCredVerificationUrl,
+        string $securityCredentialsUrl = null
+    ) {
+        $this->audience = $audience;
+        $this->regionUrl = $regionUrl;
+        $this->regionalCredVerificationUrl = $regionalCredVerificationUrl;
+        $this->securityCredentialsUrl = $securityCredentialsUrl;
+    }
+
+    public function fetchSubjectToken(callable $httpHandler = null): string
+    {
+        if (is_null($httpHandler)) {
+            $httpHandler = HttpHandlerFactory::build(HttpClientCache::getHttpClient());
+        }
+
+        $awsToken = self::fetchAwsTokenFromMetadata($httpHandler);
+
+        $signingVars = $this->securityCredentialsUrl
+            ? self::getSigningVarsFromUrl(
+                $httpHandler,
+                $this->securityCredentialsUrl,
+                self::getRoleName($httpHandler, $this->securityCredentialsUrl, $awsToken),
+                $awsToken
+            )
+            : self::getSigningVarsFromEnv();
+
+        if (is_null($signingVars)) {
+            throw new \LogicException('Unable to get credentials from ENV, and no security credentials URL provided');
+        }
+
+        $region = self::getRegion($httpHandler, $this->regionUrl, $awsToken);
+        $url = str_replace('{region}', $region, $this->regionalCredVerificationUrl);
+        $host = parse_url($url)['host'] ?? '';
+
+        // From here we use the signing vars to create the signed request to receive a token
+        [$accessKeyId, $secretAccessKey, $securityToken] = $signingVars;
+        $headers = self::getSignedRequestHeaders($region, $host, $accessKeyId, $secretAccessKey, $securityToken);
+
+        // Inject x-goog-cloud-target-resource into header
+        $headers['x-goog-cloud-target-resource'] = $this->audience;
+
+        // Format headers as they're expected in the subject token
+        $formattedHeaders= array_map(
+            fn ($k, $v) => ['key' => $k, 'value' => $v],
+            array_keys($headers),
+            $headers,
+        );
+
+        $request = [
+            'headers' => $formattedHeaders,
+            'method' => 'POST',
+            'url' => $url,
+        ];
+
+        return urlencode(json_encode($request) ?: '');
+    }
+
+    /**
+     * @internal
+     */
+    public static function fetchAwsTokenFromMetadata(callable $httpHandler): string
+    {
+        $url = 'http://169.254.169.254/latest/api/token';
+        $headers = [
+            'X-aws-ec2-metadata-token-ttl-seconds' => '21600'
+        ];
+        $request = new Request(
+            'PUT',
+            $url,
+            $headers
+        );
+
+        $response = $httpHandler($request);
+        return (string) $response->getBody();
+    }
+
+    /**
+     * @see http://docs.aws.amazon.com/general/latest/gr/sigv4-create-canonical-request.html
+     *
+     * @internal
+     *
+     * @return array<string, string>
+     */
+    public static function getSignedRequestHeaders(
+        string $region,
+        string $host,
+        string $accessKeyId,
+        string $secretAccessKey,
+        ?string $securityToken
+    ): array {
+        $service = 'sts';
+
+        # Create a date for headers and the credential string in ISO-8601 format
+        $amzdate = date('Ymd\THis\Z');
+        $datestamp = date('Ymd'); # Date w/o time, used in credential scope
+
+        # Create the canonical headers and signed headers. Header names
+        # must be trimmed and lowercase, and sorted in code point order from
+        # low to high. Note that there is a trailing \n.
+        $canonicalHeaders = sprintf("host:%s\nx-amz-date:%s\n", $host, $amzdate);
+        if ($securityToken) {
+            $canonicalHeaders .= sprintf("x-amz-security-token:%s\n", $securityToken);
+        }
+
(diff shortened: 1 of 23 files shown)


src/CredentialSource/AwsNativeSource.php
```diff
@@ -0,0 +1,314 @@
...
+            $headers['x-amz-security-token'] = $securityToken;
+        }
+
+        return $headers;
+    }
+
+    /**
+     * @internal
+     */
+    public static function getRegion(callable $httpHandler, string $regionUrl, string $awsToken): string
```
aeitzman, line 215: We should support getting region via the environment Variable and check that before we make an off box call:

https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#:~:text=AWS_DEFAULT_REGION%20%E2%80%93%20The%20default,overrides%20the%20AWS_DEFAULT_REGION

https://github.com/googleapis/google-auth-library-java/blob/aa047396112cbc0dd64f3fc147ed15f976c96bf0/oauth2_http/java/com/google/auth/oauth2/AwsCredentials.java#L260

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -102,6 +103,39 @@ private static function buildCredentialSource(array $jsonKey): ExternalAccountCr
...
+            && 1 === preg_match('/^aws(\d+)$/', $credentialSource['environment_id'], $matches)
+        ) {
+            if ($matches[1] !== '1') {
+                throw new InvalidArgumentException(
+                    "aws version \"$matches[1]\" is not supported in the current build."
+                );
+            }
+            if (!array_key_exists('region_url', $credentialSource)) {
+                throw new InvalidArgumentException(
+                    'The region_url field is required for aws1 credential source.'
```
aeitzman, line 117: region_url should not be required (because we can get it from the env variables in some cases)

src/CredentialSource/AwsNativeSource.php
```diff
@@ -0,0 +1,314 @@
...
+
+    public function fetchSubjectToken(callable $httpHandler = null): string
+    {
+        if (is_null($httpHandler)) {
+            $httpHandler = HttpHandlerFactory::build(HttpClientCache::getHttpClient());
+        }
+
+        $awsToken = self::fetchAwsTokenFromMetadata($httpHandler);
+
+        $signingVars = $this->securityCredentialsUrl
```
aeitzman, line 57: Nit: to be consistent with the other libraries, we should be getting the security credentials from the environment variables, even if the url has been provided in the credential configuration.

src/CredentialSource/AwsNativeSource.php
```diff
@@ -0,0 +1,314 @@
...
+        $this->regionalCredVerificationUrl = $regionalCredVerificationUrl;
+        $this->securityCredentialsUrl = $securityCredentialsUrl;
+    }
+
+    public function fetchSubjectToken(callable $httpHandler = null): string
+    {
+        if (is_null($httpHandler)) {
+            $httpHandler = HttpHandlerFactory::build(HttpClientCache::getHttpClient());
+        }
+
```
aeitzman, line 54: Should support Imsdv2 session token: https://github.com/googleapis/google-auth-library-nodejs/pull/1363, https://github.com/googleapis/google-auth-library-java/pull/850 this is used to get another header value that needs to be used for the region and security credentials call

src/CredentialSource/AwsNativeSource.php
bshaffer: Got it, so rather than "If securityCredentialsUrl then use the URL, otherwise use env vars", the logic should be "If env vars, use them, otherwise use the security credentials URL"?

src/Credentials/ExternalAccountCredentials.php
bshaffer: thank you! This should now be fixed.

src/CredentialSource/AwsNativeSource.php
bshaffer: (this should now be fixed)

src/CredentialSource/AwsNativeSource.php
bshaffer: Done! Added checking for `AWS_REGION` and `AWS_DEFAULT_REGION` env vars.

src/Credentials/ExternalAccountCredentials.php
bshaffer: If `region_url` isn't required, we need to update the AIP: https://google.aip.dev/auth/4117

Are you sure this is the case? I remember reading here (and also in some internal implementation docs) that the `region_url` field is required.

bshaffer: Submitted https://github.com/aip-dev/google.aip.dev/pull/1211 to track this

src/CredentialSource/AwsNativeSource.php
```diff
@@ -0,0 +1,360 @@
...
+        $headers = [];
+        if ($this->imdsv2SessionTokenUrl) {
+            $headers = [
+                'X-aws-ec2-metadata-token' => self::getImdsV2SessionToken($this->imdsv2SessionTokenUrl, $httpHandler)
```
vishwarajanand, line 73: Does it need `x-aws-ec2-metadata-token-ttl-seconds` as well? It doesnt seem to be present in AIP, but Ruby has it [here](https://github.com/BigTailWolf/google-auth-library-ruby/blob/dddbf2a50d538dbab6f4041b5a9b54d0b758e0b0/lib/googleauth/external_account/aws_credentials.rb#L114)

vishwarajanand: Okay, I saw its added in `getImdsV2SessionToken`

tests/fixtures6/aws_credentials.json
vishwarajanand, line 1: Where is this file used?

vishwarajanand: I think it was intended to be used in `ApplicationDefaultCredentialsTest->provideExternalAccountCredentials` ?

src/CredentialSource/AwsNativeSource.php
```diff
@@ -0,0 +1,360 @@
...
+    /**
+     * Return HMAC hash in binary string
+     */
+    private static function hmacSign(string $key, string $msg): string
+    {
+        return hash_hmac('sha256', self::utf8Encode($msg), $key, true);
+    }
+
+    /**
+     * @TODO add a fallback when mbstring is not available
```
vishwarajanand, line 340: Seems like the function `mb_convert_encoding` is present in all supported PHP versions, [ref](https://www.php.net/manual/en/function.mb-convert-encoding.php)

tests/fixtures6/aws_credentials.json
bshaffer: Good catch! When I split these PRs apart, this must have been dropped. Thanks for catching this.

src/CredentialSource/AwsNativeSource.php
bshaffer: See the [installation page](https://www.php.net/manual/en/mbstring.installation.php):

```
mbstring is a non-default extension. This means it is not enabled by default. You must explicitly enable the module with the configure option.
```
It's possible to compile PHP without the MB extension, or to have the extension compiled but not enabled, so we still want to include a fallback.

src/CredentialSource/AwsNativeSource.php
bshaffer: I've added support for this, but using the credentials that I received for testing (which do not have the `imdsv2_session_token_url` property in the JSON), this is now failing because the library fails to fetch the region from the server without the URL. 

This would imply that the parameter is required in cases where the signing vars are not set as environment variables. If this is the case, I think my app should throw an exception saying "imdsv2_session_token_url not found in credentials and is required when signing vars are not present as environment variables", or something along these lines. WDYT?

aeitzman: The parameter is not always required when the url is used, since users do not need to have imsdv2 enforced on their EC2 instances, so we shouldn't throw that error

bshaffer: After speaking with @aeitzman, the answer is that this would be an edgecase error due to a user mistake (such as my own) due to the wrong configuration being used on a machine it isn't meant for. We can handle this error if we'd like, but I agree that it's probably best to just throw the exception as-is.

vishwarajanand (approved):
The PR looks great! I've reviewed the changes and tested them on an Amazon EC2 machine to ensure they work correctly without failures. And an access token is generated successfully which can be used to make other calls.

However, I'd like to bring up a potential improvement and concern:

**Improvement:**

It might be a good idea for later into the PHP Client Library to better handle the `project_id` field. Currently, the AWS JSON key doesn't seem to include include `project_id` or `projectId` by default, unlike other authentication flows (e.g., service accounts) that do contain this key. While the audience URL does include a project number, there may be cases where having the `project_id` explicitly in the JSON key could be beneficial.

**Concern:**

I've noticed that the `StorageClient` may encounter issues in situations where users rely on automatic detection of the `project_id`. Other clients may also typically expect this field to be provided explicitly or get derived automatically from the JSON key. In the absence of the `project_id`, these cases could potentially lead to failures in user's application.


How I generated the token:
```
use Google\Auth\Credentials\ExternalAccountCredentials;

putenv(sprintf('GOOGLE_APPLICATION_CREDENTIALS=%s/%s', __DIR__, 'key.json'));
$creds = ApplicationDefaultCredentials::getCredentials('https://www.googleapis.com/auth/cloud-platform');
print_r($creds->fetchAuthToken())
```

bshaffer:
@vishwarajanand 

> the AWS JSON key doesn't seem to include include project_id or projectId by default, unlike other authentication flows (e.g., service accounts) that do contain this key

This is not accurate - no credential types contain or return the project ID. They return the "quota project ID", which is something different. However, it may be a good idea to have `ExternalAccountCredentials` support `quota_project_id`, which it currently doesn't support.

> I've noticed that the StorageClient may encounter issues in situations where users rely on automatic detection of the project_id

We've gone through a few iterations of "project ID detection", but it's a complex problem because the various client libraries do it a little differently (env var, client option, credential field, etc). However, this is a separate issue that is out of scope of this PR.

bshaffer:
@vishwarajanand  I stand corrected, it IS supported (via `ProjectIdProviderInterface`). It doesn't seem like this is supported in the NodeJS implementation or the python implementation, at least not from JSON credentials

https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L68-L83

https://github.com/googleapis/google-auth-library-python/blob/117e3253d27ca4519dabe169825a3a0473aff482/google/auth/external_account.py#L350-L353

I'll do some more research here...

bshaffer:
Okay it looks like there IS a way you can potentially get the project ID from the audience or from the workload identity pool, but after speaking with @aeitzman, this is a part of Workforce credentials, and can be a follow-up PR.

From the audience: 
 1. [Parse the project number from the audience string](https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L480C1-L497C1)
 2. [Exchange the project number for a project ID](https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L336-L343)

From Workload Identity Pool
1. [Get workforce pool user project from the JSON file](https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L186)
2. Exchange for a project ID (same as above)

PR #476, opened 2023-09-05
Fix implementation for psr/cache 3

plokko:
Fixed fatal error for missing return types


bshaffer:
For `psr/cache: v3` you should be using the `TypedItem` class (see [here](https://github.com/googleapis/google-auth-library-php/blob/main/src/Cache/TypedItem.php)), although this is handled internally by the cache pool class and so you shouldn't even need to use it. 

The `Item` class you are modifying is for `psr/cache: v2`, and should not have types added, as this breaks the implementation.

PR #477, opened 2023-09-07
feat: add support for universe domain (phase 1)

bshaffer:
Requirements:

1. For all credentials: Create the `universe_domain` read property. The default value for the property is `googleapis.com`.
2. For Service Account Credentials: Read the `universe_domain` from the credentials file and expose the value through the property created above. If the credentials file does not have the universe_domain value, the property should use the default value `googleapis.com`.

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -31,6 +31,7 @@ use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Exception\ServerException;
 use GuzzleHttp\Psr7\Request;
 use InvalidArgumentException;
+use UnexpectedValueException;
 
 /**
  * GCECredentials supports authorization on Google Compute Engine.
@@ -180,6 +181,11 @@ class GCECredentials extends CredentialsLoader implements
     private ?string $universeDomain;
 
     /**
+     * @var bool
+     */
+    private ?string $expectedUniverseDomain;
+
+    /**
      * @param Iam $iam [optional] An IAM instance.
      * @param string|string[] $scope [optional] the scope of the access request,
      *        expressed either as an array or as a space-delimited string.
@@ -188,8 +194,9 @@ class GCECredentials extends CredentialsLoader implements
      *   charges associated with the request.
      * @param string $serviceAccountIdentity [optional] Specify a service
      *   account identity name to use instead of "default".
-     * @param string $universeDomain [optional] Specify a universe domain to use
-     *   instead of fetching one from the metadata server.
+     * @param string $expectedUniverseDomain [optional] Specify a universe
+     *   domain which will be checked against the one returned by the metadata
+     *   server, and throw an exception on mismatch.
      */
     public function __construct(
         Iam $iam = null,
@@ -197,7 +204,7 @@ class GCECredentials extends CredentialsLoader implements
         $targetAudience = null,
         $quotaProject = null,
         $serviceAccountIdentity = null,
-        string $universeDomain = null
+        string $expectedUniverseDomain = null
     ) {
         $this->iam = $iam;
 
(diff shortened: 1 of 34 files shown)


src/Credentials/GCECredentials.php
```diff
@@ -500,6 +526,45 @@ public function getProjectId(callable $httpHandler = null)
...
+        }
+
+        // In the case of 404, we return the default universe domain.
+        try {
+            $this->universeDomain = $this->getFromMetadata(
+                $httpHandler,
+                self::getUniverseDomainUri()
+            );
+        } catch (ClientException $e) {
+            return self::DEFAULT_UNIVERSE_DOMAIN;
```
bshaffer, line 562: This should only return when the client exception is a 404
```suggestion
            if ($e->hasResponse() && 404 == $e->getResponse()->getStatusCode()) {
                return self::DEFAULT_UNIVERSE_DOMAIN;
            }
            throw $e;
```

Aungkhant884:
Patch

Aungkhant884 (approved):

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -328,6 +334,19 @@ public function getQuotaProject()
...
+     *
+     * @return string
+     */
+    public function getUniverseDomain(): string
+    {
+        if (null === $this->universeDomain) {
+            return self::DEFAULT_UNIVERSE_DOMAIN;
+        }
+        return $this->universeDomain;
+    }
```
vishwarajanand, line 348: Shorter?
```suggestion
        return $this->universeDomain ?: self::DEFAULT_UNIVERSE_DOMAIN;
    }
```

bshaffer: I prefer the way it is for readability.

bshaffer:
@vishwarajanand we won't cut a release until the freeze, but I'd love to get an approval on this before the EOY

src/GetUniverseDomainInterface.php
```diff
@@ -0,0 +1,35 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+/**
+ * An interface implemented by objects that can get quota projects.
```
vishwarajanand, line 21:
```suggestion
 * An interface implemented by objects that can get universe domain for Google Cloud.
```

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -99,6 +99,11 @@ class ServiceAccountCredentials extends CredentialsLoader implements
      */
     private $jwtAccessCredentials;
 
+    /**
+     * @var string|null
```
vishwarajanand, line 103: It seems this is needed to preserve backwards compatibility for the case of manually created CredentialsWrapper by users.

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -328,6 +334,19 @@ public function getQuotaProject()
...
     }
 
+    /**
+     * Get the universe domain configured in the JSON credential.
+     *
+     * @return string
+     */
+    public function getUniverseDomain(): string
+    {
+        if (null === $this->universeDomain) {
```
vishwarajanand, line 344: Do you want to handle the case when `universeDomain` is an empty string? Or is it guaranteed to be having a valid value (eg. when always downloaded from gcp console)

vishwarajanand (approved):
Lgtm

src/Credentials/ServiceAccountCredentials.php
bshaffer: Good question. The spec does not provide any guidance around empty strings for universe domain (go/cloudsdk-tpc-authlibs), except when it is a 404 error from the MDS (which is handled in phase 2).

If we added special handling for empty strings, it would be to throw an exception. But as an empty string for a universe domain will throw an error in the client libraries anyway (because we throw an error for a user-supplied empty string, and also for a mismatch), I can't think of a real important reason why we need to throw an exception now, and would prefer to leave it out (unless we can think of a specific case where it would be important to do so).

Issue #463, opened 2023-07-13
feat: AccessToken respecting certificate expiration


ynikitin-etsy:
## Is your feature request related to a problem? Please describe.

I am using AccessToken to validate Google Service Account JWTs and that's it. When given the cache, AccessToken caches the public certificates so not to fetch them from Google very time. Currently, AccessToken.php expires the fetched public certificates after one hour. However, Google and other providers often rotate their keys. Thus, it might be possible to cache public certs that are stale for almost an hour (~59min, if fetched right before rotation) given specific timing requirements. 

## Describe the solution you'd like
Google recommends respecting the Cache-Control header, which it seems this library currently doesn't.

https://developers.google.com/identity/gsi/web/guides/verify-google-id-token

## Describe alternatives you've considered
It might be possible to catch the UnexpectedValueException with message `"kid" invalid, unable to lookup correct key` and clear the cache so to force a refetch on revalidation. But, this would happen if you are given a JWT that's signed by another provider public keys' and you would refetch certs all the time.

## Additional context
Not sure if my understanding is correct here, or there is a better way to make sure certs are always fresh and fetched at appropriate times. Are PR's welcomed for this?

bshaffer:
Hello! Thank you for submitting this issue. You are correct that the `Cache-Control` header which comes back with the response for the certs is currently not being used. I've submitted a PR (#479) to add this behavior. Please take a look and leave a review if you'd like. And to answer your questions, PR's are ALWAYS welcome!

ynikitin-etsy:
@bshaffer Thank you, this is amazing! I will review asap!

PR #479, opened 2023-09-08
feat: respect cache control for access token certs

bshaffer:
addresses #463

Extracts `max-age` from the `Cache-Control` header in order to set the cache lifetime of the certs.

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -372,10 +372,10 @@ class AccessToken
             if ($cacheControl = $response->getHeaderLine('Cache-Control')) {
                 array_map(function ($value) use (&$expireTime) {
                     list($key, $value) = explode('=', $value) + [null, null];
-                    if ($key == 'max-age') {
+                    if (trim($key) == 'max-age') {
                         $expireTime = '+' . $value . ' seconds';
                     }
-                }, explode(', ', $cacheControl));
+                }, explode(',', $cacheControl));
             }
             return [
                 json_decode((string) $response->getBody(), true),
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -33,6 +33,7 @@ use phpseclib\Math\BigInteger as BigInteger2;
 use phpseclib3\Crypt\PublicKeyLoader;
 use phpseclib3\Math\BigInteger as BigInteger3;
 use Psr\Cache\CacheItemPoolInterface;
+use Psr\Http\Message\ResponseInterface;
 use RuntimeException;
 use SimpleJWT\InvalidTokenException;
 use SimpleJWT\JWT as SimpleJWT;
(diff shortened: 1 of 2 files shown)


src/AccessToken.php
```diff
@@ -360,14 +359,28 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
 
         if ($response->getStatusCode() == 200) {
-            return json_decode((string) $response->getBody(), true);
+            if ($cacheControl = $response->getHeaderLine('Cache-Control')) {
+                array_map(function ($value) use (&$expireTime) {
+                    list($key, $value) = explode('=', $value) + [null, null];
+                    if ($key == 'max-age') {
+                        $expireTime = '+' . $value . ' seconds';
+                    }
+                }, explode(', ', $cacheControl));
```
ynikitin-etsy, line 378: I feel like it might be better to seperate on a comma, and trim the key and value after. At least this page, doesnt say anything about spaces always being present after commas: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

src/AccessToken.php
```diff
@@ -360,14 +359,28 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
 
         $httpHandler = $this->httpHandler;
         $response = $httpHandler(new Request('GET', $url), $options);
 
         if ($response->getStatusCode() == 200) {
-            return json_decode((string) $response->getBody(), true);
+            if ($cacheControl = $response->getHeaderLine('Cache-Control')) {
+                array_map(function ($value) use (&$expireTime) {
+                    list($key, $value) = explode('=', $value) + [null, null];
+                    if ($key == 'max-age') {
```
ynikitin-etsy, line 375: Same here probably, better to trim key and value just in case?

src/AccessToken.php
```diff
@@ -360,14 +359,28 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
+                        $expireTime = '+' . $value . ' seconds';
```
ynikitin-etsy, line 376: this seems to work, but the PHP spec seems to say only sec or second is valid, not seconds? https://www.php.net/manual/en/datetime.formats.php#datetime.formats.relative

ynikitin-etsy (approved):
Thanks a bunch for this!

src/AccessToken.php
bshaffer: I believe that is referring to "second" in the sense of "first, second, third...". The keyword "seconds" is definitely valid.

src/AccessToken.php
bshaffer: good idea!

src/AccessToken.php
bshaffer: done!

src/AccessToken.php
bshaffer: Here:

```
(( 'msec' | 'millisecond' | 'µsec' | 'microsecond' | 'usec' | 'sec' | 'second' | 'min' | 'minute' | 'hour' | 'day' | 'fortnight' | 'forthnight' | 'month' | 'year') 's'?)
```
You can see there's the optional `s` at the end, which confirms `seconds` is valid.

ynikitin-etsy (approved):

vishwarajanand (approved):
Lgtm

PR #480, opened 2023-09-14
feat: Add Metrics Trait

yash30201:
* Add Metrics Trait
* Add unit tests for metrics trait

### DIFF
--- a/src/MetricsTrait.php
+++ b/src/MetricsTrait.php
@@ -1,0 +1,144 @@
+<?php
+/*
+ * Copyright 2023 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace Google\Auth;
+
+/**
+ * Provides methods for fetching and updation of auth metrics headers.
+ *
+ * @internal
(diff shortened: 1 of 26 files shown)


src/MetricsTrait.php
```diff
@@ -0,0 +1,143 @@
...
+    public static $requestTypeAccessToken = 'auth-request-type/at';
+    public static $requestTypeIdToken = 'auth-request-type/it';
+    public static $requestTypeMdsPing = 'auth-request-type/mds';
+
+    // Credential type
+    public static $credTypeUser = 'cred-type/u';
+    public static $credTypeSaAssertion = 'cred-type/sa';
+    public static $credTypeSaJwt = 'cred-type/jwt';
+    public static $credTypeSaMds = 'cred-type/mds';
+    public static $credTypeSaImpersonate = 'cred-type/imp';
```
bshaffer, line 39: At least based on their usage in this class, these should all be private by default. But I think I'd get rid of them all together and just replace their usage with the strings themselves - they're only used in one place each, and they're not variable, so they shouldn't be variables. 

They could be constants, but as traits don't allow constants, it won't work here.

src/MetricsTrait.php
```diff
@@ -0,0 +1,143 @@
...
+
+    // TODO: Find a way to get the auth version
+    // Auth library version
+    public static $version = '10.0.0';
```
bshaffer, line 43: Why is the placeholder `10.0.0`?

I'd suggest making this a placeholder function instead, like this:

```php
private static getAuthLiberaryVersion(): string
{
    throw new LogicException('This must be implemented by the class which uses the trait');
}
```

bshaffer (commented):
Great PR! I've left a few initial thoughts, and here are a few more:

Does this need to be in a trait? Traits are only necessary if multiple classes share this functionality. If it's exposed as a trait, but not actually used by classes IN the auth library, it might be better to put it in the library where the classes are which DOES use it (e.g. GAX).

Is there another PR coming which shows how this trait will be implemented? It's hard to review this PR without that one.

yash30201:
Will work on a different approach as per design doc.

PR #482, opened 2023-09-20
feat: support universe domain in service account and metadata credentials

bshaffer:
Requirements
 - [x] 1. Authentication libraries must read the universe_domain property from the credentials files (done #477)
 - [x] 2. Authentication libraries must allow the end-user to set the universe_domain in code
 - [x] 3. Authentication libraries must use the universe_domain variable/property/method name
 - [x] 4. Authentication libraries must not use hardcoded non-constant information in the authentication flows
 - [x] 5. Authentication libraries must implement retrieving the universe_domain from the VM Metadata Server
 - [x] 6. Authentication libraries must expose the universe_domain in all credentials classes (done in #477)
 - [x] 7. ~~Authentication libraries must add the universe_domain property to the token authentication credentials constructor (if supported) and document the default behavior~~ _Not applicable unless we add an Access Token credential type_
 - [x] 8. Authentication libraries must only attempt self-signed JWT flow for Service Accounts when authenticating outside `googleapis.com`
 - [x] 9. Authentication libraries must throw an Exception if domain-delegation is configured for Service Accounts when authenticating outside `googleapis.com`
 - [x] 10. Authentication libraries must implement `IdToken` flow for Service Accounts when authenticating outside `googleapis.com` or throw an exception

Dependencies
 - [x] This now relies on https://github.com/googleapis/google-auth-library-php/pull/505 for requirement 9, as this previously was not behaving as expected.


src/Credentials/GCECredentials.php
```diff
@@ -500,6 +526,56 @@ public function getProjectId(callable $httpHandler = null)
...
+     * Fetch the default universe domain from the metadata server.
+     *
+     * Returns null if called outside GCE.
+     *
+     * @param callable $httpHandler Callback which delivers psr7 request
+     * @return string
+     */
+    public function getUniverseDomain(callable $httpHandler = null): string
+    {
+        if ($this->universeDomain) {
```
bshaffer, line 539: This should be strict
```suggestion
        if (null !== $this->universeDomain) {
```

src/Credentials/GCECredentials.php
```diff
@@ -500,6 +526,56 @@ public function getProjectId(callable $httpHandler = null)
...
+            }
+            $this->universeDomain = self::DEFAULT_UNIVERSE_DOMAIN;
+        }
+
+        // We expect in some cases the metadata server will return an empty string for the universe
+        // domain. In this case, the auth library MUST return the default universe domain.
+        if ('' === $this->universeDomain) {
+            $this->universeDomain = self::DEFAULT_UNIVERSE_DOMAIN;
+        }
+
```
bshaffer, line 575: I believe this concession is no longer a requirement, and should be removed.
```suggestion
```

tests/Credentials/GCECredentialsTest.php
```diff
@@ -517,7 +517,60 @@ public function testGetUniverseDomain()
...
+            );
+            return new Psr7\Response(200, [], Utils::streamFor(''));
+        };
+
+        // Assert the default universe domain is returned instead of the empty string.
+        $this->assertEquals(
+            GCECredentials::DEFAULT_UNIVERSE_DOMAIN,
+            $creds->getUniverseDomain($httpHandler)
+        );
+    }
```
bshaffer, line 568: This requirement has changed, and should no longer apply
```suggestion
    public function testGetUniverseDomainEmptyStringReturnsEmptyString()
    {
        $creds = new GCECredentials();
        $creds->setIsOnGce(true);

        // Pretend we are on GCE and mock the MDS returning an empty string for the universe domain.
        $httpHandler = function ($request) {
            $this->assertEquals(
                '/computeMetadata/v1/universe/universe_domain',
                $request->getUri()->getPath()
            );
            return new Psr7\Response(200, [], Utils::streamFor(''));
        };

        // Assert the default universe domain is returned instead of the empty string.
        $this->assertEquals(
            '',
            $creds->getUniverseDomain($httpHandler)
        );
    }
```

src/Credentials/GCECredentials.php
```diff
@@ -500,6 +526,56 @@ public function getProjectId(callable $httpHandler = null)
...
+            ?: HttpHandlerFactory::build(HttpClientCache::getHttpClient());
+
+        if (!$this->hasCheckedOnGce) {
+            $this->isOnGce = self::onGce($httpHandler);
+            $this->hasCheckedOnGce = true;
+        }
+
+        if (!$this->isOnGce) {
+            return self::DEFAULT_UNIVERSE_DOMAIN;
+        }
```
vishwarajanand, line 553: Lets say a user is not on GCE but wants to use TPC, will they see a default universe domain always?

src/Credentials/GCECredentials.php
```diff
@@ -500,6 +526,56 @@ public function getProjectId(callable $httpHandler = null)
...
+            // should return the default universe domain.
+            if (!$e->hasResponse() || 404 != $e->getResponse()->getStatusCode()) {
+                throw $e;
+            }
+            $this->universeDomain = self::DEFAULT_UNIVERSE_DOMAIN;
+        }
+
+        // We expect in some cases the metadata server will return an empty string for the universe
+        // domain. In this case, the auth library MUST return the default universe domain.
+        if ('' === $this->universeDomain) {
```
vishwarajanand, line 572: do you think empty would be a better check? 
On a side note, the doc does mention the case of an empty string, but let's say the returned value is a long string with only spaces, that may lead to cascading failures.

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -164,7 +164,7 @@ public function __construct(
         ]);
 
         $this->projectId = $jsonKey['project_id'] ?? null;
-        $this->universeDomain = $jsonKey['universe_domain'] ?? null;
+        $this->universeDomain = $jsonKey['universe_domain'] ?? self::DEFAULT_UNIVERSE_DOMAIN;
```
vishwarajanand, line 167: Should we also throw in case there's `domain-delegation`?
I am not very sure but it appears that we need to throw (near line#146) in case `sub` as well as `universe_domain` is set in the key, as per AL9 of [cloudsdk-tpc-authlibs](http://go/cloudsdk-tpc-authlibs#heading=h.y388tia22eux)

vishwarajanand (approved):
Lgtm, approving since there are no major comments.

## Tests:

1. I tested the changes on `apis-tpclp.goog` universe domain and validated that `ServiceAccountCredentials` generates a valid access_token while using the provided universe domain value. 

2. I was also able to query the compute instances in the project on the URL: `https://compute.apis-tpclp.goog/compute/staging_v1/<more params>`, which captures the universe domain in the endpoint.

3. When I use the access token generated in step#1, and use the regular compute api: 
`https://compute.googleapis.com/compute/staging_v1/<more params>` I see an expected failure with error code 401 and status `UNAUTHENTICATED`:
```
Request had invalid authentication credentials. Expected OAuth 2 access token, login cookie or other valid authentication credential. See https://developers.google.com/identity/sign-in/web/devconsole-project.
```

src/Credentials/GCECredentials.php
bshaffer: Only if they create gce credentials manually. Otherwise application default would throw an exception.

src/Credentials/GCECredentials.php
bshaffer: I don't think we should add checks for hypothetical bugs from the MDS. We should assume if it's 200, then the value is accurate. I don't even like the empty string check (and it recently changed to an optional requirement, so we COULD remove it), but I left it in since it's part of the spec.

src/Credentials/ServiceAccountCredentials.php
bshaffer: Thank you for catching this, I completely missed it. PHP definitely supports `domain-delegation`

https://github.com/googleapis/google-auth-library-php/blob/b687d5efa8dc71782a5181c501adc28c4d89cdac/src/Credentials/ServiceAccountCredentials.php#L114-L115

I'll look more into it

src/Credentials/GCECredentials.php
vishwarajanand: We are using `empty` in gax and other places, so I think its better to be consistent for this field in all places.

src/Credentials/ServiceAccountCredentials.php
bshaffer: Okay, I added the exception and test in 5f8b409c2839b0a64e760af4faa8a050e783b47b, and added a fix for domain-wide delegation that we had missed in #505

vishwarajanand: Approved it

Issue #483, opened 2023-10-05
Add support for ProjectIdProviderInterface to external account credentials


bshaffer:
See https://github.com/googleapis/google-auth-library-php/pull/474

As part of Workforce Credentials, we will need support getting a ProjectID (e.g. implementing `ProjectIdProviderInterface`) in `ExternalAccountCredentials`. Here's examples of how this is done in other languages:

From the audience: 
 1. [Parse the project number from the audience string](https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L480C1-L497C1)
 2. [Exchange the project number for a project ID](https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L336-L343)

From Workload Identity Pool
1. [Get workforce pool user project from the JSON file](https://github.com/googleapis/google-auth-library-nodejs/blob/bf23bad915cae8994b2953b406d04112848c5224/src/auth/baseexternalclient.ts#L186)
2. Exchange for a project ID (same as above)

**IMPORTANT**: Because the caching layer for the auth library is currently implemented in a `FetchAuthTokenCache` class (which is a pattern that has caused us a lot of headache), we will need to consider refactoring caching implementation to make this feature work. One way would be to implement the `CacheTrait`. Another would be to have it extend `FetchAuthTokenCache`. A third option would be create a new `AuthTokenCacheTrait` which contains logic from `FetchAuthTokenCache` to be shared between the two classes. A potentially larger refactor is also on the table (see https://github.com/googleapis/google-auth-library-php/pull/322/files)

PR #485, opened 2023-10-05
feat: workforce credentials

bshaffer:
addresses #483 

(note: relevant testing requested in https://github.com/phpspec/prophecy/issues/611)


bshaffer:
@aeitzman I'd love to get your review on this one as well!

src/FetchAuthTokenCache.php
```diff
@@ -146,9 +146,12 @@ public function signBlob($stringToSign, $forceOpenSsl = false)
...
         }
 
-        // Pass the access token from cache to GCECredentials for signing a blob.
-        // This saves a call to the metadata server when a cached token exists.
-        if ($this->fetcher instanceof Credentials\GCECredentials) {
+        // Pass the access token from cache for credentials that sign blobs
+        // using the IAM API. This saves a call to fetch an access token when a
+        // cached token exists.
+        if ($this->fetcher instanceof Credentials\GCECredentials
+            || $this->fetcher instanceof Credentials\ImpersonatedServiceAccountCredentials
```
aeitzman, line 153: Is this related to this change?

src/FetchAuthTokenCache.php
```diff
@@ -189,6 +192,15 @@ public function getProjectId(callable $httpHandler = null)
             );
         }
 
+        // Pass the access token from cache for credentials that require an
+        // access token to fetch the project ID. This saves a call to fetch an
+        // access token when a cached token exists.
+        if ($this->fetcher instanceof Credentials\ExternalAccountCredentials) {
```
aeitzman, line 198: Just double checking since I don't fully understand the caching flow in this library, this call is only used if getProjectId is explicitly called by a user/some other library right? Just want to make sure that the default flow for getting an access token doesn't break if cloudresourcemanager is down.

src/FetchAuthTokenCache.php
bshaffer: Hmm, no. I am not sure why this would be in here. I probably just realized that it was something that needed fixing, when adding similar logic below for `getProjectId`.

src/FetchAuthTokenCache.php
bshaffer: Yes that's correct - this is ONLY for calls to `$credentials->getProjectId()`

aeitzman (approved):

tests/Credentials/ExternalAccountCredentialsTest.php
```diff
@@ -326,15 +318,198 @@ public function testFetchAuthTokenWithImpersonation()
...
+            $this->assertEquals('Bearer some-token', $request->getHeaderLine('authorization'));
+            $body = $this->prophesize(StreamInterface::class);
+            $body->__toString()->willReturn(json_encode(['projectId' => 'test-project-id']));
+
+            $response = $this->prophesize(ResponseInterface::class);
+            $response->getBody()->willReturn($body->reveal());
+            $response->hasHeader('Content-Type')->willReturn(false);
+
+            return $response->reveal();
+        };
```
yash30201, line 409: Can you help me understand why are we defining this httphandler if the token and eventually the project id would be fetched from the mocked cache?

yash30201 (commented):
Just left few nits and comments to double check my understand. Looks good implementation wise.

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -90,14 +101,25 @@ public function __construct(
...
 
         $this->auth = new OAuth2([
             'tokenCredentialUri' => $jsonKey['token_url'],
             'audience' => $jsonKey['audience'],
             'scope' => $scope,
             'subjectTokenType' => $jsonKey['subject_token_type'],
             'subjectTokenFetcher' => self::buildCredentialSource($jsonKey),
+            'additionalOptions' => $this->workforcePoolUserProject
+                ? ['userProject' => $this->workforcePoolUserProject]
+                : [],
```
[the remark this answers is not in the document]
bshaffer: Why remove the training comma? This is now best practice for arrays (adding a new item will only diff the added line)

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -238,4 +261,68 @@ public function getQuotaProject()
...
+     *        token. **Defaults to** `null`.
+     * @return string|null
+     */
+    public function getProjectId(callable $httpHandler = null, string $accessToken = null)
+    {
+        if (isset($this->projectId)) {
+            return $this->projectId;
+        }
+
+        $projectNumber = $this->getProjectNumber() ?: $this->workforcePoolUserProject;
```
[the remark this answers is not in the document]
bshaffer: Yes, they are equivalent. As far as I know, one is not preferred over the other (as in this case, the left side would never be undefined)

tests/Credentials/ExternalAccountCredentialsTest.php
bshaffer: Sure. This method is testing the getProjectId method, which makes an API call using the project number in order to retrieve the project ID. It's also testing that a cached access token is used when that API call is made, instead of retrieving a new one.

So the cache is for the access token, which is needed for the API request to get the project ID. The httpHandler is the HTTP request made when getProjectId is called, which is not cached.

src/Credentials/ExternalAccountCredentials.php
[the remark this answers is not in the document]
bshaffer:
```suggestion
        $projectNumber = $this->getProjectNumber() ?? $this->workforcePoolUserProject;
```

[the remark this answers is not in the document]
yash30201: > [See here for more info](https://stackoverflow.com/questions/34571330/php-short-ternary-elvis-operator-vs-null-coalescing-operator)

Ohh, thanks for ref. I learnt something new : )   !

> However, I'm happy to change it if you think it's more readable.

It's up to you. Personally I also don't believe in enforcing some way if it's the exact same as the other.

tests/Credentials/ExternalAccountCredentialsTest.php
yash30201: Oh, now this makes sense. 
I feel adding a small explanation there in comment would go a long way to quickly understand it.

PR #489, opened 2023-10-17
fix: allowed_algs not properly set for string value

bshaffer:


vishwarajanand (approved):
Seems like a bug..

bshaffer:
> Seems like a bug..

It was indeed :(

PR #492, opened 2023-10-26
chore: Refactor Auth Token Middleware's token logic

yash30201:
go/php-auth-token-unifomity

### DIFF
--- a/src/Middleware/AuthTokenMiddleware.php
+++ b/src/Middleware/AuthTokenMiddleware.php
@@ -42,7 +42,10 @@ class AuthTokenMiddleware
     private $httpHandler;
 
     /**
-     * @var FetchAuthTokenInterface
+     * It must be an implementation of FetchAuthTokenInterface.
+     * It may also implement UpdateMetadataInterface allowing direct
+     * retrieval of auth related headers
+     * @var FetchAuthTokenInterface|UpdateMetadataInterface
      */
     private $fetcher;
 
@@ -123,7 +126,7 @@ class AuthTokenMiddleware
     /**
      * Fetch auth headers.
      *
-     * @return array
+     * @return array<string, array|string>
      */
     private function fetchAuthHeaders()
     {
--- a/src/Middleware/AuthTokenMiddleware.php
+++ b/src/Middleware/AuthTokenMiddleware.php
@@ -126,7 +126,7 @@ class AuthTokenMiddleware
     /**
      * Fetch auth headers.
      *
-     * @return array<string, array|string>
+     * @return array<string, string[]|string>
      */
     private function fetchAuthHeaders()
     {
--- a/src/Middleware/AuthTokenMiddleware.php
+++ b/src/Middleware/AuthTokenMiddleware.php
@@ -105,20 +105,7 @@ class AuthTokenMiddleware
                 return $handler($request, $options);
             }
 
-            if (!$this->fetcher instanceof UpdateMetadataInterface ||
-                ($this->fetcher instanceof FetchAuthTokenCache && !$this->fetcher->getFetcher() instanceof UpdateMetadataInterface)
-            ) {
-                $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
-            } else {
-                $headers = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
-                $request = Utils::modifyRequest($request, ['set_headers' => $headers]);
-                if ($this->tokenCallback) {
-                    $auth_tokens = $this->fetcher->fetchAuthToken($this->httpHandler);
-                    if (array_key_exists('access_token', $auth_tokens)) {
-                        call_user_func($this->tokenCallback, $this->fetcher->getCacheKey(), $auth_tokens['access_token']);
-                    }
-                }
-            }
+            $request = $this->addAuthHeaders($request);
 
             if ($quotaProject = $this->getQuotaProject()) {
                 $request = $request->withHeader(
@@ -132,32 +119,33 @@ class AuthTokenMiddleware
     }
 
     /**
-     * Call fetcher to fetch the token.
+     * Adds auth related headers to the request.
      *
-     * @return string|null
+     * @param RequestInterface $request
+     * @return RequestInterface
      */
-    private function fetchToken()
+    private function addAuthHeaders(RequestInterface $request)
     {
-        $auth_tokens = (array) $this->fetcher->fetchAuthToken($this->httpHandler);
+        if (!$this->fetcher instanceof UpdateMetadataInterface ||
+            ($this->fetcher instanceof FetchAuthTokenCache &&
+             !$this->fetcher->getFetcher() instanceof UpdateMetadataInterface)
+        ) {
+            $token = $this->fetcher->fetchAuthToken();
+            $request = $request->withHeader(
+                'authorization', 'Bearer ' . ($token['access_token'] ?? $token['id_token'])
+            );
+        } else {
+            $headers = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
+            $request = Utils::modifyRequest($request, ['set_headers' => $headers]);
+        }
 
-        if (array_key_exists('access_token', $auth_tokens)) {
-            // notify the callback if applicable
-            if ($this->tokenCallback) {
-                call_user_func(
-                    $this->tokenCallback,
-                    $this->fetcher->getCacheKey(),
-                    $auth_tokens['access_token']
-                );
+        if ($this->tokenCallback && ($token = $this->fetcher->getLastReceivedToken())) {
+            if (array_key_exists('access_token', $token)) {
+                call_user_func($this->tokenCallback, $this->fetcher->getCacheKey(), $token['access_token']);
             }
-
-            return $auth_tokens['access_token'];
         }
 
-        if (array_key_exists('id_token', $auth_tokens)) {
-            return $auth_tokens['id_token'];
-        }
-
-        return null;
+        return $request;
     }
 
     /**
--- a/src/Middleware/AuthTokenMiddleware.php
+++ b/src/Middleware/AuthTokenMiddleware.php
@@ -108,7 +108,7 @@ class AuthTokenMiddleware
             if (!$this->fetcher instanceof UpdateMetadataInterface ||
                 ($this->fetcher instanceof FetchAuthTokenCache && !$this->fetcher->getFetcher() instanceof UpdateMetadataInterface)
             ) {
-                $request =$request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
+                $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
             } else {
                 $headers = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
                 $request = Utils::modifyRequest($request, ['set_headers' => $headers]);
--- a/src/Middleware/AuthTokenMiddleware.php
+++ b/src/Middleware/AuthTokenMiddleware.php
@@ -46,7 +46,7 @@ class AuthTokenMiddleware
      * It must be an implementation of FetchAuthTokenInterface.
      * It may also implement UpdateMetadataInterface allowing direct
      * retrieval of auth related headers
-     * @var FetchAuthTokenInterface|UpdateMetadataInterface
+     * @var FetchAuthTokenInterface
      */
     private $fetcher;
 
--- a/src/Middleware/AuthTokenMiddleware.php
+++ b/src/Middleware/AuthTokenMiddleware.php
@@ -17,8 +17,10 @@
 
 namespace Google\Auth\Middleware;
 
+use Google\Auth\FetchAuthTokenCache;
 use Google\Auth\FetchAuthTokenInterface;
 use Google\Auth\GetQuotaProjectInterface;
+use Google\Auth\UpdateMetadataInterface;
 use Psr\Http\Message\RequestInterface;
 
 /**
@@ -99,7 +101,13 @@ class AuthTokenMiddleware
                 return $handler($request, $options);
             }
 
-            $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
+            foreach ($this->fetchAuthHeaders() as $key => $value) {
+                if ($key == 'authorization') {
+                    $request = $request->withHeader($key, $value);
+                } else {
+                    $request = $request->withAddedHeader($key, $value);
+                }
+            }
 
             if ($quotaProject = $this->getQuotaProject()) {
                 $request = $request->withHeader(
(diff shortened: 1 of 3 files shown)


bshaffer (commented):
The way you have it now seems to be making more changes than necessary. If the goal of this PR is to just to use `updateMetadata` when available, I think it would be better to simplify this logic as something like this:

```php
// Use "updateMetadata" if it exists, otherwise use "fetchAuthToken"
if (!$this->fetcher instanceof UpdateMetadataInterface ||
   ($this->fetcher FetchAuthTokenCache && !$this->fetcher->getFetcher() instanceof UpdateMetadataInterface)
) {
    // previous behavior (unlikely - only really happens if user supplies a custom fetcher)
    $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
} else {
    $updatedHeaders = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
    $request = Utils::modifyRequest($request, ['set_headers' => $updatedHeaders]);
}
```

src/Middleware/AuthTokenMiddleware.php
```diff
@@ -113,32 +124,54 @@ public function __invoke(callable $handler)
...
+                $authHeaders = ['authorization' => 'Bearer ' . $authTokens['access_token']];
+            } elseif (array_key_exists('id_token', $authTokens)) {
+                $authHeaders = ['authorization' => 'Bearer ' . $authTokens['id_token']];
+            }
+        }
+
+        if (!empty($authHeaders)) {
+            if (empty($authTokens)) {
+                $authTokens = $this->fetcher->getLastReceivedToken();
+            }
```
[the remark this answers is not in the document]
yash30201: Oh, what is the reason behind this decision?

yash30201:
> The way you have it now seems to be making more changes than necessary. If the goal of this PR is to just to use `updateMetadata` when available, I think it would be better to simplify this logic as something like this:
> 
> ```
> // Use "updateMetadata" if it exists, otherwise use "fetchAuthToken"
> if (!$this->fetcher instanceof UpdateMetadataInterface ||
>    ($this->fetcher FetchAuthTokenCache && !$this->fetcher->getFetcher() instanceof UpdateMetadataInterface)
> ) {
>     // previous behavior (unlikely - only happens if user supplies a custom fetcher)
>     $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
> } else {
>     // new behavior - allow the credentials to add / update request headers
>     $updatedHeaders = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
>     $request = Utils::modifyRequest($request, ['set_headers' => $updatedHeaders]);
> }
> ```

Thanks for the pointer. We would be required to fetch auth token when using update metadata because we need to call the token callback in the event it isn't an identity token request. I've amended the logic as such, please take a look. Now it looks cleaner.

src/Middleware/AuthTokenMiddleware.php
```diff
@@ -99,7 +105,20 @@ public function __invoke(callable $handler)
                 return $handler($request, $options);
             }
 
-            $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
+            if (!$this->fetcher instanceof UpdateMetadataInterface ||
+                ($this->fetcher instanceof FetchAuthTokenCache && !$this->fetcher->getFetcher() instanceof UpdateMetadataInterface)
+            ) {
+                $request =$request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
```
bshaffer, line 111: nit: CS fix
```suggestion
                $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
```

src/Middleware/AuthTokenMiddleware.php
```diff
@@ -99,7 +105,20 @@ public function __invoke(callable $handler)
                 return $handler($request, $options);
             }
 
-            $request = $request->withHeader('authorization', 'Bearer ' . $this->fetchToken());
```
bshaffer, line 102: Since we're removing the call to `fetchToken`, we should remove the `fetchToken` method below (because it's no longer being used)

src/Middleware/AuthTokenMiddleware.php
```diff
@@ -99,7 +105,20 @@ public function __invoke(callable $handler)
...
+            } else {
+                $headers = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
+                $request = Utils::modifyRequest($request, ['set_headers' => $headers]);
+                if ($this->tokenCallback) {
+                    $auth_tokens = $this->fetcher->fetchAuthToken($this->httpHandler);
+                    if (array_key_exists('access_token', $auth_tokens)) {
+                        call_user_func($this->tokenCallback, $this->fetcher->getCacheKey(), $auth_tokens['access_token']);
+                    }
+                }
+            }
```
bshaffer, line 121: A few things here: 
1. This will call `fetchAuthToken` twice, which we don't want (it would end up making multiple HTTP requests in some cases). 
2. This behavior (using `tokenCallback`) is essentially deprecated (we do not want to support it in auth V2), but since we'd like to maintain BC, we can get the access token from the auth header
3. Since this is being applied to ALL calls in the current model, it should be outside the `else` statement to ensure that BC doesn't break
```suggestion
            } else {
                $headers = $this->fetcher->updateMetadata($request->getHeaders(), null, $this->httpHandler);
                $request = Utils::modifyRequest($request, ['set_headers' => $headers]);
            }

            // support tokenCallback for compatibility (this should be removed in the next major version)
            if ($this->tokenCallback && $bearer = $request->getHeaderLine('authorization')) {
                if (0 === strpos($bearer, 'Bearer ')) {
                    call_user_func($this->tokenCallback, $this->fetcher->getCacheKey(), substr($bearer, 7));
                }
           }

```

bshaffer (commented):
Good stuff! We're getting closer here.

src/Middleware/AuthTokenMiddleware.php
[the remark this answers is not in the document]
bshaffer: The responsibility of `getLastReceivedToken` was for the convenience of caching, but with properly implemented caching, it should never be necessary. The library can rely on `fetchAuthToken` and the cache layer to get a token. 

The case you have here, that `fetchAuthToken` fails but `getLastReceivedToken` has a valid token, should never occur. `fetchAuthToken` should only happen when we need a new auth token, so if it fails, returning the last received token would just return an invalid token anyway.

src/Middleware/AuthTokenMiddleware.php
yash30201: It's used in line 111

src/Middleware/AuthTokenMiddleware.php
yash30201: This approach has two problems it seems:

1) We need to call token callback only in the case of  an `access_token` which seems impossible to determine from `updateMetadata()`'s response. 
    1) `lastTokenReceived` method solved this problem but it seems we are deprecating it.
2) Current implementation doesn't retries the `updateMetadata` call, we need to change it I think.

bshaffer: > `fetchToken` retries the fetching calls using exponential backoff.

I am confused by this - retries are not represented in `fetchToken`, and exponential backoff doesn't exist in the auth library.

> We need to call token callback only in the case of an `access_token`

Ahh good call, I missed that this only happened previously with `access_token` and not with `id_token`. 

> `lastTokenReceived` method solved this problem but it seems we are deprecating it.

True, and I agree that in this case (to preserve BC, and to not call `fetchAuthToken` twice) it's the best solution to use `getLastReceivedToken`. Yes we want to discourage the use of `getLastReceivedToken`, but we  _also_ want to discourage using `tokenCallback` (both these "features" are technical debt that should be removed). 

... So changing my code above, something like this should work instead: 

```php

if ($this->tokenCallback && $token = $this->fetcher->getLastReceivedToken()) {
    if (isset($token['access_token'])) {
        call_user_func($this->tokenCallback, $this->fetcher->getCacheKey(), $token['access_token']);
    }
}
```

...And then we can remove the `tokenCallback` logic from `fetchToken`, and keep it all in one place

src/Middleware/AuthTokenMiddleware.php
bshaffer: Thanks, I missed that!

We could potentially remove it, since we'll have the `tokenCallback` logic in one place, and it's only called once. It could just be done inline like this:

```php
$token = (array) $this->fetcher->fetchAuthToken($this->httpHandler);
$bearer = $token['access_token'] ?? $token['id_token'] ?? '';
$request = $request->withHeader('authorization', 'Bearer ' . $bearer);
```

But that's up to you

tests/Middleware/AuthTokenMiddlewareTest.php
```diff
@@ -282,6 +259,8 @@ public function testShouldNotifyTokenCallback(callable $tokenCallback)
         $this->mockFetcher->fetchAuthToken(Argument::any())
             ->shouldBeCalledTimes(1)
             ->willReturn($cachedValue);
+        $this->mockFetcher->getLastReceivedToken()
+            ->willReturn($cachedValue);
```
yash30201, line 263: This additonal prophecy is now necessary as we are introducting another check for `$token` from `getLastReceivedToken` before calling `tokenCallback`.

src/Middleware/AuthTokenMiddleware.php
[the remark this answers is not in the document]
yash30201: Makes sense. Thanks for the explaining : ).

src/Middleware/AuthTokenMiddleware.php
yash30201: > I am confused by this - retries are not represented in fetchToken, and exponential backoff doesn't exist in the auth library.

I’m sorry I created a confusion by saying stuff about `ExponentialBackoff`. At that moment, somehow I mixed google-cloud-php’s Core library knowledge with this.  

> True, and I agree that in this case (to preserve BC, and to not call fetchAuthToken twice) it's the best solution to use getLastReceivedToken. Yes we want to discourage the use of getLastReceivedToken, but we also want to discourage using tokenCallback (both these "features" are technical debt that should be removed).

Yess! I was thinking along the same lines that we should go ahead with using `getlastFetchedToken` method in the case of invoking `tokenCallback`.

tests/FetchAuthTokenTest.php
```diff
@@ -53,10 +54,20 @@ class_implements($fetcherClass)
         )) {
             $mockFetcher->getQuotaProject()->shouldBeCalledTimes(1);
         }
-        $mockFetcher->fetchAuthToken(Argument::any())
-            ->shouldBeCalledTimes(1)
-            ->will($httpHandler);
+
+        if (is_a($fetcherClass, UpdateMetadataInterface::class, true)) {
```
bshaffer, line 58: We don't really use `is_a` anymore... you can just use `instanceof` for consistency
```suggestion
        if ($fetcherClass instanceof UpdateMetadataInterface::class) {
```

tests/FetchAuthTokenTest.php
```diff
@@ -53,10 +54,20 @@ class_implements($fetcherClass)
...
+            $mockFetcher->updateMetadata(Argument::cetera())
+            ->shouldBeCalledTimes(1)->will(function () use (&$httpHandlerCalled) {
+                $httpHandlerCalled = true;
+                return ['authorization' => ['Bearer xyz']];
+            });
```
bshaffer, line 63: nit: this should be indented once more
```suggestion
                ->shouldBeCalledTimes(1)->will(function () use (&$httpHandlerCalled) {
                    $httpHandlerCalled = true;
                    return ['authorization' => ['Bearer xyz']];
                });
```

tests/Middleware/AuthTokenMiddlewareTest.php
```diff
@@ -306,6 +285,63 @@ public function testShouldNotifyTokenCallback(callable $tokenCallback)
...
+        $mock = new MockHandler([function ($request, $options) use ($authHeaders) {
+            $this->assertEquals($authHeaders['authorization'], $request->getHeaderLine('authorization'));
+            $this->assertArrayHasKey('x-goog-api-client', $request->getHeaders());
+            return new Response(200);
+        }]);
+        $callable = $middleware($mock);
+        $callable($request, ['auth' => 'google_auth']);
+    }
+
+    private function runTestCase($fetcher, bool $isGoogleAuth)
```
bshaffer, line 336: I like the idea of simplifying all these tests!

However, since there's only one place where `$isGoogleAuth` is false, I'd rather just run the test explicitly in this case (instead of calling this function) to keep this function simple.
```suggestion
    private function runTestCase($fetcher)
```

tests/Middleware/AuthTokenMiddlewareTest.php
```diff
@@ -306,6 +285,63 @@ public function testShouldNotifyTokenCallback(callable $tokenCallback)
...
+            ->willReturn($authResult);
+        $this->mockFetcher->getLastReceivedToken()
+            ->willReturn(['access_token' => '1/abcdef1234567890']);
+
+        $request = new Request('GET', 'http://foo.com');
+
+        $middleware = new AuthTokenMiddleware($this->mockFetcher->reveal());
+        $mock = new MockHandler([new Response(200)]);
+        $callable = $middleware($mock);
+        $callable($request, ['auth' => 'google_auth']);
```
bshaffer, line 306: I am not sure what this is testing. The return values are mocked, but we never actually verify that the headers were set on the request.

tests/Middleware/AuthTokenMiddlewareTest.php
```diff
@@ -306,6 +285,63 @@ public function testShouldNotifyTokenCallback(callable $tokenCallback)
...
+            ->willReturn(['access_token' => '1/abcdef1234567890']);
+
+        $middleware = new AuthTokenMiddleware($this->mockFetcher->reveal());
+        $mock = new MockHandler([function ($request, $options) use ($authHeaders) {
+            $this->assertEquals($authHeaders['authorization'], $request->getHeaderLine('authorization'));
+            $this->assertArrayHasKey('x-goog-api-client', $request->getHeaders());
+            return new Response(200);
+        }]);
+        $callable = $middleware($mock);
+        $callable($request, ['auth' => 'google_auth']);
```
bshaffer, line 333: This one is better, since we have assertions in the mock handler, but we have no guarantee that the function in `MockHandler` is called. Something like this could work for this:

```php
$called = false;
$mock = new MockHandler([function ($request, $options) use ($authHeaders, &$called) {
    $this->assertEquals($authHeaders['authorization'], $request->getHeaderLine('authorization'));
    $this->assertArrayHasKey('x-goog-api-client', $request->getHeaders());
    $called = true;
    return new Response(200);
}]);
// and then at the end of the test...
$this->assertTrue($called);
```

tests/Middleware/AuthTokenMiddlewareTest.php
```diff
@@ -306,6 +285,63 @@ public function testShouldNotifyTokenCallback(callable $tokenCallback)
...
+
+    public function testOverlappingAddAuthHeadersFromUpdateMetadata()
+    {
+        $authHeaders = [
+            'authorization' => 'Bearer 1/abcdef1234567890',
+            'x-goog-api-client' => 'extra-value'
+        ];
+
+        $request = new Request('GET', 'http://foo.com');
+        $request->withHeader('x-goog-api-client', 'default-value');
```
bshaffer, line 317: This line does nothing, since `withHeader` returns a _new_ Request object, and does not modify the existing request object. It needs to be:
```suggestion
        $request = $request->withHeader('x-goog-api-client', 'default-value');
```

But since it wasn't working before it must not be testing this value as expected.

bshaffer (commented):
This is looking great! I only have some suggestions on how to improve the tests.

tests/FetchAuthTokenTest.php
yash30201: `instanceof` isn't allowing a comparison between string representations : (. For eg, 

```php
class Alpha {}
$str = Alpha::class;
echo ($str instanceof Alpha ? 'true', 'false') . PHP_EOL;

// Prints: false
```

and 
```php
class Alpha {}
$str = Alpha::class;
$a =  ($str instanceof Alpha::class ? 'true', 'false')
```

throws a syntax error `PHP Parse error:  syntax error, unexpected 'class' (T_CLASS), expecting variable (T_VARIABLE) or '$'`

bshaffer: I didn't realize that we were dealing with a string and not an object so, you're right, `is_a` is the way to go.

tests/Middleware/AuthTokenMiddlewareTest.php
yash30201: I was testing if the underlying credentials methods (like `updateMetadata`, `getLastTokenReceived`) are called by the new update logic.

Nice callout. It makes sense to test the set headers too! I've updated the testing.

tests/Middleware/AuthTokenMiddlewareTest.php
yash30201: Thanks for the pointer. Implemented.

tests/Middleware/AuthTokenMiddlewareTest.php
yash30201: Yeah that line was placed with the purpose of testing how headers are merge but later I figured that testing this functionality is unnecessary. We're not supposed to test how the headers are merged as it's the responsibility of `updateMetadata` method which is out of scope for our context.

Just forgot to remove that line. It's removed now.

tests/Middleware/AuthTokenMiddlewareTest.php
yash30201: Makes sense. Implemented.

bshaffer (approved):

PR #500, opened 2023-12-04
chore: Add version in composer

yash30201:
`release-please` bot updates `CHANGELOG.md` and `composer.json` with every release ([code pointer](https://github.com/googleapis/release-please/blob/main/src/updaters/php/root-composer-update-packages.ts)). Right now auth library's `composer.json` doesn't has any `version` key-value pair. 

Manually adding a `version` in composer ensures release-please updates it with every release.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,6 +5,7 @@
   "keywords": ["google", "oauth2", "authentication"],
   "homepage": "http://github.com/google/google-auth-library-php",
   "license": "Apache-2.0",
+  "version": "1.33.0",
   "support": {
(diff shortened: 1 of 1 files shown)


bshaffer (commented):
I would prefer to use the `VERSION` file (as we do in all our subrepos) as opposed to conflating this with `composer.json`. See [here](https://stackoverflow.com/questions/36769404/what-is-the-meaning-of-version-field-in-composer-json) for an example for why this isn't encouraged.

yash30201:
Closing as https://github.com/googleapis/release-please/pull/2153 would allow updation of `VERSION` file which is more appropriate for the use case.

PR #505, opened 2023-12-13
feat: respect sub for domain-wide delegation in service account creds

bshaffer:
This is also a "fix" because, this should always have been the case. We want to respect the "sub" field, and use the oauth token endpoint instead of SSJs when it's supplied.


vishwarajanand (approved):
Lgtm

PR #506, opened 2023-12-13
chore: fix php warning in AuthTokenMiddleware

bshaffer:
A PHP warning was introduced in https://github.com/googleapis/google-auth-library-php/pull/492 where we didn't check if one of the values of the array was set. 

```
$ vendor/bin/phpunit tests/Middleware/AuthTokenMiddlewareTest.php --filter testDoesNotAddAnAuthorizationHeaderOnNoAccessToken
PHPUnit 9.6.13 by Sebastian Bergmann and contributors.

PHP Warning:  Undefined array key "id_token" in src/Middleware/AuthTokenMiddleware.php on line 135
.                                                                   1 / 1 (100%)
Warning: Undefined array key "id_token" in src/Middleware/AuthTokenMiddleware.php on line 135


Time: 00:00.012, Memory: 8.00 MB

OK (1 test, 1 assertion)
```

This is not too concerning, since PHP will cast this to an empty string, but we should get rid of the warning by explicitly using an empty string when neither token types are set. 

Also, we set `convertWarningsToExceptions="true"` and `convertWarningsToExceptions="true"`, where they were previously false. This will ensure warnings like this fail our tests, so they don't happen again.


yash30201 (approved):
Lgtm.

Thanks for fixing this warning Brent!

PR #511, opened 2023-12-27
fix: do not return GDU outside GCE for GCECredentials

bshaffer:


src/Credentials/GCECredentials.php
```diff
@@ -548,10 +546,6 @@ public function getUniverseDomain(callable $httpHandler = null): string
             $this->hasCheckedOnGce = true;
         }
 
-        if (!$this->isOnGce) {
-            return self::DEFAULT_UNIVERSE_DOMAIN;
```
ajupazhamayil, line 552: @bshaffer Why is this behavior changed?

ajupazhamayil: Also, is this a breaking change?

bshaffer: It's not breaking because we haven't released these changes yet (the `getUniverseDomain` function was merged into `main` but it hasn't been released)

It's changed because this is not the behavior we want. If the credentials don't think we're on GCE, we don't care, we will try to get the credentials anyway (which will probably throw an error) rather than assuming GDU. There's no reason to assume GDU here, so the logic is being removed.

noahdietz (approved):

PR #512, opened 2023-12-28
Suggestions: Metrics with trait suggestions

bshaffer:
Suggestions for https://github.com/googleapis/google-auth-library-php/pull/509

Major changes:
 - Consolidate all header methods to one `getMetricHeader` for simplicity, and remove `applyHeaders`
 - If a metrics header is supplied in the metadata, it will be used instead of the default (rather than having multiple metrics headers)
 - Add headers in `FetchAuthTokenCache` when a cached token is found (rather than in `CredentialsLoader`)

Minor changes
 - Make `$credType` a constant and add `getCredType` method to each Credential (and to `UpdateMetadataInterface`)
 - Make `$metricsHeaderKey` a constant and move it to `UpdateMetadataInterface::METRIC_METADATA_KEY` (to be consistent with `AUTH_METADATA_KEY`)
 - Rename all `$metricsHeaders` parameters to `$headers`, as this is more accurate.
 - Import `MetricsTrait` in `UpdateMetadataTrait` (since it depends on this)
 - Make `getVersion` private

### DIFF
--- a/src/Credentials/ExternalAccountCredentials.php
+++ b/src/Credentials/ExternalAccountCredentials.php
@@ -35,7 +35,6 @@ use InvalidArgumentException;
 class ExternalAccountCredentials implements FetchAuthTokenInterface, UpdateMetadataInterface, GetQuotaProjectInterface
 {
     use UpdateMetadataTrait;
-    use MetricsTrait;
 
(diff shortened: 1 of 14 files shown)


bshaffer:
Closing because these suggestions have been implemented

PR #513, opened 2023-12-29
feat: ServiceAccountJwtAccessCredentials omits expires_in and token_type

vishwarajanand:
Fixing two issues here:

1) Returning `type` and `expires_in` with the token in case of JWT keys. ref: https://github.com/googleapis/google-cloud-php/pull/6850#issuecomment-1872164327 Screenshot 2023-12-28 at 00 06 36

2) Fixing lint issues with `AccessToken.php`:
The lint check is complaining about retuning `string|array` in `loadPhpsecPublicKey` method ([ref](https://github.com/googleapis/google-auth-library-php/actions/runs/7357834969/job/20030094538?pr=513)) and expects a `string` instead. I am not sure whether this is the best fix, but looks like `__toString` method ([ref](https://github.com/phpseclib/phpseclib/blob/498e67a0c82bd5791fda9b0dd0f4ec8e8aebb02d/phpseclib/Crypt/RSA.php#L2070)) is marked as `public` also returns a key. Throwing a `TypeError` if the key is array and not a string since thats consistent with an existing error in that case from [here](https://github.com/firebase/php-jwt/blob/1b9e87184745595ef70540613c0cb9de09bebab3/src/Key.php#L31).

We confirmed that the `access_token` cannot have any other `type` and `expires_in` is usually ~3600 always. So, this change might be redundant, but IMHO its good to have.

### DIFF
--- a/src/Credentials/ServiceAccountJwtAccessCredentials.php
+++ b/src/Credentials/ServiceAccountJwtAccessCredentials.php
@@ -151,7 +151,11 @@ class ServiceAccountJwtAccessCredentials extends CredentialsLoader implements
         // Set the self-signed access token in OAuth2 for getLastReceivedToken
         $this->auth->setAccessToken($access_token);
 
-        return ['access_token' => $access_token];
+        return [
+            'access_token' => $access_token,
+            'expires_in' => $this->auth->getExpiry(),
+            'token_type' => 'Bearer'
+        ];
     }
 
     /**


bshaffer:
I think the best thing here would be to allow the method loadPhpsecPublicKey to return string or array, rather than throw a type error, which could be a breaking change.

vishwarajanand:
@bshaffer this is not a breaking change because it's limited to a `private` function return type and the `TypeError` in the affected case would anyways be thrown from `Firebase\JWT\Key` class, [here](https://github.com/googleapis/google-auth-library-php/blob/21f9292d3ae91b53df675e95dc2bcc445e8f3835/src/AccessToken.php#L243).

If we allow `string|array`, it will fail lint check because `Firebase\JWT\Key::__construct(..)` doesn't accept `array` and will throw a lint failure.

bshaffer:
> this is not a breaking change because it's limited to a private function

When throwing an exception, the visibility of the function isn't necessarily important since the exception can bubble up through to public functions 

> the TypeError in the affected case would anyways be thrown from Firebase\JWT\Key class

That's a good point 

> If we allow string|array, it will fail lint check

I see. I'd be curious to know 1. What caused the type to change (and suddenly trigger the lint error when it wasn't before) and 2. In what scenarios does phpseclib return a string here? This will help us make sure that throwing an exception is indeed the right option here (as opposed to say, excluding a version of phpseclib in our dependencies, or casting the string to array).

src/AccessToken.php
```diff
@@ -411,7 +416,11 @@ private function loadPhpsecPublicKey(string $modulus, string $exponent): string
                     $exponent
                 ]), 256),
             ]);
-            return $key->getPublicKey();
+            $formattedPublicKey = $key->getPublicKey();
+            if (is_string($formattedPublicKey)) {
+                return $formattedPublicKey;
+            }
+            throw new TypeError('Failed to initialize the key');
         }
```
vishwarajanand, line 424: We can actually remove this first `if` block since this is from an old phpseclib version, which is not used anymore due to our [composer.json](https://github.com/googleapis/google-auth-library-php/blob/b8e4dad3fd2f96bb76699bf9a22e40c1b1d2f471/composer.json#L25). Thoughts?
```
    "phpseclib/phpseclib": "^3.0",
```

vishwarajanand:
@bshaffer here are the answers:

1. `What caused the type to change (and suddenly trigger the lint error when it wasn't before)`
It's because of a recent phpseclin release v3.0.35, [relevant commit](https://github.com/phpseclib/phpseclib/commit/f0194cf63959d1d5f49e583403f862a9e4c6900f). `toString()` seems to have had a conscious change which is breaking our lint. There is only one place where the `AsymmetricKey::toString()` is implemented and doesn't return string, and that's `PublicKey::toString()`, [ref](https://github.com/phpseclib/phpseclib/blob/4b1827beabce71953ca479485c0ae9c51287f2fe/phpseclib/Crypt/RSA/PublicKey.php#L473-L475) which returns a `mixed`.
2. `throwing an exception is indeed the right option here (as opposed to say, excluding a version of phpseclib in our dependencies, or casting the string to array)`
Excluding versions likely wont help coz this will be visible in all future `3.0.35+` versions. 
Wrt Casting, I am not sure because its `mixed` so not sure of the array key, etc and this class is used in a lot of places. I will raise this with phpseclib as well, in hope of a reply.

src/AccessToken.php
```diff
@@ -421,7 +430,12 @@ private function loadPhpsecPublicKey(string $modulus, string $exponent): string
                 $exponent
             ]), 256),
         ]);
-        return $key->toString('PKCS8');
+        // Ensure the key is correctly initialized
+        $formattedPublicKey = $key->toString('PKCS8');
+        if (is_string($formattedPublicKey)) {
+            return $formattedPublicKey;
+        }
```
bshaffer, line 437: I've tracked this down to the latest version of PHPseclib. When we use `phpseclib/phpseclib: 3.0.34`, this lint error does not happen:

```
composer require --dev "phpseclib/phpseclib:3.0.34"
```

bshaffer, line 437: Okay I was able to track it down to this change:

https://github.com/phpseclib/phpseclib/commit/f0194cf63959d1d5f49e583403f862a9e4c6900f

It seems like if you do `$key->toString('Raw')`, it will return an array. So this will never return an array for us, since we have a hardcoded value.

I think the best thing to do then would just be to add a string cast to get rid of the error. Something like this:
```suggestion
            return (string) $key->toString('PKCS8');
```

Since there seems to be multiple things happen in this PR, I'm going to open a new PR to solve the linting issue, and that part can be removed.

bshaffer: Update - this still fails, so I am going to use the method here of throwing a type exception

vishwarajanand: This is exactly what I mentioned above. https://github.com/googleapis/google-auth-library-php/pull/513#issuecomment-1874447071

vishwarajanand:
@bshaffer I have split this PR into 2 more PRs, listed in order of how they could be merged:
1. For fixing lint: https://github.com/googleapis/google-auth-library-php/pull/517
2. For adding expires_in and token_type in Jwt: this PR
3. For deprecating phpseclibV2 classes: https://github.com/googleapis/google-auth-library-php/pull/518

src/AccessToken.php
vishwarajanand: Addressed in https://github.com/googleapis/google-auth-library-php/pull/518

bshaffer (approved):
I'm calling this a feature since it adds functionality that previously wasn't there

PR #515, opened 2024-01-02
fix: lint error from new phpseclib

bshaffer:
Fixes a linting error caused in the latest version ([3.0.35](https://github.com/phpseclib/phpseclib/releases/tag/3.0.35)) of PHPSeclib due to the `AsymmetricKey::toString` function [having its return type changed](https://github.com/phpseclib/phpseclib/blame/4b1827beabce71953ca479485c0ae9c51287f2fe/phpseclib/Crypt/Common/AsymmetricKey.php#L97) from `string` to `string|array`

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -38,6 +38,7 @@ use SimpleJWT\InvalidTokenException;
 use SimpleJWT\JWT as SimpleJWT;
 use SimpleJWT\Keys\KeyFactory;
 use SimpleJWT\Keys\KeySet;
+use TypeError;
 use UnexpectedValueException;
 
 /**
@@ -421,7 +422,11 @@ class AccessToken
                 $exponent
             ]), 256),
         ]);
-        return $key->toString('PKCS8');
+        $formattedPublicKey = $key->toString('PKCS8');
+        if (!is_string($formattedPublicKey)) {
+            throw new TypeError('Failed to initialize the key');
+        }
+        return $formattedPublicKey;
     }
 
     /**


ajupazhamayil (approved):

vishwarajanand:
This is already being covered here: https://github.com/googleapis/google-auth-library-php/pull/513

src/AccessToken.php
```diff
@@ -421,7 +422,11 @@ private function loadPhpsecPublicKey(string $modulus, string $exponent): string
                 $exponent
             ]), 256),
         ]);
-        return $key->toString('PKCS8');
+        $formattedPublicKey = $key->toString('PKCS8');
```
vishwarajanand, line 425: Should we remove the first `if` block here, because with phpseclib3 this condition won't happen?

```php
class_exists(RSA::class) && class_exists(BigInteger2::class)
```

ajupazhamayil:
Removing approval here as I am not sure which PR will be merged.. both look the same to me.. Thank you!

bshaffer:
@vishwarajanand 

> This is already being covered here: #513

That PR is addressing two issues in one, and also the fix is not the same as I've implemented (there's an exception being thrown for `$key->getPublicKey();` for no reason that I can tell.

If you'd like to split the fix out into a separate PR and remove the other exception being thrown for no reason, I'm happy to approve it and merge it, but we shouldn't put two different things in a single PR.

src/AccessToken.php
bshaffer: I'd rather not conflate dropping support for PHPSeclib 2 part of this PR

vishwarajanand:
We can close this.. I am splitting this PR: https://github.com/googleapis/google-auth-library-php/pull/513

PR #516, opened 2024-01-03
feat: Enable Observability Metrics (without an extra trait)

yash30201:


yash30201:
Closing this in favour of #509 after internal discussion with @bshaffer .

PR #518, opened 2024-01-03
deprecate: remove support for phpseclib V2

vishwarajanand:
Unit Tests are working in my local. 

I linked the new auth library all the way upto `google-cloud-php` and ran storage system tests against PROD, and they pass. Hence, i am confident the changes are good to review.

### DIFF
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -396,24 +396,6 @@ class AccessToken
         if (!class_exists(RSA::class)) {
             throw new RuntimeException('Please require phpseclib/phpseclib v2 or v3 to use this utility.');
         }
-
-        /**
-         * phpseclib calls "phpinfo" by default, which requires special
-         * whitelisting in the AppEngine VM environment. This function
-         * sets constants to bypass the need for phpseclib to check phpinfo
-         *
-         * @see phpseclib/Math/BigInteger
-         * @see https://github.com/GoogleCloudPlatform/getting-started-php/issues/85
-         * @codeCoverageIgnore
-         */
-        if (filter_var(getenv('GAE_VM'), FILTER_VALIDATE_BOOLEAN)) {
-            if (!defined('MATH_BIGINTEGER_OPENSSL_ENABLED')) {
-                define('MATH_BIGINTEGER_OPENSSL_ENABLED', true);
-            }
-            if (!defined('CRYPT_RSA_MODE')) {
-                define('CRYPT_RSA_MODE', RSA::MODE_OPENSSL);
-            }
-        }
     }
 
     private function loadPhpsecPublicKey(string $modulus, string $exponent): string
--- a/src/AccessToken.php
+++ b/src/AccessToken.php
@@ -395,7 +395,7 @@ class AccessToken
     private function checkAndInitializePhpsec()
     {
         if (!class_exists(RSA::class)) {
-            throw new RuntimeException('Please require phpseclib/phpseclib v2 or v3 to use this utility.');
+            throw new RuntimeException('Please require phpseclib/phpseclib v3 to use this utility.');
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/AccessToken.php
```diff
@@ -394,74 +393,24 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
-         * @codeCoverageIgnore
-         */
-        if (filter_var(getenv('GAE_VM'), FILTER_VALIDATE_BOOLEAN)) {
-            if (!defined('MATH_BIGINTEGER_OPENSSL_ENABLED')) {
-                define('MATH_BIGINTEGER_OPENSSL_ENABLED', true);
-            }
-            if (!defined('CRYPT_RSA_MODE')) {
-                define('CRYPT_RSA_MODE', RSA::MODE_OPENSSL);
-            }
-        }
```
vishwarajanand, line 452: This part is no longer needed since phpseclib no longer calls `phpinfo()`, ref commit: https://github.com/phpseclib/phpseclib/commit/7a628c62b0071d03bde6be82c2f77acdadb20df2

src/AccessToken.php
```diff
@@ -394,74 +393,24 @@ private function retrieveCertsFromLocation($url, array $options = [])
...
     {
-        if (!$this->checkAndInitializePhpsec2() && !$this->checkPhpsec3()) {
+        if (!class_exists(RSA::class)) {
             throw new RuntimeException('Please require phpseclib/phpseclib v2 or v3 to use this utility.');
         }
     }
 
     private function loadPhpsecPublicKey(string $modulus, string $exponent): string
     {
-        if (class_exists(RSA::class) && class_exists(BigInteger2::class)) {
```
vishwarajanand, line 404: These classes will never exist now. Hence removed this check.

src/ServiceAccountSignerTrait.php
```diff
@@ -37,11 +38,9 @@ public function signBlob($stringToSign, $forceOpenssl = false)
...
 
         $signedString = '';
-        if (class_exists('\\phpseclib\\Crypt\\RSA') && !$forceOpenssl) {
-            $rsa = new RSA();
-            $rsa->loadKey($privateKey);
-            $rsa->setSignatureMode(RSA::SIGNATURE_PKCS1);
-            $rsa->setHash('sha256');
+        if (class_exists(phpseclib3\Crypt\RSA::class) && !$forceOpenssl) {
+            $key = PublicKeyLoader::load($privateKey);
+            $rsa = $key->withHash('sha256')->withPadding(RSA::SIGNATURE_PKCS1);
```
vishwarajanand, line 43: The methods are sourced from here: [PrivateKey.php#L447-L452](https://github.com/phpseclib/phpseclib/blob/8516c99f9e4261c634a59c2c28ddfaa8d15726fb/phpseclib/Crypt/RSA/PrivateKey.php#L447-L452) and [PrivateKey.php#L281](https://github.com/phpseclib/phpseclib/blob/8516c99f9e4261c634a59c2c28ddfaa8d15726fb/phpseclib/Crypt/RSA/PrivateKey.php#L281)

yash30201:
Just a doubt, why is phpseclib listed as a dev dependency?

https://github.com/googleapis/google-auth-library-php/blob/main/composer.json#L25

bshaffer:
> why is phpseclib listed as a dev dependency?
because it's only needed for verifying JWT tokens, which is not a core feature of this library. So the idea is that this dependency is optional, and should be instead defined by the calling library if that functionality is used.

Generally speaking, the fewer dependencies we define the better, so that we can avoid conflicts, security vulnerabilities, etc. So we left this one out.

src/AccessToken.php
```diff
@@ -395,7 +394,7 @@ private function retrieveCertsFromLocation($url, array $options = [])
      */
     private function checkAndInitializePhpsec()
     {
-        if (!$this->checkAndInitializePhpsec2() && !$this->checkPhpsec3()) {
+        if (!class_exists(RSA::class)) {
             throw new RuntimeException('Please require phpseclib/phpseclib v2 or v3 to use this utility.');
```
bshaffer, line 398: This needs to be updated
```suggestion
            throw new RuntimeException('Please require phpseclib/phpseclib v3 to use this utility.');
```

bshaffer (changes requested):

bshaffer (approved):

Issue #523, opened 2024-01-11
Add support for "executable" credential source


bshaffer:
See the [NodeJs implementation](https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/pluggable-auth-client.ts) for reference. 

Support the "executable" credential source for external account credentials. For example:

```json
{
  "type": "external_account", 
  "audience": "//iam.googleapis.com/locations/global/workforcePools/$WF_POOL_ID/providers/$PROVIDER_ID",
  "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
  "token_url": "https://sts.googleapis.com/v1/token",
  "token_info_url": "https://sts.googleapis.com/v1alpha/token_info", 
  "credential_source": {
    "executable": {
      "command": "/path/to/get/credentials.sh --arg1=value1 --arg2=value2",
      "timeout_millis": 1500,
      "interactive_timeout_millis": 30000,
      "output_file": "/path/to/generated/cached/credentials/"
    }
  }
}
```

bshaffer:
This was added in #525

PR #525, opened 2024-01-18
feat: add ExecutableSource credentials

bshaffer:
addresses https://github.com/googleapis/google-auth-library-php/issues/523

### DIFF
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -122,7 +122,7 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
         if (
             $this->outputFile
             && file_exists($this->outputFile)
-            && !empty(trim($outputFileContents = file_get_contents($this->outputFile)))
+            && !empty(trim($outputFileContents = (string) file_get_contents($this->outputFile)))
         ) {
             try {
                 $executableResponse = $this->parseExecutableResponse($outputFileContents);
@@ -152,7 +152,7 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
                 throw new ExecutableResponseError(
                     'The executable failed to run'
                     . ($output ? ' with the following error: ' . $output : '.'),
-                    $exitCode
+                    (string) $exitCode
                 );
             }
 
@@ -175,6 +175,9 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
             : $executableResponse['id_token'];
     }
 
+    /**
+     * @return array<string, mixed>
+     */
     private function parseExecutableResponse(string $responseJson): array
     {
         $executableResponse = json_decode($responseJson, true);
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -114,9 +114,16 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
             );
         }
 
-        if ($this->outputFile && file_exists($this->outputFile)) {
-            $outputFileContents = file_get_contents($this->outputFile) ?: '';
-            return $this->parseTokenFromResponse($outputFileContents);
+        if (
+            $this->outputFile
+            && file_exists($this->outputFile)
+            && ($outputFileContents = file_get_contents($this->outputFile))
+            && !empty($outputFileContents)
+        ) {
+            $json = json_decode($outputFileContents, true);
+            if (isset($json['expiration_time']) && time() < $json['expiration_time']) {
+                return $this->parseTokenFromResponse($outputFileContents);
+            }
         }
 
         // Run the executable.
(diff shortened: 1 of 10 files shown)


bshaffer:
@aeitzman

src/CredentialSource/ExecutableSource.php
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        if (
+            $this->outputFile
+            && file_exists($this->outputFile)
+            && !empty($outputFileContents = file_get_contents($this->outputFile))
+        ) {
+            $json = json_decode($outputFileContents, true);
+            if (isset($json['expiration_time']) && time() < $json['expiration_time']) {
+                return $this->parseTokenFromResponse($outputFileContents);
```
aeitzman, line 124: nit: If there is an error here, the messages won't specify that it comes from reading an output file vs running the executable directly right? Might be helpful to differentiate that for the user.

aeitzman: Also, if the file response is valid Json but unsuccessful what happens here? In that case we want to re-run the executable.

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -150,11 +152,6 @@ private static function buildCredentialSource(array $jsonKey): ExternalAccountCr
                     'The regional_cred_verification_url field is required for aws1 credential source.'
                 );
             }
-            if (!array_key_exists('audience', $jsonKey)) {
```
aeitzman, line 153: was this intentionally removed?

src/ExecutableHandler/ExecutableHandler.php
```diff
@@ -0,0 +1,75 @@
+<?php
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
aeitzman, line 3:
```suggestion
 * Copyright 2024 Google Inc. All Rights Reserved.
```

src/ExecutableHandler/ExecutableHandler.php
```diff
@@ -0,0 +1,75 @@
...
+    /**
+     * @param string $command
+     * @return int
+     */
+    public function __invoke(string $command): int
+    {
+        $process = Process::fromShellCommandline(
+            $command,
+            null,
+            $this->env,
```
aeitzman, line 59: Does this completely replace the Env variables or does it just add?

bshaffer: It just adds to them!

src/CredentialSource/ExecutableSource.php
bshaffer: > If there is an error here, the messages won't specify that it comes from reading an output file vs running the executable directly right?

That's correct. It should be easy to add that

> if the file response is valid Json but unsuccessful what happens here?

Good question, I believe it throws an error same as if it's the response. Are you suggesting that any errors in the file with `success: false` should silently retry the executable?

Is there a good document that lists the expected logic for the `outputFile`, because it's not very intuitive. I followed the [NodeJS implementation](https://github.com/googleapis/google-auth-library-nodejs/blob/main/src/auth/pluggable-auth-client.ts), although I may have missed some things.

src/Credentials/ExternalAccountCredentials.php
bshaffer: Yes because we already verify that the "audience" key exists [earlier in this function](https://github.com/googleapis/google-auth-library-php/pull/525/files#diff-2826fe12b39f9080278b434e47cda5ff572f3e846b3f1d4e043feb39e91f9c5fR83-R87)

src/CredentialSource/ExecutableSource.php
aeitzman: go/pluggable-auth-design but it doesn't look like that has a great flow diagram. But yes, the expected behavior is that if the output file is parseable (I.e. good JSON) but success:false, then we would re-run the executable instead of just returning the error again

aeitzman (approved):

src/CredentialSource/ExecutableSource.php
bshaffer: I have manually run the tests in go/pluggable-auth-sdk-testing and they are now passing

tests/CredentialSource/ExecutableSourceTest.php
```diff
@@ -0,0 +1,328 @@
...
+        $outputFile = tempnam(sys_get_temp_dir(), 'token');
+        file_put_contents($outputFile, json_encode($cachedToken));
+
+        $executableHandler = $this->prophesize(ExecutableHandler::class);
+        $executableHandler->__invoke('fake-command')
+            ->shouldBeCalledOnce()
+            ->willReturn(0);
+        $executableHandler->getOutput()
+            ->shouldBeCalledOnce()
+            ->willReturn(json_encode($successToken));
```
yash30201, line 292: This code is being used thrice in this file, we can just export them to a private method the reuse this.

yash30201 (approved):
The PR looks good to me from PHP side, just a small nit regarding the tests : )

Great work @bshaffer!

tests/CredentialSource/ExecutableSourceTest.php
bshaffer: This is definitely a matter of preference, but I'd prefer to keep each test defining it's own mocks, rather than abstract parts out to custom functions. The way it is now is easier to follow IMHO

PR #532, opened 2024-02-13
chore: drop support for PHP 8.0

bshaffer:
- Remove support for `Google\Auth\Cache\Item`, which is no longer compatible with these dependencies
 - Drop compatibility for PHP 7.4 and 8.0
 - Drop compatibility with `guzzlehttp/guzzle` v6, and `psr/cache` v1


vishwarajanand:
I added some fixes as part of d6eba00c6daa7f67868d4b74807400818fa3f07c to fix the style errors

vishwarajanand (approved):
I think the changes in this PR are good to go. 
I don't understand when `tests.yml` specifically removed php<=v8.0 why does the PR still mark their tests as expected to be reported.

bshaffer:
> why does the PR still mark their tests as expected to be reported (and in-fact required).

Because this is managed in the GitHub repo settings (the yaml file syncs those settings, but only once they're in `main`). So they need to be changed there first.

bshaffer:
I've added the `Do Not Merge` label because we are going to postpone dropping support for PHP 8.0 until May.

We will drop support for PHP 7.4 on Feb 29th. See https://github.com/googleapis/google-auth-library-php/pull/538

PR #533, opened 2024-02-16
feat: add caching for universe domain

bshaffer:
- Allows `FetchAuthTokenCache` to cache the universe domain. 
 - Enables this cache for `GCECredentials` by default, in order to save a network call.

### DIFF
--- a/src/FetchAuthTokenCache.php
+++ b/src/FetchAuthTokenCache.php
@@ -58,6 +58,7 @@ class FetchAuthTokenCache implements
         $this->cacheConfig = array_merge([
             'lifetime' => 1500,
             'prefix' => '',
+            'cacheUniverseDomain' => $fetcher instanceof Credentials\GCECredentials,
         ], (array) $cacheConfig);
     }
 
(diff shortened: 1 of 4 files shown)


src/FetchAuthTokenCache.php
```diff
@@ -320,4 +324,16 @@ private function saveAuthTokenInCache($authToken, $authUri = null)
             $this->setCachedValue($cacheKey, $authToken);
         }
     }
+
+    private function getCachedUniverseDomain(GetUniverseDomainInterface $fetcher): string
+    {
+        $cacheKey = $this->getFullCacheKey($fetcher->getCacheKey() . 'universe_domain');
```
vishwarajanand, line 330: To get past this error, you can add a method `GetUniverseDomainInterface::getCacheKey` which will be a not a very neat method IMO
OR
use a hardcoded cache key like `cachedUniverseDomain`.

I see python seems to handling this via a boolean flag, [code pointer here](https://github.com/googleapis/google-auth-library-python/blob/32ed6015ea71f9dd1a594b3caa88fd328f56b0b7/google/auth/compute_engine/credentials.py#L88)

Also, I find implementations of `getUniverseDomain` are rather accessing static variables (except `GCECredentials::getUniverseDomain`) where its also cached at the object instance level, [code pointer here](https://github.com/googleapis/google-auth-library-php/blob/6e9c9fd4e2bbd7042f50083076346e4a1eff4e4b/src/Credentials/GCECredentials.php#L541-L542). So, maybe using an actual cache is an overkill, unless I am missing something.

vishwarajanand (approved):
Lgtm

Issue #498, opened 2023-11-28
Add default file caching


bshaffer:
For any company running at scale, a default file cache is pretty much a requirement to ensure that the client libraries and auth library function in a performant way. Creating a file cache which works out of the box (and by default) would be a huge help to our customers which are running these libraries at scale.

Rationale: customers have had to create wrappers which implement their custom caching, and this has caused headache, when the configuration for these caches should be intuitive enough and easy enough to implement that they can happen out of the box without handholding or wrapper classes.

For examples of existing FileCache, see [php-cache/filesystem-adapter](https://github.com/php-cache/filesystem-adapter), which uses the [`FilesystemCachePool`](https://github.com/php-cache/filesystem-adapter/blob/master/FilesystemCachePool.php), and is implemented in the [BaseTest](https://github.com/googleapis/google-api-php-client/blob/e70273c06d18824de77e114247ae3102f8aec64d/tests/BaseTest.php#L52) class of our Apiary client library.

 - Follow the [PHP-FIG PSR-6 caching interfaces](https://www.php-fig.org/psr/psr-6/)
 - Make sure that the [`getCacheKey` methods](https://github.com/googleapis/google-auth-library-php/blob/23b1fc1d8275f5c81a3e11d0c680ddabca308cf9/src/Credentials/ServiceAccountCredentials.php#L215) are unique and safe
 - Verify with testing and benchmarks that a PHP application receiving multiple requests and handling multiple processes will handle them in a scalable way (e.g Metadata Server credentials, JWT credentials, etc).

PR #551, opened 2024-05-07
Add a prototype for the file system cache

Hectorhammett:
This is a prototype for #498

### DIFF
--- a/src/Cache/CacheChain.php
+++ b/src/Cache/CacheChain.php
@@ -1,0 +1,153 @@
+<?php
+/**
+ * Copyright 2024 Google Inc. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ namespace Google\Auth\Cache;
+
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
+ class CacheChain implements CacheItemPoolInterface
+ {
+    private array $pools = [];
+
+    /**
+     * @param CacheItemPoolInterface[] $pools 
+     */
+    public function __construct(array $pools)
+    {
+        foreach($pools as $pool) {
+            $this->addPool($pool);
+        }
+    }
+
+    public function getItem(string $key): CacheItemInterface
+    {
+        foreach($this->pools as $pool) {
+            $item = $pool->getItem($key);
+            if ($item->isHit()) {
+                return $item;
+            }
+        }
+
+        return new TypedItem($$key);
+    }
+
+    public function getItems(array $keys = []): iterable
+    {
+        $result = [];
+
+        foreach($keys as $key) {
+            $result[$key] = $this->getItem($key);
+        }
+
+        return $result;
+    }
+
+    public function hasItem(string $key): bool
+    {
+        foreach($this->pools as $pool) {
+            if($pool->hasItem($key)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function clear(): bool
+    {
+        $result = true;
+
+        foreach($this->pools as $pool) {
+            if(!$pool->clear()) {
+                $result = false;
+            }
+        }
+
+        return $result;
+    }
+
+    public function deleteItem(string $key): bool
+    {
+        $result = true;
+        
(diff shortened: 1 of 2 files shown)


src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public function hasItem(string $key): bool
+    {
+        $itemPath = $this->cacheFilePath($key);
+
+        if (!file_exists($itemPath)) {
+            return false;
+        }
+
+        $serializedItem = file_get_contents($this->cacheFilePath($key));
```
yash30201, line 89: We can reuse the calculated path.

```suggestion
        $serializedItem = file_get_contents($itemPath);
```

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public function deleteItems(array $keys): bool
+    {
+        $result = true;
+
+        foreach ($keys as $key) {
+            if (!$this->isValidKey($key)) {
+                throw new InvalidArgumentException("The key '$key' is not valid. The key should follow the pattern |^[a-zA-Z0-9_\.! ]+$|");
+            }
+
```
yash30201, line 148: `deleteItem` method is already doing this check. 

```suggestion
```

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,212 @@
...
+    public function __construct($options = [])
+    {
+        if (array_key_exists('path', $options)) {
+            $this->cachePath = $options['path'];
+        }
+
+        if (is_dir($this->cachePath)) {
+            return true;
+        }
```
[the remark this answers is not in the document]
Hectorhammett: Oh my. Yeah haha I basically want to handle the case where the cache folder already exists. But returning true from a `__construct` is definitely a boo boo haha. 

In that topic, what do we think it would be the best option when a cache folder already exists? Just re use it? Raise an exception? Clear it?

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,212 @@
...
+     */
+    public function getItems(array $keys = []): iterable
+    {
+        $result = [];
+
+        foreach ($keys as $key) {
+            if ($this->isValidKey($key)){
+                throw new InvalidArgumentException("The key '$key' is not valid. The key should follow the pattern |^[a-zA-Z0-9_\.! ]+$|");
+            }
+
```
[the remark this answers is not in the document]
Hectorhammett: True! Good catch!

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        if (!mkdir($this->cachePath)) {
+            throw new ErrorException("Cache folde couldn't be created");
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
```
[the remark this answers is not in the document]
Hectorhammett: Yeah absolutely.

We should add more documentation on the final release.

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,212 @@
...
+    public function clear(): bool
+    {
+        if (!is_dir($this->cachePath)) {
+            return false;
+        }
+
+        foreach (scandir($this->cachePath) as $fileName) {
```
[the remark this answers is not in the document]
Hectorhammett: Absolutely.

I wonder if we should utilize the `error_reporting` function as the `psr6` spec does not specifies this function throwing exceptions. 

Besides another characteristic of caching is the fact that the cache can fail and the program should not stop working, so my thought is using some sort of logging to report the problem. 

But we definitely need to catch specific errors.

src/Cache/FileSystemCacheItemPool.php
Hectorhammett: image

src/Cache/CacheChain.php
```diff
@@ -0,0 +1,153 @@
...
+        foreach($this->pools as $pool) {
+            if(!$pool->save($item)) {
+                $result = false;
+            }
+        }
+
+        return $result;
+    }
+
+    public function saveDeferred(CacheItemInterface $item): bool
```
Hectorhammett, line 123: This method could cause issues hogging too many resources as every pool will use their own implementation for deferring an item save.

An option would be to implement a buffer on this class and call `save()` on each buffer on `commit()` to avoid exploding the resources.

src/Cache/CacheChain.php
```diff
@@ -0,0 +1,153 @@
...
+    {
+        $result = true;
+
+        foreach($this->pools as $pool) {
+            if(!$pool->clear()) {
+                $result = false;
+            }
+        }
+
+        return $result;
```
Hectorhammett, line 81: I am using the description for `psr6` where they mentioned on these `bool` results to return false if there is any error. 

This means that the user of this CacheChain will be receiving `false` when we cleared or deleted items on some caches and not all of them. 

I wonder if this would create a bug/issue.

Issue #559, opened 2024-06-03
Audit Cache Keys to determine if they are unique enough while caching auth tokens


Hectorhammett:
Currently we have limited support to Caching auth tokens. We are currently planning to make improvements on this regard and in order to accomplish this we need to ensure that we are generating keys that are unique enough. 

Tokens to be checked:
- [x] AppIdentityCredentials - Deprecated.
- [x] ExternalAccountCredentials
- [x] GCECredentials
- [x] IAMCredentials
- [x] ImpersonatedServiceAccountCredentials
- [x] InsecureCredentials
- [x] ServiceAccountCredentials
- [x] ServiceAccountJwtAccessCredentials
- [x] UserRefreshCredentials

PR #560, opened 2024-06-06
feat: Change getCacheKey implementation for more unique keys

Hectorhammett:
We made changes to how we calculate the `cacheKeys` in order to avoid collision for the oncoming support for the Caching update.

closes: #559

### DIFF
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -329,6 +329,22 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
     }
 
     /**
+     * Gets the unique key for caching
+     * For AwsNativeSource the values are:
+     * imdsv2SessionTokenUrl:securityCredentialsUrl:regionUrl:regionalCredVerificationUrl:audience
+     *
+     * @return string
+     */
+    public function getCacheKey(): string
+    {
+        return $this->imdsv2SessionTokenUrl . ':'
+            . $this->securityCredentialsUrl . ':'
+            . $this->regionUrl . ':'
+            . $this->regionalCredVerificationUrl . ':'
+            . $this->audience;
+    }
+
+    /**
      * Return HMAC hash in binary string
      */
     private static function hmacSign(string $key, string $msg): string
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -331,7 +331,7 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
     /**
      * Gets the unique key for caching
      * For AwsNativeSource the values are:
-     * imdsv2SessionTokenUrl.securityCredentialsUrl.regionUrl.regionalCredVerificationUrl
+     * Imdsv2SessionTokenUrl.SecurityCredentialsUrl.RegionUrl.RegionalCredVerificationUrl
      *
      * @return string
      */
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -331,7 +331,7 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
     /**
      * Gets the unique key for caching
      * For AwsNativeSource the values are:
-     * imdsv2SessionTokenUrl:securityCredentialsUrl:regionUrl:regionalCredVerificationUrl:audience
+     * <imdsv2SessionTokenUrl><securityCredentialsUrl><regionUrl><regionalCredVerificationUrl>
      *
      * @return string
      */
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -335,13 +335,12 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
      *
      * @return string
      */
-    public function getCacheKey(): ?string
+    public function getCacheKey(): string
     {
-        return $this->imdsv2SessionTokenUrl . ':'
-            . $this->securityCredentialsUrl . ':'
-            . $this->regionUrl . ':'
-            . $this->regionalCredVerificationUrl . ':'
-            . $this->audience;
+        return $this->imdsv2SessionTokenUrl .
+            $this->securityCredentialsUrl .
+            $this->regionUrl .
+            $this->regionalCredVerificationUrl;
     }
 
     /**
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -337,8 +337,8 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
      */
     public function getCacheKey(): string
     {
-        return ($this->imdsv2SessionTokenUrl ? $this->imdsv2SessionTokenUrl : '') .
-            '.' . ($this->securityCredentialsUrl ? $this->securityCredentialsUrl : '') .
+        return ($this->imdsv2SessionTokenUrl ?? '') .
+            '.' . ($this->securityCredentialsUrl ?? '') .
             '.' . $this->regionUrl .
             '.' . $this->regionalCredVerificationUrl;
     }
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -331,16 +331,16 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
     /**
      * Gets the unique key for caching
      * For AwsNativeSource the values are:
-     * <imdsv2SessionTokenUrl><securityCredentialsUrl><regionUrl><regionalCredVerificationUrl>
+     * imdsv2SessionTokenUrl.securityCredentialsUrl.regionUrl.regionalCredVerificationUrl
      *
      * @return string
      */
     public function getCacheKey(): string
     {
-        return $this->imdsv2SessionTokenUrl .
-            $this->securityCredentialsUrl .
-            $this->regionUrl .
-            $this->regionalCredVerificationUrl;
+        return ($this->imdsv2SessionTokenUrl ? $this->imdsv2SessionTokenUrl : '') .
+            '.' . ($this->securityCredentialsUrl ? $this->securityCredentialsUrl : '') .
+            '.' . $this->regionUrl .
+            '.' . $this->regionalCredVerificationUrl;
     }
 
     /**
--- a/src/CredentialSource/AwsNativeSource.php
+++ b/src/CredentialSource/AwsNativeSource.php
@@ -335,7 +335,7 @@ class AwsNativeSource implements ExternalAccountCredentialSourceInterface
      *
      * @return string
      */
-    public function getCacheKey(): string
+    public function getCacheKey(): ?string
     {
         return $this->imdsv2SessionTokenUrl . ':'
             . $this->securityCredentialsUrl . ':'
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -104,9 +104,9 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
      * Gets the unique key for caching
      * This source is considered non cacheable so returns null.
      *
-     * @return null
+     * @return null|string
      */
-    public function getCacheKey(): null
+    public function getCacheKey(): null|string
     {
         return null;
     }
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -108,7 +108,7 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
      */
     public function getCacheKey(): ?string
     {
-        return null;
+        return $this->command . $this->outputFile;
     }
 
     /**
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -102,7 +102,7 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
 
     /**
      * Gets the unique key for caching
-     * This source is considered non cacheable so returns null.
+     * The format for this source is <command><outputFile>
      *
      * @return ?string
      */
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -102,13 +102,13 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
 
     /**
      * Gets the unique key for caching
-     * The format for this source is <command><outputFile>
+     * The format for this source is command.outputFile
      *
      * @return ?string
      */
     public function getCacheKey(): ?string
     {
-        return $this->command . $this->outputFile;
+        return $this->command . '.' . $this->outputFile;
     }
 
     /**
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -101,6 +101,17 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
     }
 
     /**
+     * Gets the unique key for caching
+     * This source is considered non cacheable so returns null.
+     *
+     * @return null
+     */
+    public function getCacheKey(): null
+    {
+        return null;
+    }
+
+    /**
      * @param callable $httpHandler unused.
      * @return string
      * @throws RuntimeException if the executable is not allowed to run.
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -102,7 +102,8 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
 
     /**
      * Gets the unique key for caching
-     * The format for this source is command.outputFile
+     * The format for the cache key is:
+     * Command.OutputFile
      *
      * @return ?string
      */
--- a/src/CredentialSource/ExecutableSource.php
+++ b/src/CredentialSource/ExecutableSource.php
@@ -104,9 +104,9 @@ class ExecutableSource implements ExternalAccountCredentialSourceInterface
      * Gets the unique key for caching
      * This source is considered non cacheable so returns null.
      *
-     * @return null|string
+     * @return ?string
      */
-    public function getCacheKey(): null|string
+    public function getCacheKey(): ?string
     {
         return null;
     }
--- a/src/CredentialSource/UrlSource.php
+++ b/src/CredentialSource/UrlSource.php
@@ -94,4 +94,16 @@ class UrlSource implements ExternalAccountCredentialSourceInterface
 
         return $body;
     }
+
+    /**
+     * Get the cache key for the credentials.
+     * The format for the cache key is:
+     * URL + subjectTokenFieldName
+     *
+     * @return string
+     */
+    public function getCacheKey(): string
+    {
+        return $this->url . ":" . $this->subjectTokenFieldName;
+    }
 }
--- a/src/CredentialSource/UrlSource.php
+++ b/src/CredentialSource/UrlSource.php
@@ -100,10 +100,10 @@ class UrlSource implements ExternalAccountCredentialSourceInterface
      * The format for the cache key is:
      * URL + subjectTokenFieldName
      *
-     * @return string
+     * @return ?string
      */
-    public function getCacheKey(): string
+    public function getCacheKey(): ?string
     {
-        return $this->url . ':' . $this->subjectTokenFieldName;
+        return $this->url;
     }
 }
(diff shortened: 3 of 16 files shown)


src/Credentials/ExternalAccountCredentials.php
```diff
@@ -122,6 +129,8 @@ public function __construct(
                 'workforce_pool_user_project should not be set for non-workforce pool credentials.'
             );
         }
+
+        $this->jsonKey = $jsonKey;
```
bshaffer, line 133: I was thinking we should add a `getCacheKey` to the Credential Source classes (see `src/CredentialSource`), and use that instead. Then the credentials source classes can determine how to ensure their keys are "unique":

Something like:

```php
    public function getCacheKey()
    {
        return $this->auth->getSubjectTokenFetcher()->getCacheKey();
    }
```

There could be more to it as well, but this would be better than flattening the array IMHO

src/Credentials/GCECredentials.php
```diff
@@ -493,7 +493,7 @@ public function fetchAuthToken(callable $httpHandler = null)
      */
     public function getCacheKey()
     {
-        return self::cacheKey;
+        return $this->tokenUri;
```
bshaffer, line 496: Also add `/** @deprecated */` to the `cacheKey` constant

src/OAuth2.php
```diff
@@ -703,6 +703,26 @@ public function getCacheKey()
         return null;
     }
 
+    /**
+     * Obtains the scope or the Audience and formats it for external classes to use
+     * in generating a cache key.
+     *
+     * @return ?string a key that may be used to cache the auth token.
+     */
+    public function getFormattedScopeOrAudience()
```
bshaffer, line 712: Since this is a utility class, I'd prefer to not expose a public method like this on it (which is really an internal helper method). 

What if we added this method to `CacheTrait`, made it private, and made it accept `$scope` and `$audience` parameters.  Then we'd call it like

```php
return $this->auth->getClientId() . ':' . $this->getFormattedScopeOrAudience($this->auth->getScope(), $this->auth->getAudience());
```

Or even skip this method all together and just have the cache keys look like this (simpler):

```php
return $this->auth->getClientId() . ':' . implode(':', $this->auth->getScope()) . ':' . $this->auth->getAudience();
```

bshaffer (commented):
Good job! I think we can improve a few things. Additionally, I'd like to see PHPDoc comments for the `getCacheKey` method which explains which fields are included in the cache key.

src/Credentials/GCECredentials.php
Hectorhammett: Ahh I knew we needed to do something about it, but wasn't sure what. Let's add it.

src/OAuth2.php
Hectorhammett: Yeah I was also on the fence with this one, don't think formatting inside `OAuth` is a responsability of that class.

Good idea, let me change it.

Hectorhammett: On a second thought, while reviewing this, none of the credential classes uses the `CacheTrait` and it has some methods that I don't believe belong on these classes.

For example: `getCachedValue` and `setCacheValue`. These methods are currently being used by the `fetchAuthTokenCache`. I mention that these methods shouldn't be on the credentials themselves but I should stress they shouldn't be on this specific implementation as in V2 would be a different structure. But I digress.

I am not a fan of writing explicit code over and over, it always tingles my "This should be abstracted" sense, perhaps another trait? A utility static class?. That being said I also don't think is a do or die situation in this case.

bshaffer: I was worried about that (wrt the `CacheTrait`). In V2 all these classes will implement `CacheTrait`, and we'll get rid of `FetchAuthTokenCache` (although we _could_ that that now, without breaking BC, and just have `FetchAuthTokenCache` be a noop.... but I digress!)

At the end of the day it's a matter of preference. You could put it in `CredentialsLoader`, as all the credentials  in question extend that?  

Generally speaking, for simple things such as concatenating strings as we are doing here, which aren't SUPER important to keep consistent, I lean towards code repetition. To me this is clearer and easier to read than an abstraction / redirection in the code. But if we can cleanly add dryness as well (which I think we can with `CredentialsLoader`), then I'm okay with it.

Hectorhammett: I agree with your statement of readability. `ExternalAccountCredentials` do not implement `CredentialsLoader` but we just had a discussion about how different that one is that all the `sources` classes are doing their own implementation so we could add it to the `CredentialsLoader`.

Still, will go with the code repetition.

src/CredentialSource/AwsNativeSource.php
```diff
@@ -328,6 +328,22 @@ public static function getSigningVarsFromEnv(): ?array
...
     }
 
+    /**
+     * Gets the unique key for caching
+     * For AwsNativeSource the values are:
+     * imdsv2SessionTokenUrl:securityCredentialsUrl:regionUrl:regionalCredVerificationUrl:audience
+     *
+     * @return string
+     */
+    public function getCacheKey(): string
```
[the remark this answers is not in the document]
Hectorhammett: This is already on that class from this Commit:
https://github.com/googleapis/google-auth-library-php/pull/560/files/8415d958d3d67ea9d2a4526087197c1edf568662#diff-6a768888485c1e3d311e311c3343b005e2fc87535310e8cfe9f7f0acc9504282R23

Unless I am miss interpreting this comment?

src/CredentialSource/UrlSource.php
```diff
@@ -94,4 +94,16 @@ public function fetchSubjectToken(callable $httpHandler = null): string
...
+    /**
+     * Get the cache key for the credentials.
+     * The format for the cache key is:
+     * URL + subjectTokenFieldName
+     *
+     * @return string
+     */
+    public function getCacheKey(): string
+    {
+        return $this->url . ':' . $this->subjectTokenFieldName;
```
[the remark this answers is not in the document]
Hectorhammett: Well, if we really don't need it? The less code to maintain the better!

bshaffer:
@westarle We are looking to merge this soon, and would love you to take a look!

src/CredentialSource/AwsNativeSource.php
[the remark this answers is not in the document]
bshaffer: Oh no, that's great! I didn't realize it was already in there :P

src/CredentialSource/ExecutableSource.php
```diff
@@ -100,6 +100,17 @@ public function __construct(
         $this->executableHandler = $executableHandler ?: new ExecutableHandler();
     }
 
+    /**
+     * Gets the unique key for caching
+     * This source is considered non cacheable so returns null.
```
bshaffer, line 105: the comment says "This source is considered non cacheable so returns null", but yet `getCacheKey` is returning something. We need to either fix the comment or have it return `null`.

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -219,11 +219,20 @@ public function fetchAuthToken(callable $httpHandler = null)
...
      */
     public function getCacheKey()
     {
-        $key = $this->auth->getIssuer() . ':' . $this->auth->getCacheKey();
+        $scopeOrAudience = $this->auth->getScope();
+        if (!$scopeOrAudience) {
+            $scopeOrAudience = $this->auth->getAudience();
+        }
+
+        $key = $this->auth->getIssuer() . ':' . $scopeOrAudience;
```
bshaffer, line 235: Let's remove the use of colon, or replace with a `.`, since we would strip it later anyway. 
```suggestion
        $key = $this->auth->getIssuer() . '.' . $scopeOrAudience;
```

(and do this below as well)

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -219,11 +219,20 @@ public function fetchAuthToken(callable $httpHandler = null)
     }
 
     /**
+     * Return the Cache Key for the credentials.
+     * For the cache key format is:
+     * ClientEmail:Scope | Audience:sub.
```
bshaffer, line 224: Hmm, the pipe format is hard to read. How about this:
```suggestion
     * For the cache key format is one of the following:
     * ClientEmail.Scope[.Sub]
     * ClientEmail.Audience[.Sub]
```

src/Credentials/ServiceAccountJwtAccessCredentials.php
```diff
@@ -166,11 +166,20 @@ public function fetchAuthToken(callable $httpHandler = null)
     }
 
     /**
+     * Return the cache key for the credentials.
+     * The format for the Cache Key is:
+     * ClientEmail:Scope | Audience
```
bshaffer, line 171: Same here
```suggestion
     * The format for the Cache Key one of the following:
     * ClientEmail.Scope
     * ClientEmail.Audience
```

src/Credentials/ServiceAccountJwtAccessCredentials.php
```diff
@@ -166,11 +166,20 @@ public function fetchAuthToken(callable $httpHandler = null)
...
      */
     public function getCacheKey()
     {
-        return $this->auth->getCacheKey();
+        $scopeOrAudience = $this->auth->getScope();
+        if (!$scopeOrAudience) {
+            $scopeOrAudience = $this->auth->getAudience();
+        }
+
+        return $this->auth->getIssuer() . ':' . $scopeOrAudience;
```
bshaffer, line 182:
```suggestion
        return $this->auth->getIssuer() . '.' . $scopeOrAudience;
```

src/Credentials/UserRefreshCredentials.php
```diff
@@ -130,11 +130,20 @@ public function fetchAuthToken(callable $httpHandler = null, array $metricsHeade
     }
 
     /**
+     * Return the Cache Key for the credentials.
+     * The format for the Cache key is:
+     * ClientId:Scope | Audience
```
bshaffer, line 135: Same here
```suggestion
     * The format for the Cache key is one of the following:
     * ClientId.Scope
     * ClientId.Audience
```

src/Credentials/UserRefreshCredentials.php
```diff
@@ -130,11 +130,20 @@ public function fetchAuthToken(callable $httpHandler = null, array $metricsHeade
...
      */
     public function getCacheKey()
     {
-        return $this->auth->getClientId() . ':' . $this->auth->getCacheKey();
+        $scopeOrAudience = $this->auth->getScope();
+        if (!$scopeOrAudience) {
+            $scopeOrAudience = $this->auth->getAudience();
+        }
+
+        return $this->auth->getClientId() . ':' . $scopeOrAudience;
```
bshaffer, line 146:
```suggestion
        return $this->auth->getClientId() . '.' . $scopeOrAudience;
```

src/CredentialSource/AwsNativeSource.php
```diff
@@ -328,6 +328,21 @@ public static function getSigningVarsFromEnv(): ?array
...
+    {
+        return $this->imdsv2SessionTokenUrl .
+            $this->securityCredentialsUrl .
```
bshaffer, line 341: Since these first two can be `null`, we should only add them if they exist (similar to `sub` in `ServiceAccountCredentials::getCacheKey`)

Also, I think we should concatenate them with a `.` for consistency and readability.

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -276,9 +279,25 @@ public function fetchAuthToken(callable $httpHandler = null)
...
+        $scopeOrAudience = $this->auth->getAudience();
+        if (!$scopeOrAudience) {
+            $scopeOrAudience = $this->auth->getScope();
+        }
+
+        return $this->auth->getSubjectTokenFetcher()->getCacheKey() .
+            $scopeOrAudience .
+            $this->serviceAccountImpersonationUrl .
+            $this->auth->getSubjectTokenType() .
+            $this->workforcePoolUserProject;
```
bshaffer, line 300: since `serviceAccountImpersonationUrl` and `workforcePoolUserProject` can be `null`, we should only add them if they exist (similar to `sub` in `ServiceAccountCredentials::getCacheKey`)

Also, I think we should concatenate them with a `.` for consistency and readability.

bshaffer (commented):
nits and minor suggestions only!

src/CredentialSource/AwsNativeSource.php
Hectorhammett: You can concatenate null to a string in PHP and will not affect the result. If we want to do it just for being verbose? I have my reservations but I could see why:
```
$a = null;
$b = null;
$c = 'Hector';
$d = 'Mendoza';

$test = $a . $b . $c . $d;
print($test . "\n");
```
Result:
image

> Also, I think we should concatenate them with a . for consistency and readability.
My apologies, I do not understand this comment. To my understanding I am already concatenating everything with `.` unless I am misunderstanding something?

Hectorhammett: AH! Sorry, now I understand what you mean, we meant the character `.` not the operator `.` haha. I went ahead and removed all the `.` as we strip them anyway and keep everything consistent.

bshaffer: Yes! I meant let's concactinate with a `.`. This will also make the string easier to demonstrate in the PHPDoc.

Also, while you can concatinate `null`, and PHP will cast to an empty string, I'd prefer to not use implicit casting, and treat nulls as nulls and not as empty strings.

src/CredentialSource/ExecutableSource.php
```diff
@@ -102,13 +102,13 @@ public function __construct(
 
     /**
      * Gets the unique key for caching
-     * The format for this source is <command><outputFile>
+     * The format for this source is command.outputFile
```
bshaffer, line 105:
```suggestion
     * The format for this cache key is command.outputFile
```

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -282,7 +282,9 @@ public function fetchAuthToken(callable $httpHandler = null)
     /**
      * Get the cache token key for the credentials.
      * The cache token key format depends on the type of source
-     * was used to configure these credentials.
+     * The format for the Cache Key one of the following:
```
bshaffer, line 285:
```suggestion
     * The format for the cache key one of the following:
```

src/Credentials/ServiceAccountCredentials.php
```diff
@@ -221,8 +221,8 @@ public function fetchAuthToken(callable $httpHandler = null)
     /**
      * Return the Cache Key for the credentials.
      * For the cache key format is one of the following:
-     * <ClientEmail><Scope[.Sub]>
-     * <ClientEmail><Audience[.Sub]>
+     * ClientEmail.Scope[.Sub]
+     * ClientEmail.Audience[.Sub]
```
bshaffer, line 225: We should be consistent with our formatting with how we display these cache keys. Here we use `CamelCase`, but everywhere else uses `alligatorCase`. Also, here we have the optional values in `[]`, whereas in the others we don't.

I don't have a strong opinion on which formatting we use (although I like the looks of this one a little better), but we should be consistent.

Hectorhammett: You're right. I changed these formats so many times that I missed some updates, let me fix this.

src/CredentialSource/AwsNativeSource.php
```diff
@@ -328,6 +328,21 @@ public static function getSigningVarsFromEnv(): ?array
...
+     * Gets the unique key for caching
+     * For AwsNativeSource the values are:
+     * Imdsv2SessionTokenUrl.SecurityCredentialsUrl.RegionUrl.RegionalCredVerificationUrl
+     *
+     * @return string
+     */
+    public function getCacheKey(): string
+    {
+        return ($this->imdsv2SessionTokenUrl ? $this->imdsv2SessionTokenUrl : '') .
+            '.' . ($this->securityCredentialsUrl ? $this->securityCredentialsUrl : '') .
```
bshaffer, line 341: PHP trick - when you're dealing with `null`, you can use `??`, and when you're dealing with falsy, you can use `?:` to make this expression shorter
```suggestion
        return ($this->imdsv2SessionTokenUrl ?? '') .
            '.' . ($this->securityCredentialsUrl ?? '') .
```

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -276,9 +279,27 @@ public function fetchAuthToken(callable $httpHandler = null)
...
+        $scopeOrAudience = $this->auth->getAudience();
+        if (!$scopeOrAudience) {
+            $scopeOrAudience = $this->auth->getScope();
+        }
+
+        return $this->auth->getSubjectTokenFetcher()->getCacheKey() .
+            '.' . $scopeOrAudience .
+            $this->concatenateValueOrEmpty($this->serviceAccountImpersonationUrl) .
+            $this->concatenateValueOrEmpty($this->auth->getSubjectTokenType()) .
+            $this->concatenateValueOrEmpty($this->workforcePoolUserProject);
```
bshaffer, line 302: with consistency again - I think doing it this way is fine, as is leaving the `.` in and using `??` (as above), but we need to be consistent. 

My preference is to keep it simple and use `??` as above, so that we don't need the extra `concatenateValueOrEmpty` function. 
```suggestion
            '.' . ($this->serviceAccountImpersonationUrl ?? '') .
            '.' . ($this->auth->getSubjectTokenType() ?? '') .
            '.' . ($this->workforcePoolUserProject ?? '');
```

src/OAuth2.php
```diff
@@ -703,6 +705,16 @@ public function getCacheKey()
         return null;
     }
 
+    /**
+     * Gets this instance's SubjectTokenFetcher
+     *
+     * @return null|ExternalAccountCredentialSourceInterface
+     */
+    public function getSubjectTokenFetcher(): null|ExternalAccountCredentialSourceInterface
```
bshaffer, line 713: It's more standard to write this like this:
```suggestion
    public function getSubjectTokenFetcher(): ?ExternalAccountCredentialSourceInterface
```

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -100,6 +100,9 @@ public function __construct(
 
         if (array_key_exists('service_account_impersonation_url', $jsonKey)) {
             $this->serviceAccountImpersonationUrl = $jsonKey['service_account_impersonation_url'];
+        } else {
+            //If we do not initalize this, getCacheKey throws an error.
+            $this->serviceAccountImpersonationUrl = null;
         }
```
bshaffer, line 106: more concise!
```suggestion
        $this->serviceAccountImpersonationUrl = $jsonKey['service_account_impersonation_url'] ?? null
```

Hectorhammett: I have never been fan of this syntax haha. But is just an opinionated matter. Let me add it.

src/Credentials/ExternalAccountCredentials.php
Hectorhammett: I did it to keep the `null` value in mind and not have a cache key like:
```
a.b...
```
We do remove the dots though, but still there was a previous comment about handling the nulls different and I decided to not add the random dots. What do you think?

bshaffer: My point was that we want it to be consistent, as you were adding the dots already in AwsNativeSource

src/Credentials/ExternalAccountCredentials.php
bshaffer: Going from 5 lines to 1 is more readable IMHO

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -276,9 +274,27 @@ public function fetchAuthToken(callable $httpHandler = null)
         return $stsToken;
     }
 
-    public function getCacheKey()
+    /**
+     * Get the cache token key for the credentials.
+     * The cache token key format depends on the type of source
+     * The format for the cache key one of the following:
+     * FetcherCacheKey.Scope.[ServiceAccount].[TokenType].[WorkforcePoolUserProject]
+     * FetcherCacheKey.Audience.[ServiceAccount].[TokenType].[WorkforcePoolUserProject]
```
bshaffer, line 282:
```suggestion
     * The format for the cache key is:
     * FetcherCacheKey.[Scope|Audience].[TokenType].[WorkforcePoolUserProject]
```

bshaffer (approved):

src/CredentialSource/AwsNativeSource.php
```diff
@@ -328,6 +328,21 @@ public static function getSigningVarsFromEnv(): ?array
...
+     * Imdsv2SessionTokenUrl.SecurityCredentialsUrl.RegionUrl.RegionalCredVerificationUrl
+     *
+     * @return string
+     */
+    public function getCacheKey(): string
+    {
+        return ($this->imdsv2SessionTokenUrl ?? '') .
+            '.' . ($this->securityCredentialsUrl ?? '') .
+            '.' . $this->regionUrl .
+            '.' . $this->regionalCredVerificationUrl;
```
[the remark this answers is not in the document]
Hectorhammett: Brent, you a genius, I do like that better haha.

PR #569, opened 2024-07-17
chore: drop deprecated Item class

artemkurylo:
Drop deprecated `Item` class.

This class isn't used anymore and was deprecated along with drop of PHP 7.4. As a destructive example of it's presence, one of the projects that I'm working on tries to load every file from vendor directory into the memory and this class breaks it, because it doesn't comply with `CacheItemInterface` interface (missing return types, which results in a fatal error).


bshaffer:
As long as we support `psr/log: 2`, we need these item classes

https://github.com/googleapis/google-auth-library-php/blob/1277062739eedd04e7d0af86f6fe1853d2620706/composer.json#L17

PR #570, opened 2024-08-07
Debug Logging prototype

Hectorhammett:


src/Logger/LoggingTrait.php
```diff
@@ -0,0 +1,104 @@
...
+        $token = str_replace('Bearer ', '', $tokenHeader);
+
+        if (substr_count($token, '.') !== 2) {
+            return null;
+        }
+
+        [$header, $token, $_] = explode('.', $token);
+
+        return [
+            'jwt' => [
```
bshaffer, line 98: is this correct? It seems like this should _just_ return `header` and `payload`, especially since it is already added to an array with the array key `jwt`.

src/Logger/LoggingTrait.php
```diff
@@ -0,0 +1,104 @@
...
+            'severity' => 'DEBUG', //Perhaps have something like Logger::debug
+        ];
+
+        $jsonPayload = [
+            'request.method' => $event->method,
+            'request.url' => $event->url,
+            'request.headers' => $event->headers,
+            'request.payload' => $event->payload,
+            'request.rpcName' => $event->rpcName,
+            'jwt' => $this->getJwtToken($event->headers),
```
bshaffer, line 42: the way this is defined here, it will look like this in the payload:

```php
[
    'jwt' => [
        'jwt' => [
            'header' => [ /** ... */ ],
            'payload' => [ /** ... */ ],
        ]
    ]
]
```

is that intentional?

src/HttpHandler/Guzzle6HttpHandler.php
```diff
@@ -57,6 +68,37 @@ public function __invoke(RequestInterface $request, array $options = [])
...
+            $requestEvent->payload = $request->getBody()->getContents();
+            $requestEvent->retryAttempt = $options['retryAttempt'] ?? null;
+            $requestEvent->serviceName = $options['serviceName'];
+            $requestEvent->clientId = $options['clientId'];
+            $requestEvent->requestId = spl_object_id($request);
+
+            $this->logRequest($requestEvent);
+        }
+
+        return $this->client->sendAsync($request, $options)->then(function (ResponseInterface $response) use ($requestEvent) {
```
bshaffer, line 88: nit, but this could be attached only when the logger is defined, which would look a little more cleaner. e.g.

```php
$promise = $this->client->sendAsync($request, $options);
if ($this->logger) {
    $promise->then( /** ... */);
}
return $promise;
```

src/ApplicationDefaultCredentials.php
```diff
@@ -353,4 +357,21 @@ private static function onGce(
...
+
+    /**
+     * A function that returns the default logger if the GOOGLE_SDK_DEBUG_LOGGING
+     * environment variable is set. Returns null if not.
+     *
+     * @return null|LoggerInterface
+     */
+    public static function getDefaultLogger(): null|LoggerInterface
+    {
+        $loggingFlag = (string)getenv(self::SDK_DEBUG_FLAG);
```
bshaffer, line 369: nit
```suggestion
        $loggingFlag = (string) getenv(self::SDK_DEBUG_FLAG);
```

I've added a CS rule for this in https://github.com/GoogleCloudPlatform/php-tools/pull/149

src/HttpHandler/HttpHandlerFactory.php
```diff
@@ -52,11 +55,15 @@ public static function build(ClientInterface $client = null)
             $version = (int) substr(ClientInterface::VERSION, 0, 1);
         }
 
+        if (!$logger) {
+            $logger = ApplicationDefaultCredentials::getDefaultLogger();
+        }
```
bshaffer, line 60: nit (suggestion) - make this one line
```suggestion
        $logger = $logger ?? ApplicationDefaultCredentials::getDefaultLogger();
```

PR #571, opened 2024-08-13
feat: Add a file system cache class

### LINKED ISSUE #498: Add default file caching

Hectorhammett:
Add a PSR-6 compliant `FileSystemCacheItemPool` class for easy configuration fixing some issues with caching in memory as PHP handles requests in different processes. This show alleviate this issue.

closes #498


src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,224 @@
...
+use Exception;
+use Psr\Cache\CacheItemInterface;
+use Psr\Cache\CacheItemPoolInterface;
+
+class FileSystemCacheItemPool implements CacheItemPoolInterface
+{
+    /**
+     * @var string
+     */
+    private string $cachePath = 'cache/';
```
bshaffer, line 30: We should remove the default value, since it gets overwritten by what the user passes into the constructor
```suggestion
    private string $cachePath;
```

tests/Cache/FileSystemCacheItemPoolTest.php
```diff
@@ -0,0 +1,159 @@
...
+namespace Google\Auth\Tests\Cache;
+
+use Google\Auth\Cache\FileSystemCacheItemPool;
+use Google\Auth\Cache\TypedItem;
+use PHPUnit\Framework\TestCase;
+use Psr\Cache\CacheItemInterface;
+
+class FileSystemCacheItemPoolTest extends TestCase
+{
+    private string $defaultCacheDirectory = 'cache/';
```
bshaffer, line 27: nit: make this `.cache`.
```suggestion
    private string $defaultCacheDirectory = '.cache/';
```

What's the best practices with the trailing `/`? I assume it doesn't make a difference either way...

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,224 @@
...
+        $itemPath = $this->cacheFilePath($item->getKey());
+        $serializedItem = serialize($item);
+
+        $result = file_put_contents($itemPath, $serializedItem);
+
+        if (!$result) {
+            return false;
+        }
+
+        return true;
```
bshaffer, line 117: could be
```suggestion
        return false !== file_put_contents($itemPath, $serializedItem);
```

Or if you'd like to be more explicit, that's okay too, but since a [successful write](https://www.php.net/manual/en/function.file-put-contents.php) can contain `0` bytes (e.g. an empty string), we need to use a string boolean comparison here

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,224 @@
...
+    public function clear(): bool
+    {
+        if (!is_dir($this->cachePath)) {
+            return false;
+        }
+
+        $files = scandir($this->cachePath);
+        if (!$files) {
+            return false;
+        }
```
bshaffer, line 140: This is interesting because the [PSR-6 interface says](https://www.php-fig.org/psr/psr-6/):

```php
    /**
     * Deletes all items in the pool.
     *
     * @return bool
     *   True if the pool was successfully cleared. False if there was an error.
     */
    public function clear();
```

But in these first two cases where we `return false`, they could potentially happen in cases where there was no error (e.g. when the cache hasn't been created yet). So I think these should actually return `true`, because if there is no cache, you could say it's "successfully cleared"

We should investigate what other implementations do first, however. I could be wrong.

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,224 @@
...
+    public function deleteItem(string $key): bool
+    {
+        if (!$this->validKey($key)) {
+            throw new Exception("The key '$key' is not valid. The key should follow the pattern |^[a-zA-Z0-9_\.! ]+$|");
+        }
+
+        $itemPath = $this->cacheFilePath($key);
+
+        if (!file_exists($itemPath)) {
+            return false;
```
bshaffer, line 167: Same here as above, the spec is a bit ambiguous

```php
     * @return bool
     *   True if the item was successfully removed. False if there was an error.
     */
    public function deleteItem($key);

```

What happens if the item wasn't successfully removed (because it doesn't exist), but also there wasn't an error? We should verify what other implementations do!

bshaffer (commented):
First pass - looks really good!

src/Cache/FileSystemCacheItemPool.php
Hectorhammett: I thought about that case and in my opinion if you have 0 bytes, then what's the use of that cache object 🤔. But I see an argument for returning the actual value.

I personally don't like how the shorthand reads 😭 haha.

src/Cache/FileSystemCacheItemPool.php
Hectorhammett: It seems that most implementations tends to use an underlying system for handling most of the things and just return that 🤔 Here is one example:
https://github.com/php-cache/filesystem-adapter/blob/master/FilesystemCachePool.php#L94

I personally do agree with the idea of just using it for errors though, will change the implementation.

tests/Cache/FileSystemCacheItemPoolTest.php
```diff
@@ -0,0 +1,229 @@
...
+            $this->getNewItem('NewItem3')
+        ];
+
+        foreach ($items as $item) {
+            $this->pool->save($item);
+        }
+
+        $itemKeys = array_map(function (CacheItemInterface $item) {
+            return $item->getKey();
+        }, $items);
```
bshaffer, line 105:
```suggestion
        $itemKeys = array_map(fn($item) => $item->getKey(), items);
```

tests/Cache/FileSystemCacheItemPoolTest.php
```diff
@@ -0,0 +1,229 @@
...
+
+    public function testGetItemsWithIncorrectKeyShouldThrowAnException()
+    {
+        foreach ($this->invalidChars as $char) {
+            try {
+                $item = $this->getNewItem($char);
+                $this->pool->getItems([$item->getKey()]);
+                $this->fail('The key ' . $char . ' is passing validation when should not be valid');
+            } catch (InvalidArgumentException $e) {
+                $this->assertTrue($e instanceof InvalidArgumentException);
```
bshaffer, line 178: For all of these, use PHPUnit's built-in `expectException` and `expectExceptionMessage` functions, which you should call at the top of the test, e.g.

```php
    public function testGetItemsWithIncorrectKeyShouldThrowAnException()
    {
        $this->expectException(InvalidArgumentException:class);
        $this->expectExceptionMessage('Whatever your expected exception message is');
       // .. continue with the rest of the test
    }

```

tests/Cache/FileSystemCacheItemPoolTest.php
```diff
@@ -0,0 +1,229 @@
...
+            }
+        }
+    }
+
+    public function testDeleteItemWithIncorrectKeyShouldThrowAnException()
+    {
+        foreach ($this->invalidChars as $char) {
```
bshaffer, line 198: instead of using a `foreach` with `invalidChars`, use a `@dataProvider` annotation. For example:

```php
    /**
     * @dataProvider provideInvalidChars
     */
    public function testDeleteItemWithIncorrectKeyShouldThrowAnException($char)
    {
        $this->expectException(InvalidArgumentException::class);
        $this->expectExceptionMessage('Character ' . $char . ' is invalid'); // or whatever the exception message really is
        $item = $this->getNewItem($char);
        $this->pool->hasItem($item->getKey());
    }

    public function provideInvalidChars()
    {
        // each element in the array is an array of arguments, so wrap each char in an array
        return array_map(fn($char) => [$char], $this->invalidChars);
    }
```

src/Cache/FileSystemCacheItemPool.php
```diff
@@ -0,0 +1,226 @@
...
+     * {@inheritdoc}
+     */
+    public function save(CacheItemInterface $item): bool
+    {
+        if (!$this->validKey($item->getKey())) {
+            return false;
+        }
+
+        $itemPath = $this->cacheFilePath($item->getKey());
+        $serializedItem = serialize($item);
```
bshaffer, line 108: is this how it's done in other file caching libraries? Just curious

tests/Cache/FileSystemCacheItemPoolTest.php
Hectorhammett: I actually used that one but because I made this test with this in mind:
If someone edits the list of characters or the `validKey` method gets updated and one of them "passes", using `expectException` passes even when  one character is valid.

For example, I tried proving that this test would fail by adding the char `a` to the invalid characters and the test was passing when in reality should have failed.

I think this might be fixed with the comment below 👇

bshaffer: Yes, you will need to split every exception out into a separate test 👍

src/Cache/FileSystemCacheItemPool.php
Hectorhammett: It depends on what the cache is using, in this case as we're using files it is necessary to serialize the object we want to cache.

In both the `fileSystem` library and symfony's `filesystemAdapter` `serialize` is used:
* Here: https://github.com/php-cache/filesystem-adapter/blob/f1faaae40aaa696ef899cef6f6888aedb90b419b/FilesystemCachePool.php#L113
* And here: https://github.com/symfony/symfony/blob/99b25a1a9de655961ac7c8fbf18e5d00f84ed798/src/Symfony/Component/Cache/Marshaller/DefaultMarshaller.php#L44

bshaffer: I can see why it's important to serialize the contents of the cache item to the cache, but serializing the cache item itself is what struck me as odd. For instance, I do not think we are supposed to preserve `expiresAt` each time, but in this case it'll be preserved as part of the `Item`. 

I believe this is the case with other caching libraries - they serialize the contents of the Item and then create a new one with those contents on hit.

bshaffer: See https://github.com/firebase/php-jwt/issues/543#issuecomment-1842592433

The spec seems to say items which do not have `expiry` set explicitly should be interpreted as never expiring (evne once retrieved from the cache). I don't know if I agree with this, but I think we should be consistent with the spec!

Hectorhammett: https://github.com/symfony/symfony/issues/26241#issuecomment-366975853

PSR-6 is beautiful, innit?

Hectorhammett: Updated!

bshaffer (approved):

PR #578, opened 2024-09-04
feat: add StdOutLogger and LoggingTrait

Hectorhammett:
related: https://github.com/googleapis/gax-php/pull/585

Add logging support to the HTTP calls for easy debugging.


src/Logging/StdOutLogger.php
```diff
@@ -0,0 +1,112 @@
...
+    {
+        $this->log(LogLevel::INFO, $message);
+    }
+
+    public function debug(string|Stringable $message, array $context = []): void
+    {
+        $this->log(LogLevel::DEBUG, $message);
+    }
+
+    public function log($level, string|Stringable $message, array $context = []): void
```
bshaffer, line 95: Adding something like this will get rid of the error?
```
    /**
     * Logs with an arbitrary level.
     *
     * @param mixed $level
     * @param string $message
     * @param array $context
     * @return void
     */
```

src/HttpHandler/Guzzle6HttpHandler.php
```diff
@@ -57,6 +100,41 @@ public function __invoke(RequestInterface $request, array $options = [])
...
+            $requestEvent = new LogEvent();
+
+            $requestEvent->method = $request->getMethod();
+            $requestEvent->url = $request->getUri()->__toString();
+            $requestEvent->headers = $request->getHeaders();
+            $requestEvent->payload = $request->getBody()->getContents();
+            $requestEvent->retryAttempt = $options['retryAttempt'] ?? null;
+            $requestEvent->serviceName = $options['serviceName'] ?? null;
+            $requestEvent->clientId = spl_object_id($this->client);
+            $requestEvent->requestId = spl_object_id($request);
```
bshaffer, line 115: since this code is duplicated above, why not have private methods like `logHttpRequest($request)` and `logHttpResponse($response, $requestEvent)`?

src/HttpHandler/HttpHandlerFactory.php
```diff
@@ -45,6 +51,12 @@ public static function build(ClientInterface $client = null)
             $client = new Client(['handler' => $stack]);
         }
 
+        if ($logger === false) {
+            $logger = null;
+        } else {
+            $logger = $logger ?? ApplicationDefaultCredentials::getDefaultLogger();
+        }
```
bshaffer, line 58: This is 100% not important, if you don't like it please disregard. I only suggest it because it sets the variable in one statement, so it _might_ be considered a little easier to follow (but that's entirely subjective)
```suggestion
        $logger = ($logger === false) 
            ? null
            : $logger ?? ApplicationDefaultCredentials::getDefaultLogger();
```

src/Logging/LogEvent.php
```diff
@@ -0,0 +1,121 @@
...
+     * The Client Id for easy trace
+     *
+     * @var null|int $clientId
+     */
+    public null|int $clientId = null;
+
+    /**
+     * The Request id for easy trace
+     *
+     * @var null|int $requestId;
```
bshaffer, line 102: the format `null|[type]` came from symfony, right? I seem to remember we found a precedent for it and liked it better.

src/Logging/LoggingTrait.php
```diff
@@ -0,0 +1,166 @@
...
+    private function logRequest(LogEvent $event): void
+    {
+        $debugEvent = [
+            'timestamp' => $event->timestamp,
+            'severity' => strtoupper(LogLevel::DEBUG),
+            'clientId' => $event->clientId,
+            'requestId' => $event->requestId ?? null,
+        ];
+
+        $debugEvent = array_filter($debugEvent, fn ($value) => !is_null($value));
```
bshaffer, line 36: i was going to suggest using the default behavior of `array_filter`, so you don't need to provide the second argument, but I see that [it behaves as a call to `empty`](https://www.php.net/manual/en/function.array-filter.php), which is not quite the same as you have here. Is that intentional, that you'd like to capture values of `0`? if so then I think what you have here is great. If not, we can just use `array_filter` without any arguments.

(same question applies to the calls below to `array_filter`)

src/Logging/LoggingTrait.php
```diff
@@ -0,0 +1,166 @@
...
+            ]
+        ];
+
+        $infoEvent = array_filter($infoEvent, fn ($value) => !is_null($value));
+        $infoEvent['jsonPayload'] = array_filter(
+            $infoEvent['jsonPayload'],
+            fn ($value) => !is_null($value)
+        );
+
+        $this->logger->info((string) json_encode($infoEvent));
```
bshaffer, line 139: is there a reason we don't wrap this one in the `$stringifiedEvent === false` conditional as we do the others?

src/Logging/LoggingTrait.php
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    /**
+     * @param array<mixed> $headers
+     * @return null|array<mixed, mixed>
```
bshaffer, line 144: we can make this a little smarter
```suggestion
     * @return null|array<string, string|false>
```

src/Logging/LoggingTrait.php
```diff
@@ -0,0 +1,166 @@
...
+        $debugEvent = array_filter($debugEvent, fn ($value) => !is_null($value));
+        $debugEvent['jsonPayload'] = array_filter(
+            $debugEvent['jsonPayload'],
+            fn ($value) => !is_null($value)
+        );
+
+        $stringifiedEvent = json_encode($debugEvent);
+
+        // There was an error stringifying the event, return to not break execution
+        if ($stringifiedEvent === false) {
```
bshaffer, line 87: do we still want to try to log the infoEvent even if the debug event fails? for instance do this instead?

```php
if ($stringifiedEvent !== false) {
    $this->logger->debug($stringifiedEvent);
}
```

src/Logging/StdOutLogger.php
```diff
@@ -0,0 +1,144 @@
...
+
+        print($message . "\n");
+    }
+
+    /**
+     * @param string $levelName
+     * @return int
+     * @throws InvalidArgumentException
+     */
+    private function getLevelMap(string $levelName): int
```
bshaffer, line 136: nit: to me `getLevelMap` would return the `levelMapping` variable. What if we instead called it `getLevelNumber` or `getLevelFromName`?

tests/Logging/StdOutLoggerTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+        $logger = new StdOutLogger(LogLevel::DEBUG);
+        $expectedString = 'test';
+        $logger->debug($expectedString);
+        $buffer = ob_get_contents();
+
+        $this->assertEquals($expectedString . "\n", $buffer);
+
+        ob_end_clean();
+    }
```
bshaffer, line 45: Instead of usign `ob_start` and `ob_end_clean`, use `expectOutputString`. See https://docs.phpunit.de/en/10.5/writing-tests-for-phpunit.html#testing-output for more info

src/ApplicationDefaultCredentials.php
```diff
@@ -321,6 +328,40 @@ public static function getIdTokenCredentials(
...
+        if ($loggingFlag !== 'true' && $loggingFlag !== '1') {
+            // Env var is set to a non valid value
+            if ($loggingFlag !== 'false' && $loggingFlag !== '0') {
+                trigger_error('The ' . self::SDK_DEBUG_FLAG . ' is set, but it is set to another value than false, true, 0 or 1. Logging is disabled');
+            }
+
+            return null;
+        }
+
+        return new StdOutLogger();
```
bshaffer, line 362: Having lots of `!==` is a bit hard to read. What about something like this?

**NOTE**: You may want to handle the case of empty string (e.g. someone sets `GOOGLE_SDK_DEBUG_LOGGING=` in their environment) to have the same behavior, which is why I included the empty string in the array.

```suggestion
        // Env Var is true
        if ($loggingFlag === 'true' || $loggingFlag === '1') {
            return StdOutLogger();
        }

        // Env Var is false or not specified
        if (in_array($loggingFlag, [null, '', 'false', '0'], true)) {
            return null;
        }
        
        // Env var is set to a non valid value
        trigger_error('The ' . self::SDK_DEBUG_FLAG . ' is set, but it is set to another value than false, true, 0 or 1. Logging is disabled');

        // still return null because we don't want debugging to halt execution
        return null;
```

I personally think this reads a lot cleaner

src/HttpHandler/Guzzle6HttpHandler.php
```diff
@@ -44,7 +56,38 @@ public function __construct(ClientInterface $client)
...
+            $this->logRequest($requestEvent);
+        }
+
+        $response = $this->client->send($request, $options);
+
+        if ($this->logger) {
+            $responseEvent = new LogEvent($requestEvent->timestamp);
+
+            $responseEvent->headers = $response->getHeaders();
+            $responseEvent->payload = $response->getBody()->getContents();
```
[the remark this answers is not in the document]
Hectorhammett: Yeah, there was a discussion about truncating the data for this and there is no definite conclusion, but one idea is to truncate the length and be configurable via the `StdOutLogger` constructor.

src/HttpHandler/Guzzle6HttpHandler.php
Hectorhammett: In my mind, formatting the `LogEvent` and filling that data is not duty of the `log` function. The `log` function takes the generic log object and turns it into a `json` string to the docs specification, that's why I decided to have this code "duplicated", which in my mind is not really "duplicated" but is basically the same, mostly because one is for the response and one is for the request, and I did it this way in order to keep them separate from the logging logic.

I also thought perhaps a function to make a `logEvent` but voted against it as it would have to be flexible enough to acomodate the small differences between `response` and `request` but it can definitely be done.

src/HttpHandler/HttpHandlerFactory.php
Hectorhammett: I personally don't love the `??` inside a ternary, but I can see the appeal.

src/Logging/LogEvent.php
Hectorhammett: We found that Symfony had a mixture of `null|type` and the `?` but we theorize that it was because PHP didn't used to support union types before haha.

src/Logging/LoggingTrait.php
Hectorhammett: Yeah, this is intentional. The first version was using the default behaviour of `filter` but whenever there was a value of `0` or `false` it was getting removed, so I changed it.

src/Logging/LoggingTrait.php
Hectorhammett: Yes, me. I am the reason, I am just dumb 🫠 haha.

Let me fix it.

src/ApplicationDefaultCredentials.php
```diff
@@ -321,6 +328,40 @@ public static function getIdTokenCredentials(
...
+
+        // Env var is not set
+        if (!is_string($loggingFlag)) {
+            if (is_array($loggingFlag)) {
+                trigger_error('The ' . self::SDK_DEBUG_FLAG . ' is set, but it is set to another value than false, true, 0 or 1. Logging is disabled');
+                return null;
+            }
+
+            return null;
+        }
```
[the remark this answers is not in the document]
Hectorhammett: I remember having a specific issue that I needed to handle but I can't quite remember what was it exactly. Although it might have been related to PHPSTAN mentioning that it can return an `array`? But I am unsure. 

Still let me take a look in order to make it cleaner :)

src/HttpHandler/Guzzle6HttpHandler.php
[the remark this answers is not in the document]
Hectorhammett: This is now truncated here:
https://github.com/googleapis/google-auth-library-php/pull/578/files#diff-1d54313afff4c789962b3bd7538deadfcf14ccd1787d9b47cbdd362e2db23702R156

src/ApplicationDefaultCredentials.php
```diff
@@ -318,6 +324,34 @@ public static function getIdTokenCredentials(
         return $creds;
     }
 
+    /**
+     * Returns a StdOutLogger instance
+     *
+     * @return null|LoggerInterface
+     */
+    public static function getDefaultLogger(): null|LoggerInterface
```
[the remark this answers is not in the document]
Hectorhammett: Marked it as internal now!

tests/ApplicationDefaultCredentialsTest.php
```diff
@@ -772,6 +775,35 @@ public function testExternalAccountCredentials(string $jsonFile, string $expecte
...
+    {
+        putenv($this::SDK_DEBUG_FLAG . '=false');
+        $logger = ApplicationDefaultCredentials::getDefaultLogger();
+
+        $this->assertNull($logger);
+
+        putenv($this::SDK_DEBUG_FLAG . '=0');
+        $logger = ApplicationDefaultCredentials::getDefaultLogger();
+
+        $this->assertNull($logger);
```
bshaffer, line 795: Add a test for empty string?
```php
        putenv($this::SDK_DEBUG_FLAG . '=');
        $logger = ApplicationDefaultCredentials::getDefaultLogger();

        $this->assertNull($logger);
```

tests/HttpHandler/Guzzle7HttpHandlerTest.php
```diff
@@ -31,4 +38,47 @@ public function setUp(): void
...
+            ->willReturn($requestPromise)
+            ->shouldBeCalledTimes(1);
+
+        $request = new Request('GET', 'https://domain.tld');
+        $options = ['key' => 'value'];
+
+        /**
+         * @var LoggerInterface $mockLogger
+         * @var ClientInterface $mockClient
+         */
```
bshaffer, line 80:
```suggestion
```

tests/HttpHandler/Guzzle7HttpHandlerTest.php
```diff
@@ -31,4 +38,47 @@ public function setUp(): void
...
+        $handler = new Guzzle7HttpHandler($this->client->reveal());
+        $handler->async($request, $options)->wait();
```
bshaffer, line 82: Add assertion to make sure output is empty?
```php
$this->expectOutputString('');
```

tests/Logging/LoggingTraitTest.php
```diff
@@ -0,0 +1,118 @@
...
+use Google\Auth\Tests\BaseTest;
+use Psr\Log\LoggerInterface;
+
+class LoggingTraitTest extends BaseTest
+{
+    private MockClassWithLogger $loggerContainer;
+
+    public function setUp(): void
+    {
+        $this->loggerContainer = new MockClassWithLogger();
```
bshaffer, line 32: nit: consider using anonymous classes because they're kinda cool

```suggestion
        $this->loggerContainer = new class() {
            use LoggingTrait;
            private LoggerInterface $logger;

            public function __construct()
            {
                $this->logger = new StdOutLogger();
            }

            public function logRequestEvent(RpcLogEvent $event): void
            {
                $this->logRequest($event);
            }

            public function logResponseEvent(RpcLogEvent $event): void
            {
                $this->logResponse($event);
            }
        };
```

src/HttpHandler/Guzzle6HttpHandler.php
```diff
@@ -59,32 +59,13 @@ public function __invoke(RequestInterface $request, array $options = [])
...
-            $responseEvent = new RpcLogEvent($requestEvent->milliseconds);
-
-            $responseEvent->headers = $response->getHeaders();
-            $responseEvent->payload = $response->getBody()->getContents();
-            $responseEvent->status = $response->getStatusCode();
-            $responseEvent->processId = $requestEvent->processId;
-            $responseEvent->requestId = $requestEvent->requestId;
-
-            $this->logResponse($responseEvent);
+           $this->responseLog($response, $requestEvent);
```
bshaffer, line 68:
```suggestion
            $this->responseLog($response, $requestEvent);
```

src/HttpHandler/Guzzle6HttpHandler.php
```diff
@@ -103,38 +84,52 @@ public function async(RequestInterface $request, array $options = [])
...
+                $this->responseLog($response, $requestEvent);
                 return $response;
             });
         }
 
         return $promise;
     }
+
+    /**
+     * @internal
```
bshaffer, line 103:
```suggestion
     * @internal
     * @param RequestInterface $request
     * @param array<mixed> $options
```

tests/Logging/LoggingTraitTest.php
```diff
@@ -0,0 +1,130 @@
...
+    public function logRequestEvent(RpcLogEvent $event): void
+    {
+        $this->logRequest($event);
+    }
+
+    public function logResponseEvent(RpcLogEvent $event): void
+    {
+        $this->logResponse($event);
+    }
+}
```
bshaffer, line 130:
```suggestion
```

tests/HttpHandler/Guzzle7HttpHandlerTest.php
```diff
@@ -18,6 +18,13 @@
 namespace Google\Auth\Tests\HttpHandler;
 
 use Google\Auth\HttpHandler\Guzzle7HttpHandler;
+use Google\Auth\Logging\StdOutLogger;
+use GuzzleHttp\ClientInterface;
```
bshaffer, line 22:
```suggestion
```

tests/HttpHandler/Guzzle7HttpHandlerTest.php
```diff
@@ -18,6 +18,13 @@
...
+use GuzzleHttp\Promise\Promise;
+use GuzzleHttp\Psr7\Request;
+use GuzzleHttp\Psr7\Response;
+use Prophecy\Argument;
+use Psr\Log\LoggerInterface;
```
bshaffer, line 27:
```suggestion
```

bshaffer (approved):

PR #580, opened 2024-10-01
feat: add support for Impersonating ID Tokens

bshaffer:
Although we've had the class `ImpersonatedServiceAccountCredentials` for a while, it does not actually make the request to exchange credentials for impersonated ones. So I'm not convinced this actually ever worked.

This PR adds support for the following for  `ImpersonatedServiceAccountCredentials`:

|| ID Tokens |ID Tokens with Universe Domain (https://github.com/googleapis/google-auth-library-php/pull/581)|
|---| ------------- | ------------- |
|UserRefreshCredentials       | ✅  | n/a |
|ServiceAccountCredentials | ✅  | ✅ |
|ExternalAccountCredentials| ✅ |  ✅ |
|Arbitrary Credentials (e.g. `GCECredentials`)| ✅ |✅|

The same support, but for access tokens, can be found in https://github.com/googleapis/google-auth-library-php/pull/586

**NOTE**: Tests have been added for all rows for the first two columns above. Universe domain test cases have not been added because 1. they've been tested manually E2E and 2. There is no change of behavior in `ImpersonatedServiceAccountCredentials` for universe domain, as the logic is handled entirely by the underlying source credentials.


src/Credentials/ImpersonatedServiceAccountCredentials.php
```diff
@@ -33,50 +41,90 @@ class ImpersonatedServiceAccountCredentials extends CredentialsLoader implements
...
-            $scope,
-            $jsonKey['source_credentials']
-        );
+        if (
+            $targetAudience !== null
+            && $jsonKey['source_credentials']['type'] === 'service_account'
+        ) {
+            // Service account tokens MUST request a scope, and as this token is only used to impersonate
+            // an ID token, the narrowest scope we can request is `cloud-platform`.
+            $scope = 'https://www.googleapis.com/auth/cloud-platform';
```
bshaffer, line 122: I would like confirmation from @aeitzman that this is okay to do

bshaffer:
cc @gjvanahee I'd love for you to test this implementation and see if it works for your use-case!

gjvanahee:
I will test it today! Let you know

src/Credentials/ImpersonatedServiceAccountCredentials.php
```diff
@@ -123,11 +171,52 @@ public function getClientName(callable $unusedHttpHandler = null)
...
+            false => [
+                'scope' => $this->targetScope,
+                'delegates' => $this->delegates,
+                'lifetime' => sprintf('%ss', $this->lifetime),
+            ]
+        };
+
+        $request = new Request(
+            'POST',
+            $this->serviceAccountImpersonationUrl,
```
gjvanahee, line 205: I get an error (http 400) here about the json payload in the outgoing request "Invalid JSON payload received. Unknown name \"audience\": Cannot find ...". When I update the url by replacing `generateAccessToken` to `generateIdToken` everything works as expected!

This is what I changed: 
```php
        if ($this->isIdTokenRequest()) {
            $url = str_replace('generateAccessToken', 'generateIdToken', $this->serviceAccountImpersonationUrl);
            $body = [
                'audience' => $this->targetAudience,
                'includeEmail' => true,
            ];
        } else {
            $body = [
                'scope' => $this->targetScope,
                'delegates' => $this->delegates,
                'lifetime' => sprintf('%ss', $this->lifetime),
            ];
            $url = $this->serviceAccountImpersonationUrl;
        }

        $request = new Request(
            'POST',
            $url,
            $headers,
            (string) json_encode($body)
        );
```

src/Credentials/ImpersonatedServiceAccountCredentials.php
bshaffer: @aeitzman replied to me offline, and pointed to the python auth library, which is doing the same thing - https://github.com/googleapis/google-auth-library-python/blob/484c8db151690a4ae7b6b0ae38db0a8ede88df69/google/auth/impersonated_credentials.py#L207C1-L214C14

I've updated the scope to use the `auth/iam` scope instead, which definitely seems more appropriate.

src/Credentials/ImpersonatedServiceAccountCredentials.php
bshaffer: @gjvanahee that value is defined in your credentials in the `service_account_impersonation_url` JSON value. I do not want to do a string find-and-replace on values in the credentials, so I'm not sure what the appropriate approach is here.

Python does seem to do some sort of templating here, so maybe this requires further consideration: 

https://github.com/googleapis/google-auth-library-python/blob/484c8db151690a4ae7b6b0ae38db0a8ede88df69/google/auth/iam.py#L41-L54

gjvanahee: I agree with the str_replace. I just wanted to make the change more explicit. I don't like to have to change a generated key file. In my PR I used `"https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{$impersonatedServiceAccount}:generateIdToken",` but it looks a lot nicer to have all the options together in constants like in the python example

gjvanahee: Could `Google\Auth\Iam` class be changed to add methods for requesting an id token (and perhaps access_token)? All the actions on that endpoint would then be together in that class. Something like
```
    private const ID_TOKEN_PATH = '%s:generateIdToken';
    public function generateIdToken(string $email, string $targetAudience, string $accessToken): string
    {
        $httpHandler = $this->httpHandler;
        $name = sprintf(self::SERVICE_ACCOUNT_NAME, $email);
        $apiRoot = str_replace('UNIVERSE_DOMAIN', $this->universeDomain, self::IAM_API_ROOT_TEMPLATE);
        $uri = $apiRoot . '/' . sprintf(self::ID_TOKEN_PATH, $name);

        $body = [
            'audience' => $targetAudience,
            'includeEmail' => true,
        ];

        $headers = [
            'Authorization' => 'Bearer ' . $accessToken,
            'Content-Type' => 'application/json',
            'Cache-Control' => 'no-store',
        ];

        $request = new Psr7\Request(
            'POST',
            $uri,
            $headers,
            Utils::streamFor(json_encode($body))
        );

        $res = $httpHandler($request);
        $body = json_decode((string) $res->getBody(), true);

        return $body['token'];
    }
```
and some refactoring to avoid duplication of course.
It would skip the call to `applyTokenEndpointMetrics` though or that has to be used in the `Iam` class too...

bshaffer: You seem to be describing the work I did in https://github.com/googleapis/google-auth-library-php/pull/581 :)

gjvanahee: Hehehe, it seems I'm always one step behind... I also found it strange that the service account being impersonated is not explicitly mentioned in the key file. The endpoints for the actions are [documented](https://cloud.google.com/iam/docs/reference/credentials/rest) and [discoverable](https://iamcredentials.googleapis.com/$discovery/rest?version=v1), but that is your call ;)

src/Credentials/ImpersonatedServiceAccountCredentials.php
```diff
@@ -71,10 +77,12 @@ class ImpersonatedServiceAccountCredentials extends CredentialsLoader implements
      *     @type int                            $lifetime The lifetime of the impersonated credentials
      *     @type string[]                       $delegates The delegates to impersonate
      * }
+     * @param string|null $targetAudience The audience to request an ID token.
      */
     public function __construct(
         $scope,
-        $jsonKey
+        $jsonKey,
```
Hectorhammett, line 84: I see there are type annotations on the doc comment, should we type these? Just a thought though.

Hectorhammett (approved):

src/Credentials/ImpersonatedServiceAccountCredentials.php
bshaffer: Good catch! added in b49eeb9

majna:
Hey @bshaffer,

This PR breaks google/cloud-bigquery with impersonated service account, and possibly others.

Relates to https://github.com/googleapis/google-auth-library-php/issues/614

bshaffer:
@majna thanks for the heads up, I'll take a look

PR #581, opened 2024-10-01
feat: use IamCredentials endpoint for generating ID tokens outside GDU

bshaffer:
Uses the `iamcredentials` service in order to generate an ID token when `universe_domain` is supplied and is not the default (`googleapis.com`). 

 - [x] verify generating the ID token works
 - [x] verify generated ID token is valid
 - [x] write unit tests
 - [x] consider merging https://github.com/googleapis/google-auth-library-php/pull/468 as part of this

### DIFF
--- a/src/Iam.php
+++ b/src/Iam.php
@@ -46,10 +46,10 @@ class Iam
     private string $universeDomain;
 
     /**
-     * @param callable $httpHandler [optional] The HTTP Handler to send requests.
+     * @param callable|null $httpHandler [optional] The HTTP Handler to send requests.
      */
     public function __construct(
-        callable $httpHandler = null,
+        ?callable $httpHandler = null,
         string $universeDomain = GetUniverseDomainInterface::DEFAULT_UNIVERSE_DOMAIN
     ) {
         $this->httpHandler = $httpHandler
--- a/src/Iam.php
+++ b/src/Iam.php
@@ -118,19 +118,21 @@ class Iam
      * @param string $clientEmail The service account email.
      * @param string $targetAudience The audience for the ID token.
      * @param string $bearerToken The token to authenticate the IAM request.
+     * @param array<string, string> $headers [optional] Additional headers to send with the request.
      *
      * @return string The signed string, base64-encoded.
      */
     public function generateIdToken(
         string $clientEmail,
         string $targetAudience,
-        string $bearerToken
+        string $bearerToken,
+        array $headers = []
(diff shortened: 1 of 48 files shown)


src/Iam.php
```diff
@@ -102,9 +102,52 @@ public function signBlob($email, $accessToken, $stringToSign, array $delegates =
...
+     * @param string $clientEmail The service account email.
+     * @param string $targetAudience The audience for the ID token.
+     * @param string $bearerToken The token to authenticate the IAM request.
+     *
+     * @return string The signed string, base64-encoded.
+     */
+    public function generateIdToken(
+        string $clientEmail,
+        string $targetAudience,
+        string $bearerToken
```
gjvanahee, line 127: perhaps there could be a fourth parameter `array $headers = []`. This way the calling class (e.g. `ImpersonatedServiceAccountCredentials`) could add a token endpoint metrics header. (The bearer token could stay as-is as it is required for the id token to be returned.)

bshaffer: Thanks for this feedback! Could you provide me an example of what that would look like? I assume it would look like adding the following in the call to [Iam::generateIdToken](https://github.com/googleapis/google-auth-library-php/pull/581/files#diff-248500c0207a8c4941997a322cc9895767971446b14eb6722257d5c694c3c7ffR239):

```php
            $idToken = (new Iam($httpHandler, $this->getUniverseDomain()))->generateIdToken(
                $this->auth->getIssuer(),
                $this->auth->getAdditionalClaims()['target_audience'],
                $jwt,
                $this->applyTokenEndpointMetrics([], 'it') // this is the new line
            );
```

gjvanahee: Yes, that is exactly what I had in mind. All requests I encountered getting an acces or id token send metrics headers, so it makes sense to add them here too. The implementation would be something like this.
```
    /**
     * @param array $headers Optional headers to send to the api, defaults to an empty array
     */
    public function generateIdToken(
        string $clientEmail,
        string $targetAudience,
        string $bearerToken,
        array $headers = [] // -> added optional parameter
): string {
        // ....
       // updated header assignment
       $headers['Authorization'] = 'Bearer ' . $bearerToken; 
        // ....
```

Hectorhammett (approved):

src/Iam.php
bshaffer: This has been added in  475bc74

PR #586, opened 2024-11-05
feat: add service account impersonation for access tokens

bshaffer:
Same as https://github.com/googleapis/google-auth-library-php/pull/580, but with all ID token logic removed (for now)


bshaffer:
This would be great to get merged next (by Nov 15)

src/FetchAuthTokenInterface.php
```diff
@@ -26,6 +26,7 @@ interface FetchAuthTokenInterface
      * Fetches the auth tokens based on the current state.
      *
      * @param callable|null $httpHandler callback which delivers psr7 request
+     *
```
Hectorhammett, line 29: Nit, Is the @param and @return docs separated by a new line?

Hectorhammett (approved):

src/FetchAuthTokenInterface.php
bshaffer: Using regex, I found 8 instances where we had a newline between `@param` and `@return`, and 92 occurrences where we didn't.

bshaffer:
```suggestion
```

Issue #590, opened 2024-11-20
Backporting PHP 8.4 compatibility


mikkamp:
PHP 8.4 compatibility was released in version [1.43.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.43.0) through PR #584

We need to offer compatibility to PHP 7.4+ meaning we are currently locked to the last version which supported PHP 7.4 (version [1.37.1](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.37.1))

So we'd like to request for the PHP 8.4 compatibility to be backported to this same version to be able to support PHP 7.4 up to and including PHP 8.4.

bshaffer:
Fixed in release `v1.37.2`

PR #592, opened 2024-12-04
fix: add PHP 8.4 compatibility to last release compatible with PHP 7.4

bshaffer:
Ensures that the last `google/auth` release which was compatible with PHP 7.4 will be compatible with PHP 8.4 as well.

Once this is merged, we will tag this as `v1.37.2`

fixes https://github.com/googleapis/google-auth-library-php/issues/590

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,7 +11,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php: [ "7.4", "8.0", "8.1", "8.2", "8.3" ]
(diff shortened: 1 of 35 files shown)


mikkamp:
Thanks @bshaffer 
We've tested using these branches with our extension in https://github.com/woocommerce/google-listings-and-ads/pull/2714
Everything looks good and we no longer receive deprecation notices from this package.

bshaffer:
@mikkamp thanks for your review! I've tagged `v1.37.2` with these changes, so this should be good to go.

PR #598, opened 2025-01-16
docs: add warning against accepting untrusted credentials

bshaffer:


src/ApplicationDefaultCredentials.php
```diff
@@ -69,6 +69,12 @@
  *
  * $res = $client->get('myproject/taskqueues/myqueue');
  * ```
+ *
+ * **Important**: If you accept a credential configuration (credential JSON/File/Stream) from an
+ * external source for authentication to Google Cloud Platform, you must validate it before
+ * providing it to any Google API or library. Providing an unvalidated credential configuration to
+ * Google APIs can compromise the security of your systems and data. For more information
+ * {@see https://cloud.google.com/docs/authentication/external/externally-sourced-credentials}
```
sai-sunder-s, line 77: can you put the warning as part of the method that loads any json/file/stream and builds an appropriate cred for it.

See https://github.com/googleapis/google-auth-library-python/pull/1655/files as an example

sai-sunder-s (changes requested):

src/ApplicationDefaultCredentials.php
bshaffer: The only issue there is that we would need to add it to `ApplicationDefaultCredentials::getCredentials`, `ApplicationDefaultCredentials::getMiddleware`, `ApplicationDefaultCredentials::getIdTokenCredentials`, `ApplicationDefaultCredentials::getIdTokenMiddleware`, and `ApplicationDefaultCredentials::getIdTokenProxyCredentials`. I can do it, but it just seems a bit over-verbose. Since it applies to essentially every method in the class, it makes more sense to me to add it at the top of the class.

sai-sunder-s: Sounds good.

sai-sunder-s: Actually .Net has similar case. looks ok to update everywhere https://github.com/googleapis/google-api-dotnet-client/pull/2916

bshaffer: As mentioned offline, because the methods I list above don't accept arbitrary credentials, but rather only can load from the well-known file location or the well-known env var, the warning does not need to be added to them. I still think it makes sense to keep the warning on the ADC class as a whole, but I'm happy to remove it if you think otherwise.

I've added the warning to `CredentialsLoader::makeCredentials`, which is the only method in this auth library that takes an arbitrary set of credentials.

sai-sunder-s: We are not adding warning for ADC in other languages. Anyway most people do not use ADC directly.

I think for consistency, we can remove warning in ADC. But I am ok to keep as well, no harm.

sai-sunder-s (approved):

src/ApplicationDefaultCredentials.php
bshaffer: ok I'll remove it! I'd rather not have it, but it seemed like the right thing to do.

bshaffer, line 77:
```suggestion
```

amanda-tarafa (approved):

PR #604, opened 2025-02-04
fix: fix `getLastReceivedToken` to respect impersonated access token

mruell:


mruell:
Am currently not able to sign the `CLA` without discussing this first internally
But honestly it's an easy fix - so you should be able to fix this "in your own words" :grimacing:

bshaffer:
Thanks for finding this and for proposing the fix! I've created my own fix and added tests here:

https://github.com/googleapis/google-auth-library-php/pull/606

Feel free to leave a review if you'd like. We should get this released soon!

PR #612, opened 2025-03-05
feat: add support for $_ENV and $_SERVER in CredentialsLoader

### LINKED ISSUE #432: feat: add support for using $_ENV instead of getenv/putenv
`getenv` and `putenv` are considered not threadsafe. As a result, 
We should expose a way (similar to symphony's `usePutenv` function) to use the threadsafe version (`$_ENV`) instead. 

As @DracoBlue mentioned in https://github.com/googleapis/google-auth-library-php/pull/429:

> The current version of google auth library php uses getenv/putenv to read and set environment variables in php. But the  function putenv is not threadsafe, which made libraries (Dotenv https://github.com/symfony/dotenv/blob/6.2/Dotenv.php#L61 and phpdotenv https://github.com/vlucas/phpdotenv#putenv-and-getenv) to disable putenv support by default. Frameworks like symfony (deprecated in 4.3 https://github.com/symfony/symfony/pull/31062 and removed in 5.0) and also in laravel there was a similiar solution to not use it anymore directly (https://github.com/vlucas/phpdotenv/issues/76).
>
> Thus here is a pull request to change all the ocurences of:
>
> - `getenv($variable)` to `array_key_exists($variable, $_ENV) ? $_ENV[$variable] : false`
> - `putenv($variable)` to `unset($_ENV[$variable])`
> - `putenv($variable . "=")` to `$_ENV[$variable]=''`
> - `putenv($variable . "=" . $value)` to `$_ENV[$variable]=$value`
>
> Threads about this:
>
> - https://bugs.php.net/bug.php?id=74143 (putenv does not work if fast cgi sets the variable!)
> - https://github.com/symfony/symfony/pull/31062
> - https://github.com/vlucas/phpdotenv/issues/76
> - https://github.com/laravel/framework/issues/7354 
> - https://github.com/laravel/framework/commit/866a8cfc58430830069bd24feacf2d81e99385a2
> - https://github.com/silverstripe/silverstripe-framework/pull/7484
> - https://github.com/laravel/framework/issues/27949

jannes-io:
Many modern frameworks use dotenv or similar methods of loading environment variables into the superglobals, but do not use putenv in doing so. Which makes it harder to use this library, having to use factories to dynamically build client libs and setting the authentication that way instead of using this default method of authenticating.

Partially fixes #432  (only applies to the credentials loader)

Relates to: symfony/symfony#31062

Backwards compatibility is provided by always first checking `getenv`. If you want me to extract this function and replace all occurrences of getenv with it, then that's fine too. Just let me know.


jannes-io:
CLA signed.

bshaffer (approved):

bshaffer:
Hello @jannes-io! Thanks for your contributions!

It looks like I don't have permissions to update your branch, so could you update your branch with the changes from `main` in this repo? That should also trigger the tests, which need to run in order for me to merge this.

Thanks again!

jannes-io:
Hi @bshaffer ,

The "Allow edits by maintainers" is also enabled, so you should be able to add commits. But it seems like there are no updates to main? The branch appears to be up to date.

image

Let me know if I can help.

bshaffer:
@jannes-io You're right, my mistake! Okay, I created an empty commit to trigger the actions.

bshaffer (approved):

bshaffer:
@jannes-io hmm, now I'm thinking, is there any real reason we want to load from `_SERVER`? It seems to me that `_ENV` would be sufficient and preferable.

jannes-io:
> @jannes-io hmm, now I'm thinking, is there any real reason we want to load from `_SERVER`? It seems to me that `_ENV` would be sufficient and preferable.

Honestly? I don't have a reason for it besides symfony/dotenv placing all `.env` variables in both `_SERVER` and `_ENV`.

So my reasoning is that they probably had a good and well thought out reason for doing both, so 3rd party libs that read things from the environment should probably be checking both for existence of env vars? 😆

bshaffer:
@jannes-io okay that's what I thought! Thanks for your quick reply.

So it makes sense to WRITE to both (since SERVER also contains ENV), but I can't think of a reason to READ from both (symfony is writing, and we're reading). I think I'll submit a PR to remove it (I get that it's a little pedantic, but this is an auth library, so it's good to keep thing simple and by the book)

bshaffer:
@jannes-io released in [v1.47.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.47.0). Thanks for your contribution!!

PR #616, opened 2025-04-15
chore: remove server superglobal support

bshaffer:
follow up to https://github.com/googleapis/google-auth-library-php/pull/612

Remove use of `$_SERVER` as it isn't really necessary to support it.

A bit picky of me, but I'd like to keep things as simple as possible.

cc @jannes-io

### DIFF
--- a/src/CredentialsLoader.php
+++ b/src/CredentialsLoader.php
@@ -295,6 +295,6 @@ abstract class CredentialsLoader implements
 
     private static function getEnv(string $env): mixed
     {
-        return getenv($env) ?: $_SERVER[$env] ?? $_ENV[$env] ?? null;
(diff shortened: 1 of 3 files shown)


Hectorhammett (approved):

jannes-io:
Hi @bshaffer,

Totally agree things should be kept as simple and semantic as possible 👍 

After some quick tests, `$_SERVER` is a valid location for env variables to exist when running from PHP cli for example.

Try the following script:
```php
<?php

var_dump($_ENV);
var_dump($_SERVER);
```
`php test.php` results in
```
array(0) {
}
array(52) {
<my system's environment variables here>
}
```

So learning about this, keeping `$_SERVER` seems like the right thing to do.

PR #662, opened 2026-06-04
feat: add ExternalAccountAuthorizedUser credentials

bshaffer:


src/Credentials/ExternalAccountAuthorizedUserCredentials.php
```diff
@@ -0,0 +1,166 @@
...
+     * Return the Cache Key for the credentials.
+     * The format for the Cache key is one of the following:
+     * ClientId.Scope
+     * ClientId.Audience
+     *
+     * @return string
+     */
+    public function getCacheKey()
+    {
+        return $this->clientId . '.' . $this->auth->getScope();
```
lsirac: The clientId is always the same. So the clientId + scope can be the same across creds with different refresh tokens. 

Please update + add a test.

src/Credentials/ExternalAccountAuthorizedUserCredentials.php
```diff
@@ -0,0 +1,166 @@
...
+            );
+        }
+        if (!array_key_exists('token_url', $jsonKey)) {
+            throw new InvalidArgumentException(
+                'json key is missing the token_url field'
+            );
+        }
+        $this->clientId = $jsonKey['client_id'];
+        $this->clientSecret = $jsonKey['client_secret'];
+        $this->auth = new OAuth2([
```
lsirac, line 92: Missing universe domain. We need to make sure universe domain is not dropped and is read and used from the config.

src/Credentials/ExternalAccountAuthorizedUserCredentials.php
bshaffer: good catch. Done!

src/Credentials/ExternalAccountAuthorizedUserCredentials.php
bshaffer: good catch again. done.

lsirac (approved):
Thank you!

src/Credentials/ExternalAccountAuthorizedUserCredentials.php
```diff
@@ -0,0 +1,183 @@
...
+            throw new InvalidArgumentException(
+                'json key is missing the refresh_token field'
+            );
+        }
+        if (!array_key_exists('token_url', $jsonKey)) {
+            throw new InvalidArgumentException(
+                'json key is missing the token_url field'
+            );
+        }
+        $this->clientId = $jsonKey['client_id'];
```
Hectorhammett, line 89: Very basic nit:
Add one space here. I like having code like test, adding a paragraph to denote a different idea. First validation, then the paragraph for initialization.

src/CredentialsLoader.php
```diff
@@ -182,6 +183,11 @@ public static function makeCredentials(
             return new ExternalAccountCredentials($anyScope, $jsonKey);
         }
 
+        if ($jsonKey['type'] == 'external_account_authorized_user') {
```
Hectorhammett, line 186: Can we have this `external_account_authorized_user` in a constant perhaps?

Hectorhammett (approved):

src/CredentialsLoader.php
bshaffer: We can, but then we should put all credential types in constants, for consistency

src/Credentials/ExternalAccountAuthorizedUserCredentials.php
bshaffer: By space do you mean newline?

bshaffer:
```suggestion

        $this->clientId = $jsonKey['client_id'];
```

PR #663, opened 2026-06-04
WIP - add RAB to ExternalAccountAuthorizedUser

bshaffer:
Depends on #649 before we can add it to #662


tests/Credentials/ExternalAccountAuthorizedUserCredentialsTest.php
```diff
@@ -1,5 +1,6 @@
 <?php
 
+
```
bshaffer:
```suggestion
```

bshaffer:
added in https://github.com/googleapis/google-auth-library-php/pull/649/changes/d5a9824f15529131b2c9d108aaf233e48132c503

PR #664, opened 2026-06-04
feat: add X509 credential source

bshaffer:


Hectorhammett:
We migrated this library into our monorepo. Pr re opened here:
https://github.com/googleapis/google-cloud-php/pull/9681

PR #677, opened 2026-07-21
feat: add support for Guzzle 8

GrahamCampbell:
Guzzle 8.0.0 was released on July 20th. Lots of people want to use it, and thus popular packages in the ecosystem like this one need to be updated to support it, so here I am, adding support. I've also removed some dead code that is no longer reachable after a previous removal of support for Guzzle 6, although I left on unused class in to avoid a technical BC break - let me know if you want it removed too.


src/Credentials/RegionalAccessBoundaryTrait.php
```diff
@@ -8,6 +8,7 @@
 use Google\Auth\HttpHandler\HttpHandlerFactory;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Psr7\Request;
+use Psr\Http\Client\NetworkExceptionInterface;
 use InvalidArgumentException;
```
bshaffer: please swap these to appease our picky cs fixer
```suggestion
use InvalidArgumentException;
use Psr\Http\Client\NetworkExceptionInterface;
```

bshaffer (approved):
This looks great! Thank you for your work on this!

tests/BaseTest.php
```diff
@@ -2,36 +2,14 @@
 
 namespace Google\Auth\Tests;
 
-use GuzzleHttp\ClientInterface;
 use PHPUnit\Framework\TestCase;
 
 abstract class BaseTest extends TestCase
 {
+    // TODO: Drop once Guzzle6HttpHandlerTest is no longer a shared test base.
```
bshaffer: Feel free to remove this method and remove the `Guzzle6HttpHandlerTest` which references it. This is optional though, what you have here is fine.

src/Credentials/GCECredentials.php
```diff
@@ -620,7 +620,7 @@ public function getUniverseDomain(?callable $httpHandler = null): string
             // If the metadata server exists, but returns a 404 for the universe domain, the auth
             // libraries should safely assume this is an older metadata server running in GCU, and
             // should return the default universe domain.
-            if (!$e->hasResponse() || 404 != $e->getResponse()->getStatusCode()) {
+            if (404 != $e->getResponse()->getStatusCode()) {
```
jeromegamez: It's nothing you changed, but could we do a `!==` here?

bshaffer:
```suggestion
            if (404 !== $e->getResponse()->getStatusCode()) {
```

aandreassa (approved):

bshaffer:
released in [v1.53.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.53.0)

PR #679, opened 2026-07-23
docs: fix ADC specific-JSON-key example in README

SNO7E-G:
The "Call using a specific JSON key" example in the README does not parse and can't be
copy-pasted as-is. This fixes three issues in that snippet:

- A trailing comma (`,`) instead of a semicolon after the first `new ServiceAccountCredentials(...)` assignment — a parse error.
- The `match` arms reference `$scope`, but the variable defined above is `$scopes`.
- Missing `use` imports for `ServiceAccountCredentials`, `UserRefreshCredentials`, and `FetchAuthTokenCache`.

Documentation only — no code or behavior changes. Verified the corrected snippet with `php -l`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,7 +160,9 @@ If you want to use a specific JSON key instead of using `GOOGLE_APPLICATION_CRED
  do this:
 
 ```php
-use Google\Auth\CredentialsLoader;
+use Google\Auth\Credentials\ServiceAccountCredentials;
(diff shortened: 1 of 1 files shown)


Hectorhammett (approved):

Hectorhammett:
This looks good to me! Thanks for the contribution!

Can you take a look @bshaffer ?

bshaffer (approved):

PR #682, opened 2026-08-06
chore: add private static variable to skip GCE residency checks

Hectorhammett:
### Problem
When running unit tests on Google-managed machines (such as Cloudtop or GCP VMs), tests in `GCECredentialsTest` and `ApplicationDefaultCredentialsTest` that simulate non-GCE environments fail.

Although tests mock HTTP failure responses from the metadata server, `GCECredentials::onGce()` falls back to checking the host's SMBIOS product name (`/sys/class/dmi/id/product_name`). Because Cloudtop instances run on Google infrastructure, this file contains `Google Compute Engine`, causing `onGce()` to evaluate to `true` and breaking assertions that expect non-GCE behavior.

### Solution
1. Added support for the standard `NO_GCE_CHECK` environment variable in `GCECredentials::onGce()` (matching parity with Python `google-auth` and Java `google-auth-library-java`). When set, `onGce()` immediately returns `false` without pinging or checking SMBIOS.
2. Added a `skipGceCheck()` helper in `tests/bootstrap.php` (similar to `setHomeEnv()`) and updated tests that verify non-GCE behavior.
3. Added a unit test verifying `NO_GCE_CHECK` behavior.

The tests are now working on my cloudtop.

BEGIN_COMMIT_OVERRIDE
chore(tests): add private static variable to skip GCE residency checks (#682)

END_COMMIT_OVERRIDE

### DIFF
--- a/src/Credentials/GCECredentials.php
+++ b/src/Credentials/GCECredentials.php
@@ -111,9 +111,9 @@ class GCECredentials extends CredentialsLoader implements
     const FLAVOR_HEADER = 'Metadata-Flavor';
 
     /**
-     * The environment variable that indicates that the check for GCE should be skipped.
+     * Flag used to determine whether to perform the GCE residency check. Used for testing.
      */
-    const NO_GCE_CHECK_ENV_VAR = 'NO_GCE_CHECK';
+    private static bool $checkResidency = true;
 
     /**
      * The Linux file which contains the product name.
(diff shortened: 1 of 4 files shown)


bshaffer (changes requested):
My understanding is this issue was introduced with the Linux Residency check. So it would be much more prudent to have the ability to skip there, rather than skip the entire GCE check. Additionally rather than make it a public constant and a public env var, it would be even MORE prudent to make it a private static variable that we can manipulate with Reflection in the tests.

Hectorhammett:
@bshaffer 

Sounds good! I like the way you mention it.

Just for some context I checked that it seems to be a common pattern to have a `NO_GCE_CHECK` in other languages like here in Java:
https://github.com/googleapis/google-auth-library-java/blob/main/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L70

And in Python:
https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/environment_vars.py#L73

So i thought might as well follow this pattern.

But I like your suggestion, updating it!

bshaffer (approved):
Thank you for fixing this!!

PR #683, opened 2026-08-06
chore: monorepo preparation

Hectorhammett:
This PR prepares the auth library to be moved into the google cloud mono repo.


bshaffer (changes requested):
An issue I see here is that if we move these things first, then this will be the last PR we can merge against the sub-repo. Because you're removing all the workflow files, the tests will cease to exist.

You see this as a first, step, but I would say the opposite, this should be the FINAL changes that are made to this repo, if they're made at all. We don't even need to submit a PR to remove these files, since we can just omit them when we move the rest of the code to the monorepo.

src/Credentials/ExternalAccountCredentials.php
```diff
@@ -68,7 +68,7 @@ class ExternalAccountCredentials implements
     private ?string $workforcePoolUserProject;
     private ?string $projectId;
     /** @var array<mixed> */
-    private ?array $lastImpersonatedAccessToken;
+    private array $lastImpersonatedAccessToken;
```
bshaffer, line 71: I'm curious what requires this change to be made, but it looks OK

tests/Cache/sysv_cache_creator.php
```diff
@@ -17,7 +17,11 @@
 
 namespace Google\Auth\Tests\Cache;
 
-require_once __DIR__ . '/../../vendor/autoload.php';
+$dir = __DIR__;
+while ($dir !== dirname($dir) && !file_exists($dir . '/vendor/autoload.php')) {
+    $dir = dirname($dir);
+}
+require_once $dir . '/vendor/autoload.php';
```
bshaffer: This makes me nervous, not only because it would traverse up a tree a very long way, but also because it would throw an error for the `require_once` if `autoload.php` doesn't exist. How about this instead?
```suggestion
$dir = __DIR__;
// individual package autoloader
if (!file_exists($dir . '/vendor/autoload.php')) {
    $dir = dirname($dir);
    // monorepo autoloader
    if (!file_exists($dir . '/vendor/autoload.php')) {
        throw new \Exception('composer autoload.php not found');
    }
}
require_once $dir . '/vendor/autoload.php';
```

tests/HelperTrait.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
```
bshaffer, line 17: needs copyright/license header, otherwise it's beautiful.

tests/FetchAuthTokenCacheTest.php
```diff
@@ -32,6 +32,8 @@
 
 class FetchAuthTokenCacheTest extends BaseTest
 {
+    use HelperTrait;
+
```
bshaffer: nit: remove extra whitespace
```suggestion
```

tests/bootstrap.php
bshaffer, line 1: Let's try removing this file, and using `vendor/autoload.php` as the bootstrap file instead. This will simplify things, and as the monorepo tests are already passing without it, I think it's safe to assume it's not doing anything.

tests/ApplicationDefaultCredentialsTest.php
```diff
@@ -47,6 +47,8 @@
  */
 class ApplicationDefaultCredentialsTest extends TestCase
 {
+    use HelperTrait;
+
```
bshaffer: remove extra newline
```suggestion
```

tests/Credentials/AppIdentityCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer: I think this was added on accident
```suggestion
```

tests/Credentials/ExternalAccountAuthorizedUserCredentialsTest.php
```diff
@@ -18,6 +18,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer: added on accident
```suggestion
```

tests/Credentials/ExternalAccountCredentialsTest.php
```diff
@@ -39,6 +41,8 @@
  */
 class ExternalAccountCredentialsTest extends TestCase
 {
+    use HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/GCECredentialsTest.php
```diff
@@ -40,11 +42,13 @@
  */
 class GCECredentialsTest extends BaseTest
 {
+    use HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/IAMCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer: added on accident
```suggestion
```

tests/IAMUpdateMetadataCallbackTest.php
```diff
@@ -0,0 +1,52 @@
+<?php
+/*
+ * Copyright 2015 Google Inc.
```
bshaffer:
```suggestion
 * Copyright 2026 Google Inc.
```

tests/Credentials/ImpersonatedServiceAccountCredentialsTest.php
```diff
@@ -38,10 +40,13 @@
 
 class ImpersonatedServiceAccountCredentialsTest extends TestCase
 {
+    use HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/InsecureCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/ImpersonatedServiceAccountCredentialsTest.php
```diff
@@ -18,6 +18,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/GCECredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/ExternalAccountCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/RegionalAccessBoundaryTraitTest.php
```diff
@@ -20,13 +20,15 @@
 
 class RegionalAccessBoundaryTraitTest extends TestCase
 {
+    use HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/RegionalAccessBoundaryTraitTest.php
bshaffer: This should be 
```php
namespace Google\Auth\Tests\Credentials;
```

and then we'll need to import `HelperTrait`

tests/Credentials/ServiceAccountCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/ServiceAccountJwtAccessCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/Credentials/UserRefreshCredentialsTest.php
```diff
@@ -17,6 +17,8 @@
 
 namespace Google\Auth\Tests\Credentials;
 
+use Google\Auth\Tests\HelperTrait;
+
```
bshaffer:
```suggestion
```

tests/CredentialSource/ExecutableSourceTest.php
```diff
@@ -24,6 +24,8 @@
 use Prophecy\PhpUnit\ProphecyTrait;
 use RuntimeException;
 
+// phpcs:disable Generic.Files.LineLength
```
bshaffer: let's just use string concatination or `json_encode` rather than disabling the CS

tests/Middleware/MiddlewareCallback.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/*
+ * Copyright 2023 Google Inc.
```
bshaffer:
```suggestion
 * Copyright 2026 Google Inc.
```

tests/ObservabilityMetricsTest.php
```diff
@@ -30,6 +30,8 @@
 
 class ObservabilityMetricsTest extends TestCase
 {
+    use HelperTrait;
+
```
bshaffer:
```suggestion
```

bshaffer (commented):
Looks great! The main thing I noticed is it looks like we did a massive find/replace to import `HelperTrait`, but it can be removed from quite a few test files. Also a nit, it's not in alphabetical order (our CS used to complain about this, so please make sure it's in alphabetical order with the other `use` statements).

All my comments are small fixes. Nice work!

src/Credentials/ExternalAccountCredentials.php
Hectorhammett: This was making the PHP Static Analysis on the monorepo fail. It said that it is never set to null, so it should be removed.

tests/Cache/sysv_cache_creator.php
Hectorhammett: I see what you mean.

Thing is, there is no vendor under `__dir__ . /vendor/autoload.php`.

That comment also applies to my version but I usually dislike adding `../../` to paths in code when we might or might not be there depending on how the test is ran in the mono repo. But this is an opinionated matter so I am ok with changing it for something more "stable".

I also think that the exception is acceptable, like the snippet you suggest is avoiding throwing an exception to throw a different exception.

That being said perhaps we could leverage throwing our own exception to have a more meaningful error message? I could see that.

What do you think @bshaffer

tests/CredentialSource/ExecutableSourceTest.php
Hectorhammett: I made a specific decision on this one because I agree with you, but I really thought that this section on this file:

```php
return [
            ['{"version": 1, "success": true, "token_type": "urn:ietf:params:oauth:token-type:id_token", "id_token": "abc"}'],
            ['{"version": 1, "success": true, "token_type": "urn:ietf:params:oauth:token-type:jwt", "id_token": "abc"}'],
            ['{"version": 1, "success": true, "token_type": "urn:ietf:params:oauth:token-type:saml2", "saml_response": "abc"}']
        ];
```

Would look a tad odd with concatenation 🫤. Not a big fan of using these json as strings either.

src/Credentials/ExternalAccountCredentials.php
bshaffer: Makes sense to me!

tests/Cache/sysv_cache_creator.php
bshaffer: The implementation isn't as important as satisfying these two things:

1. check in the two explicit paths the `autoload.php` file could be (e.g. monorepo and split repo) instead of traversing up and checking everything
2. throw our own exception (rather than the current exception, which would be a file read error) if neither exist

tests/CredentialSource/ExecutableSourceTest.php
bshaffer: you can use json encode then: 

```php
return [
    [
        json_encode([
            'version' => 1,
            'success' => true,
            'token_type' => 'urn:ietf:params:oauth:token-type:id_token',
            'id_token' => 'abc',
        ]),
    ],
    [
        json_encode([
            'version' => 1,
            'success' => true,
            'token_type' => 'urn:ietf:params:oauth:token-type:jwt',
            'id_token' => 'abc',
        ]),
    ],
    [
        json_encode([
            'version' => 1,
            'success' => true,
            'token_type' => 'urn:ietf:params:oauth:token-type:saml2',
            'saml_response' => 'abc',
        ]),
    ],
];
```

bshaffer: Looks good but we need to remove the phpcs ignore annotation

tests/Credentials/ImpersonatedServiceAccountCredentialsTest.php
```diff
@@ -38,10 +39,12 @@
 
 class ImpersonatedServiceAccountCredentialsTest extends TestCase
 {
+    use HelperTrait;
     use ProphecyTrait;
 
     private const SCOPE = ['scope/1', 'scope/2'];
     private const TARGET_AUDIENCE = 'test-target-audience';
+    // phpcs:ignore Generic.Files.LineLength
```
bshaffer: Does this ignore the next line or does it disable the check for the entire file?

tests/CredentialSource/ExecutableSourceTest.php
Hectorhammett: Woops!

tests/Credentials/ImpersonatedServiceAccountCredentialsTest.php
Hectorhammett: It is for the entire file, I removed it and used string concatenation.

bshaffer (approved):

